PR #3, opened 2016-09-17
Update ReadME to NextCloud links

raghunayyar:
@georgehrke can you add more of the readme related changes here? The correct URIs for Travis and Scrutinizer and such stuff. ;)


georgehrke:
👍

Issue #37, opened 2016-09-17
Cleanup VTimezones when editing events


georgehrke:
cc @wrexroad

wrexroad:
This fix has been implemented upstream (mozilla-comm/ical.js#270). Once that dependency has been updated this issue should be closed.

PR #43, opened 2016-09-17
cleanup vtimezone

georgehrke:
fixes https://github.com/nextcloud/calendar/issues/37

by @wrexroad

### DIFF
--- a/js/app/controllers/editorcontroller.js
+++ b/js/app/controllers/editorcontroller.js
@@ -104,6 +104,7 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
 
 			$scope.prepareClose();
 			$scope.properties.patch();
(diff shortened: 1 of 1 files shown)


georgehrke:
I will again give it some thought were to put this.
I'm still not happy with the editor controller because we need this in other places too:
- In day or week view: drop an event from a time-slot to the all-day area
- importing events

wrexroad:
I agree that the editor controller isn't a good place for this, but it was the best I could think of at the time. It turns out that there has actually been an existing issue for this in the ical.js repo (mozilla-comm/ical.js#188) so I submitted a similar pull request over there. That seems like the better to have the iCal library handle this sort of thing rather than the calendar itself.

wrexroad:
Closing this pull request as the problem as been fixed upstream in ical.js (mozilla-comm/ical.js#270).

Issue #17, opened 2016-09-17
Show dropdown to suggest time based on title


georgehrke:
http://littlebigdetails.com/post/130947228205/osx-calender-new-events-default-to-starting-at

Keridos:
Any update on this?

georgehrke:
@Keridos Nope, otherwise the label `1 - to develop` would have been removed and replaced with `2 - developing` :)

vince-from-nice:
The issues #113 and #1470 has been closed because they should be implemented via that issue. But I don't see the point. The original issues was about event duplication as here we are talking about the possibility to reuse fields of existing events if I understand well. 

That means that the user must create a new event and then start to type a title with corresponds to an existing event (ie. the original event) in order to retrieve some (all?) of its data. IMHO that's not user friendly at all in comparison of a simple "duplicate" button on the original event. 

Maybe I don't have understood so please detail a bit...

PS: in the worst case I won't use the web interface but another client like KOrganizer or Thunderbird which provide a "duplicate" button (and event search, which is the another vital feature missing for me).

georgehrke:
> That means that the user must create a new event and then start to type a title with corresponds to an existing event (ie. the original event) in order to retrieve some (all?) of its data. 

That was my original plan. yes.

> IMHO that's not user friendly at all in comparison of a simple "duplicate" button on the original event.

How often often do you duplicate an event and why do you do it? I'm trying to understand your use-case and the workflow here.

> (and event search, which is the another vital feature missing for me).

see #8

vince-from-nice:
I use duplication very often. A typical use case is to duplicate my working days since I work ~4 days a week and my free day varies each week. A fixed recurrence rule is therefore poorly adapted in my case. A duplicate button (or even better copy/paste) would be very appreciated. Please have a look on the last comments of #113 
Thanks, Vincent.

JohnArcher:
I also use duplication very often. We use it (or better: want to use it) for the school calender of my kids. I create an event for a parent evening for grade 7 and then I (want to) duplicate that event 4 or 5 times for the whole school year, as time etc. is all the same, just the date differs obviously.

sarans-h:
Hii is it open to work ?? @jancborchardt If yes I need more clarity on this

PR #44, opened 2016-09-17
Mathieu schopfer parsing schedule

georgehrke:
continuing the discussion here 

@MathieuSchopfer

### DIFF
--- a/js/app/controllers/editorcontroller.js
+++ b/js/app/controllers/editorcontroller.js
@@ -86,6 +86,44 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
 		$scope.registerPostHook = function(callback) {
 			$scope.postEditingHooks.push(callback);
 		};
+		
+		$scope.parse_summary = function() {
+			var summary = $scope.properties.summary.value;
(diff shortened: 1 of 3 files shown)


tcitworld:
@MathieuSchopfer Can you add a few tests ?

MathieuSchopfer:
@tcitworld Sorry, what for tests ?
See this [fiddle](https://jsfiddle.net/MathieuSchopfer/0co07f8p/) for an equivalent example.

tcitworld:
Test that entering `2:15-4:30 Title` correctly creates an event with the according title and correct times.

MathieuSchopfer:
Oh ok, you want to automate the tests. I have to admit, I have never done it before.
Where should they be located (I guess calendar/tests/js/unit/controllers/) ? And how do you actually run them?

tcitworld:
In fact I think it would be better to have this code as a service. cc @georgehrke

georgehrke:
This pull-request lacks support for different locales, which will definitely not be an easy task to implement.

Closing in favour of https://github.com/nextcloud/calendar/issues/17 for now.

PR #57, opened 2016-09-19
add workaround to fix webcal creation under Nextcloud9

georgehrke:
Nextcloud9 has a bug, that it doesn't save color, enabled and displayname when sending a MKCOL for a subscription.
This patch mitigates the issue by sending an additional update request when running on Nextcloud 9.

### DIFF
--- a/controller/viewcontroller.php
+++ b/controller/viewcontroller.php
@@ -65,10 +65,10 @@ class ViewController extends Controller {
 	 */
 	public function index() {
 		$runningOn = $this->config->getSystemValue('version');
-		$runningOnServer91OrLater = version_compare($runningOn, '9.1', '>=');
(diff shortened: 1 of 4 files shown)


georgehrke:
decreased coverage is fine for now. There are no tests for the calendar service yet.

georgehrke:
What do you think @tcitworld @raghunayyar? :)

tcitworld:
Maybe we can send the fix to server too ?

georgehrke:
I'm not sure if it's going to be back ported. And I want to get 1.4.0 out by today :)

georgehrke:
It's already fixed in Nextcloud 10 and master

georgehrke:
So is this ok with you to merge? :)

tcitworld:
:+1: Needs rebase.

Issue #58, opened 2016-09-19
Clear URL from new subscription field


hitam4450:
I just installed version 1.4 and subscribed to a web calendar from Google. Everything worked fine for the first time. However, adding the second calendar, the "old" URL - from the first web calendar - was still in the box. It should be cleared after pressing (erstellen) Create!

regards, hitam4450

tcitworld:
Thought this was fixed, my bad.

PR #62, opened 2016-09-19
fix webcal input field not clearing & clean subscription controller

tcitworld:
Fixes https://github.com/nextcloud/calendar/issues/58


georgehrke:
Hm, why does `$scope.subscription.newSubscriptionUrl = '';` work but `$scope.newSubscriptionUrl = '';` doesn't?

tcitworld:
Seems like only the properties of the model can update the view. Not sure about it but there's informations that agree with this on Google.

georgehrke:
👍

Issue #65, opened 2016-09-19
New event popup editor shows update instead of create


tcitworld:
I guess the `isNew()` function isn't working properly.

PR #66, opened 2016-09-19
Fix update button appearing everytime

tcitworld:
Fixes https://github.com/nextcloud/calendar/issues/65


georgehrke:
This breaks read-only events and shows them as editable 😕

tcitworld:
Seems to be fixed.

georgehrke:
👍

Issue #35, opened 2016-09-17
global timezones broken


georgehrke:
from @wrexroad

### Steps to reproduce
1. Create a new event by clicking a date on the calendar.
2. Click the "More" option.
3. Select UTC, GMT, or Z timezone.
4. Click "Create" (or "Update" for an existing event).

### Expected behaviour
A new event with a global timezone should be created.

### Actual behaviour
The creating the event fails silently and the sidebar does not close.

### Server configuration
**Operating system**:
Ubuntu 14.04
**Web server:**
Apache
**Database:**
MySQL
**PHP version:**
5.5.9
**ownCloud version:** (see ownCloud admin page)
Nextcloud 9.0.53 (stable)
**Calendar version:** (see ownCloud apps page)
1.3.1 (Also tested on GitHub master branch commit 8addb4).
**Updated from an older ownCloud or fresh install:**
Freshly built from GitHub repo.
**Signing status (ownCloud 9.0 and above):**

```
No errors have been found.
```

### Client configuration
**Browser:**
Chrome
**Operating system:**
Fedora 24

georgehrke:
@wrexroad you said you were looking forward to fixing it this weekend. Can you send the pull request against this repository? :)

wrexroad:
Yes, I will push the update here. It looks like this is going to get bumped until tomorrow though.

PR #69, opened 2016-09-20
Fix global tz

wrexroad:
I implemented the changes from [this](https://github.com/owncloud/calendar/pull/667)  previously reverted pull request. Additionally, before creating the ICAL objects, `newSimpleData.dtstart.value._isUTC` and `newSimpleData.dtend.value._isUTC`  are set to `false`. This value being `true` was causing Moment.js to output UTC times, even when local times were requested which caused the additional offset. `_isUTC` is then returned to its previous once the ICAL objects are created so nothing else should be affected by this change.

I'm not sure why `_isUTC` is being set to true, but it seems like when `SimpleEvent.specificReader.date()`  is being called the Moment objects are in a state where they think they are local, but `_isUTC` is true so trying to get a local time returns a UTC time and getting a UTC time returns a doubly offset time.

@georgehrke, I want to do a little more poking around and testing before this is merged in, but I wanted to post this now so you could have a look at it and tell me if it fixes the offset problem you were seeing.

### DIFF
--- a/js/app/models/simpleeventmodel.js
+++ b/js/app/models/simpleeventmodel.js
@@ -495,10 +495,33 @@ app.factory('SimpleEvent', function() {
 				};
 			}
 
-			var start = ICAL.Time.fromJSDate(newSimpleData.dtstart.value.toDate(), false);
-			start.isDate = newSimpleData.allDay;
-			var end = ICAL.Time.fromJSDate(newSimpleData.dtend.value.toDate(), false);
-			end.isDate = newSimpleData.allDay;
+			var startTz = newSimpleData.dtstart.parameters.zone === "floating" ? ICAL.Timezone.localTimezone : ICAL.TimezoneService.get(newSimpleData.dtstart.parameters.zone);
+ 			var start = ICAL.Time.fromData(
+ 				{
+						year: newSimpleData.dtstart.value.year(), 
+						month: newSimpleData.dtstart.value.month() + 1, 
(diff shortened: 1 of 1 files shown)


wrexroad:
As an example of the doubly offset UTC times, here is some debugging output when I created an event starting at 7AM PDT (-0700)

```
newSimpleData.dtstart.value.hour() //15
newSimpleData.dtstart.value.local().hour() //15
newSimpleData.dtstart.value.utc().hour() //22
```

wrexroad:
It looks like `fullcalendar` is doing something with the `moment` objects that is causing them to have both the `_isUTC` and `_ambigZone:` flags set to `true`. I think for now just turning off `_isUTC`, getting the time info, then turning it back on is the way to go.

wrexroad:
@georgehrke , I noticed pull request #98 changed the name to simpleeventmodel.js. I looked though your commits and it looks like some of the relevant blocks of code for this pull request have changed a bit. Do you want me to fix this pull request to match the work you have done, or do you have other plans for fixing issue #35?

georgehrke:
> it looks like some of the relevant blocks of code for this pull request have changed a bit

It changed only very little :)
I will adapt your pr to the changes made in #98

wrexroad:
Yeah, it looked like only a few lines changed, but I didn't know if you had changed other things elsewhere (e.g. the moment.js version or how it is used) that might make this pull request irrelevant. Thanks!

wrexroad:
Replaced by #150

Issue #68, opened 2016-09-19
Scrolling not possible in left calendar menu


prilly:
### Steps to reproduce
1. Add multiple callendars to fill the left menu
2. when the left menu contains so many callendars that the menu needs scrolling to see them all
3. no scrolling function for left menu

### Expected behaviour
It should be possible to scroll the left callendar menu independantly from the main callendar page.

### Actual behaviour
callendars/subscriptions in left menu is hidden when the menu is filling up.

### Server configuration
**Operating system**:
Debian 8
**Web server:**
nginx/apache
**Database:**
mariadb
**PHP version:**
7.0
**Server version:** (see your admin page)
10.0
**Calendar version:** (see the apps page)
1.4.0
**Updated from an older installed version or fresh install:**
upgraded from oc 9.1
**Signing status (ownCloud/Nextcloud 9.0 and above):**

```
Login as admin user into your cloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your instance's installation folder
```

**The content of config/config.php:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your instance's installation folder

or 

Insert your config.php content here
(Without the database password, passwordsalt and secret)
```

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

#### LDAP configuration (delete this part if not used)

```
With access to your command line run e.g.:
sudo -u www-data php occ ldap:show-config
from within your instance's installation folder

Without access to your command line download the data/owncloud.db to your local
computer or access your SQL server remotely and run the select query:
SELECT * FROM `oc_appconfig` WHERE `appid` = 'user_ldap';


Eventually replace sensitive data as the name/IP-address of your LDAP server or groups.
```

### Client configuration
**Browser:**

**Operating system:**

**CalDAV-clients:**

#### Web server error log
```
Insert your webserver log here
```

#### Log file (data/owncloud.log)
```
Insert your ownCloud.log file here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log 
c) ...
```

georgehrke:
Please fill out the entire issue template. Especially the client configuration

tcitworld:
Fixed.

PR #75, opened 2016-09-20
fix scrollability of calendar-list

georgehrke:
possible fix for https://github.com/nextcloud/calendar/issues/68

please review @raghunayyar @tcitworld

### DIFF
--- a/css/app/calendarlist.css
+++ b/css/app/calendarlist.css
@@ -95,6 +95,10 @@
 	width: 80px;
 }
 
+#app-navigation .app-navigation-list {
+	width: 100%;
+}
+
 #app-navigation .app-navigation-list-item .calendarCheckbox {
 	position: absolute;
 	margin-left: 17px;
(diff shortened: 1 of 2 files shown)


tcitworld:
calendar-scroll

- [ ] Increase the width of the sidebar so that there's no scrollbar on x

georgehrke:
What browser/OS is that?

tcitworld:
Firefox 48/Debian

georgehrke:
Okay, so we should decrease the width of the new calendar field.

The width of the app navigation is defined in the nextcloud server styles. I would rather not override that.

tcitworld:
> The width of the app navigation is defined in the nextcloud server styles. I would rather not override that.

Indeed, but that leave even less space for long calendar names. :(

georgehrke:
we could also say `overflow-x: hidden`

tcitworld:
I think I had tested and it makes list item not centered and maybe even hid the Create button a little. Still acceptable though.

georgehrke:
the initial spinner moved a bit to the left. that needs to be fixed

tcitworld:
calendar-scroll-2

Not perfect, but that the only issue and I can go with it.

georgehrke:
Yes, it's still a huge improvement over the current situation :)

raimund-schluessler:
For the Tasks app I used something like
```
display: inline-flex;
flex-wrap: wrap;
width: 100%;
```
for the colorpicker and
```
width: calc(100% - XX px);
```
for the input field. This leads to a look like this:

PR #84, opened 2016-09-23
show tooltip for description on event hover

tcitworld:
Currently to view the description of an event you need (with default settings) : 
- to click on an event
- click _More_ on the popup

This intends to bring a tooltip when hovering on an event.
However, since afaik we can't apply the angular attributes (ui.bootstrap.tooltip) on the fullcalendar element, any idea on how I could do this ?

@georgehrke @raghunayyar @jancborchardt

### DIFF
--- a/css/app/calendar.scss
+++ b/css/app/calendar.scss
@@ -54,3 +54,13 @@
 #fullcalendar td.fc-day.fc-sun {
 	background-color: nc-darken($color-main-background, 3%);
 }
+
+.tooltip-inner {
+	white-space: pre-line;
+	max-height: 175px !important;
+	overflow: hidden;
+	display: -webkit-box;
+	-webkit-line-clamp: 10;
+	-webkit-box-orient: vertical;
+}
+
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -295,6 +295,17 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 							});
 						}
 					}
+
+					if (event.description) {
+						element.tooltip({
+							container: 'body',
+							placement: function(tip, element) {
+								const position = $(element).offset();
+								return (position.top < 300) ? 'bottom' : 'top';
+							},
+							title: event.description
+						});
+					}
 				}
 		};
 	}
--- a/js/app/models/fcEventModel.js
+++ b/js/app/models/fcEventModel.js
@@ -54,7 +54,8 @@ app.factory('FcEvent', function(SimpleEvent) {
 			backgroundColor: vevent.calendar.color,
 			borderColor: vevent.calendar.color,
 			textColor: vevent.calendar.textColor,
-			title: event.getFirstPropertyValue('summary')
+			title: event.getFirstPropertyValue('summary'),
+			description: event.getFirstPropertyValue('description')
 		};
 
 		Object.defineProperties(iface, {


georgehrke:
Do you mind if I hijack this PR?
I see a few issues here.

georgehrke:
- it doesn't support new lines :/

And one question: 
calendar - nextcloud chromium today at 2 54 47 pm
How do we want to handle a really long description?
Where do we want to cut it off?

tcitworld:
> it doesn't support new lines :/

This happens in the side panel too. A `nl2br`-like function might be handy here.

> Where do we want to cut it off?

Cut to 150/200 characters should be good.

tcitworld:
- [ ] Try displaying the tooltip on the bottom if the event is on top of the calendar (month) view.

georgehrke:
> This happens in the side panel too. A `nl2br`-like function might be handy here.

It does for you?
In the sidebar editor i can see new lines in the description and also add them. 

We currently parse the description as text. If we were to add nl2br, we would have to parse it as html and run into XSS issues. I'd like to avoid that.

georgehrke:
```css
.tooltip-inner {
    white-space:pre-wrap;
}
```
does the trick for new lines. will add it

tcitworld:
> It does for you?
In the sidebar editor i can see new lines in the description and also add them. 

It happens only on subscriptions events. This is another issue.

jancborchardt:
Some time ago we talked about a quicker, more integrated event popover kind of like in Apple Calendar (if I remember correctly). This should also show on hover to reveal the quick details, and then stay when clicked. Isn’t that more what we would want on hover?

Because lots of events don’t have descriptions. Of course if an event does have a description (or participants, or other important info), we could also show it in that popover.

georgehrke:
> and then stay when clicked. 

Hm, but when do we open the editor then?

tcitworld:
Rebased.

jancborchardt:
@tcitworld can you post a quick screenshot for easy review by @nextcloud/designers? :)

georgehrke:
@jancborchardt We still need to figure out how to cut off text for long descriptions like here:

calendar - nextcloud chromium today at 2 54 47 pm

We should probably go with a mix of character limit and new line limit

georgehrke:
Chromium:
calendar - nextcloud chromium today at 1 33 21 pm

Firefox:
calendar - nextcloud firefox today at 1 34 05 pm

I use special `-webkit-` css magic for the vertical ellipsis. Didn't figure out how to do it in Firefox in a reasonable time.

georgehrke:
please review @tcitworld :)

tcitworld:
`display: -webkit-box;` is basically an old spec for flexbox so I don't think we should have this here.

georgehrke:
AFAIK it's necessary when using `-webkit-line-clamp`. but maybe @jancborchardt has a better suggestion for the entire ellipsis thing :)

florom:
Hi @georgehrke , @tcitworld , @jancborchardt ,

As @jancborchardt suggested on 
https://help.nextcloud.com/t/see-whole-event-text-in-calender-app/18726/3
I will keep it in github.

Basically cutting off text in the event description makes the calendar in browser very diffecult to use,
caus it lacks overwiew in mounthly and weekly view. 

nextcloud screenshot cut text month

As you can see much of the text is not visible which is normally valuable information.

There should always be new lines instead determining the row width with the event of most lines.
like here: 

no text cut


Other views:
 
nextcloud screenshot cut text week

Same here in week.

nextcloud screenshot cut text day

Only in daily view it is slightly better.

I could not change this behaviour myself so I would also be open for quick&dirty suggestions to change it in the code/css stylesheet myself, everytime I update nextcloud, even if it costs performance, ...
Still an additional new issue would be great.

Thanks for reviewing in the first place.

georgehrke:
@florom I think we are talking about different issues here.
This PR is about showing the event description as a tooltip.
You propose to properly show multiple line summaries, right?

georgehrke:
> AFAIK it's necessary when using `-webkit-line-clamp`. but maybe @jancborchardt has a better suggestion for the entire ellipsis thing :)

ping @jancborchardt @skjnldsv

skjnldsv:
@georgehrke As far as I know, there is NOT a better solution unfortunately! :( 
W3C needs to get busy again! :)

florom:
@georgehrke 

> You propose to properly show multiple line summaries, right?

yes i think its simmilar to the tooltip but allows an overview at first sight. Many other calendar use this solution too. I don´t want it to be forgotten, its really missing in the webuse.

@skjnldsv :( :( I am really unhappy now reading these devastating words.

Still thanks for giving it some thought again.

georgehrke:
> yes i think its simmilar to the tooltip but allows an overview at first sight. Many other calendar use this solution too. I don´t want it to be forgotten, its really missing in the webuse.

similar from a  usability pov, but completely unrelated from a coding pov.
Do you mind creating a dedicated issue about it? :)

skjnldsv:
Any news here? :)

georgehrke:
Another issue, besides the entire overflow issues:
- The popover stays there when you click the event and doesnt disappear even if you move the mouse

burningTyger:
This is confusing. Are tooltips coming to calendar?

tcitworld:
For reference : https://css-tricks.com/line-clampin/

georgehrke:
Closing in favour of https://github.com/nextcloud/calendar/issues/680

PR #97, opened 2016-09-27
Import an .ics directly through Files app

tcitworld:
This registers an action in the files app to open an .ics an import it's content into the calendar app.

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -26,7 +26,7 @@
 * Description: The fullcalendar controller.
 */
 
-app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEventService', 'SettingsService', 'TimezoneService', 'VEvent', 'is', 'fc', 'EventsEditorDialogService', 'PopoverPositioningUtility', '$window', 'isPublic', 'constants', 'settings',
+app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEventService', 'SettingsService', 'TimezoneService', 'VEvent', 'is', 'fc', 'EventsEditorDialogService', 'PopoverPositioningUtility', '$window', 'isPublic', 'constants', 'settings', '$uibModal',
 	function ($scope, Calendar, CalendarService, VEventService, SettingsService, TimezoneService, VEvent, is, fc, EventsEditorDialogService, PopoverPositioningUtility, $window, isPublic, constants, settings) {
 		'use strict';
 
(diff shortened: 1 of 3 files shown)


controller/importcontroller.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	/**
+	 * @param integer $fileid
+	 * @return JSONResponse
+	 * @throws \Exception
+	 */
+	public function import($fileid) {
+		try {
+			$file = $this->folder->getById($fileid);
```
georgehrke: I haven't tried it, but `$folder` is the rootFolder. I'm pretty sure this allows a user to access all files in the Nextcloud instance 👎

js/app/controllers/calcontroller.js
```diff
@@ -133,6 +133,38 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
...
+						files: function () {
+							return [file];
+						}
+					},
+					scope: $scope
+				});
+			});
+		};
+
+		var hashParts = window.location.hash.substr(1).split('/');
```
georgehrke: please use `$window`. This makes unit testing much easier.

controller/importcontroller.php
tcitworld: Oh sure. Does `getUserFolder()` works properly ?

js/app/controllers/calcontroller.js
```diff
@@ -133,6 +133,38 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 			$scope.$apply();
 		});
 
+		$scope.importFileById = function(id) {
```
tcitworld: All of this should be moved to a dedicated file.

georgehrke:
I think this `ImportController` is an overkill.
You could use the path  instead of the fileid and simply load the file via WebDav?
GET `/remote.php/dav/files/admin/` + path

What do you think about this?

tcitworld:
It seems a good solution indeed, I just see two things that may be problematic : 

* Encoding issues in the GET request with special characters. Also path may be too long for an url ?
* What about huge (fake) files ?

georgehrke:
> It seems a good solution indeed, I just see two things that may be problematic : 

> * Encoding issues in the GET request with special characters. Also path may be too long for an url ?

The encoding issues are solved with a simple `encodeURI` in javascript.
The path being too long for the url is a very unrealistic case. Also a user would troubles accessing the files via WebDav, because it has the same limitations on URLs.

> * What about huge (fake) files ?

I see no difference to the ImportController approach here. A user could also hand over the fileid of a huge file to the import controller.

tcitworld:
- [ ] Test if federation-shared files work

js/app/controllers/calcontroller.js
georgehrke: Indeed.
Maybe I'm over-engineering here again, but we could at a HashService that parses the hash and other services could register for certain pattern of hashes.

I definitely see multiple use cases, where we need hashes in the future:
- this pr for importing a file
- linking directly to an event from a different app
- (linking directly to a date (-range) from a different app)
- register a protocol handler for adding webcal subscriptions

tcitworld:
Rebased.

js/app/controllers/calcontroller.js
```diff
@@ -146,6 +146,36 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 			});
 		}
 
+		$scope.importFileByPath = function(path) {
+			return $http.get(
+				OC.linkToRemoteBase('dav') + '/files/' + OC.getCurrentUser().displayName + '/' + decodeURI(path)
```
georgehrke: Displayname won't work, you have to use the userId

Mannshoch:
Little Question. Will I be able to decide between Linking the ics as Webcal into the Calendar and Importing into an existing calendar?

tcitworld:
> Will I be able to decide between Linking the ics as Webcal into the Calendar

This isn't intended since it would imply the ics is published, and I don't see the use case anyway.

Mannshoch:
I thought @georgehrke's post from  on 27 Sep 2016 speaks about this.
I have a published .ics file to allow People to subscribe to a calendar I publish. But the Webcal tool in Calendar could not connect this .ics file so at moment I collect all data with Thunderbird in a Webdav calendar and then Export and replace the published Files in Nextcloud periodical. I hoped to remove this manual part.

raoulbhatia:
Anything i can do to help?

tcitworld:
Rebased and little fixes.

freechelmi:
Any progress ? 

Use case I see : adding an ICS to calendar from a webmail app like the Rainloop one.

as rainloop can save attachement to nextcloud

georgehrke:
Closing in favour of https://github.com/nextcloud/calendar/issues/1483

PR #98, opened 2016-09-27
refactor events object

georgehrke:


js/app/models/simpleEventModel.js
```diff
@@ -117,85 +107,69 @@ app.factory('SimpleEvent', function() {
...
-					});
+					currentElement.value = valueParser(property);
 				}
 
 				data[key].push(currentElement);
-				properties[pKey].setParameter('x-oc-group-id', group.toString());
-				group++;
-			}
+				property.setParameter('x-oc-group-id', group.toString());
+				group++;
```
tcitworld: There's unnecessary spaces here.

js/app/models/veventModel.js
```diff
@@ -0,0 +1,289 @@
...
+
+				if (iCalEvent.isRecurring()) {
+					const duration = rawDtend.subtractDate(rawDtstart);
+					const iterator = new ICAL.RecurExpansion({
+						component: vevent,
+						dtstart: rawDtstart
+					});
+
+					let next;
+					while ((next = iterator.next())) {
```
tcitworld: Too many `(` `)`

georgehrke: I'd like to keep this actually, because we need to evaluate the result of the assignment.
I'm not sure if it works in all browsers when using `while(next=iterator.next()){...}`

js/app/models/simpleEventModel.js
georgehrke: fixed that

georgehrke:
please review @raghunayyar @tcitworld

raghunayyar:
👍

PR #100, opened 2016-09-28
Accessibility/semantic additions

tyrylu:
This pull request adds some role="button" attributes where it makes sense. Note that those spots are definitely not all of where it would be needed, but some additional semantics would require adding additional aria attributes (e. g. the day/month/week radio, the buttons showing/hiding part of the ui would probably want aria-collapsed etc.) and i did not feel adding ngaria dependency on my first contribution, but it is probably the only future choice, as i could not find a built in directive which would allow setting arbitrary attributes.
I did not decide what to do with the days in the calendar yet, maybe also add button roles to them, because for a visually impaired user (e. g. me) it was not obvious that you should click them when adding an event.

### DIFF
--- a/templates/part.datepicker.php
+++ b/templates/part.datepicker.php
@@ -23,13 +23,13 @@
  */
 ?>
 <div class="datepicker-heading">
-	<button type="button" class="btn btn-default btn-sm btn-arrow pull-left" ng-click="prev()">
+	<button type="button" class="btn btn-default btn-sm btn-arrow pull-left" ng-click="prev()" aria-label="&lt;">
 		<i class="glyphicon glyphicon-chevron-left"></i>
 	</button>
 	<button type="button" class="btn btn-default btn-sm btn-date" ng-click="toggle()">
 		<strong ng-cloak>{{ dt | datepickerFilter:selectedView }}</strong>
 	</button>
-	<button type="button" class="btn btn-default btn-sm btn-arrow pull-right" ng-click="next()">
+	<button type="button" class="btn btn-default btn-sm btn-arrow pull-right" ng-click="next()" aria-label="&gt;">
 		<i class="glyphicon glyphicon-chevron-right"></i>
 	</button>
 </div>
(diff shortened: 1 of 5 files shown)


templates/part.datepicker.php
```diff
@@ -23,13 +23,13 @@
  */
 ?>
 <div class="datepicker-heading">
-	<button type="button" class="btn btn-default btn-sm btn-arrow pull-left" ng-click="prev()">
+	<button type="button" class="btn btn-default btn-sm btn-arrow pull-left" ng-click="prev()" aria-label="&lt;">
```
georgehrke: How does a screenreader software pronounce `<`?

What about using `aria-label="previous"` instead?

georgehrke:
@tyrylu Thx a lot for your contribution :)
The `role="button"` change definitely makes sense!

templates/part.datepicker.php
georgehrke: or maybe "Go back" and "Go forward"

tyrylu:
@georgehrke those would be better, probably, yes. I am however not sure if they're in the translated strings already or what's the procedure when adding a new translatable string.,

georgehrke:
You don't have to worry about the translations :)
Just add `<?php p($l->t('Go back')) ?>` to the code.
There is a cronjob running every day, that checks the code for translatable strings and if necessary adds them to our transifex account. Community members will translate them then.

tyrylu:
Thanks for the tip. I will change the labels then. But i think that Next/Previous is sufficient and it is used everywhere else i can remember.

georgehrke:
👍

georgehrke:
can we get another review here? :) 
@jancborchardt @raghunayyar @tcitworld

georgehrke:
@jancborchardt We should probably work on some Nextcloud accessibility guideline for developers.
What do you think?

tyrylu:
Good idea.

tyrylu:
May be this link could be useful, because angular is used so often in the javascript ui: https://docs.angularjs.org/guide/accessibility

tcitworld:
:+1:

jancborchardt:
@tyrylu great accessibility work, thanks a lot! :) Pull requests in this area area always welcome!

tyrylu:
You are welcome. :smile:

PR #108, opened 2016-10-01
show email for attendee

### LINKED ISSUE #45: Attendee-Autocomplete: show e-mail address behind name
cc @schiessle

tcitworld:
I would like to have some `&hellip;` at the end when they're truncated, but `uib-typeahead` doesn't seem to accept ternaries directly, so I guess we need a full filter just for this.

Fixes #45

### DIFF
--- a/controller/contactcontroller.php
+++ b/controller/contactcontroller.php
@@ -101,12 +101,10 @@ class ContactController extends Controller {
 				$r['EMAIL'] = [$r['EMAIL']];
 			}
 
-			foreach ($r['EMAIL'] as $email) {
-				$contacts[] = [
-					'email' => $email,
+			$contacts[] = [
+					'email' => $r['EMAIL'],
 					'name' => $name
-				];
-			}
+			];
 		}
 
 		return new JSONResponse($contacts);
--- a/css/app/eventdialog.css
+++ b/css/app/eventdialog.css
@@ -266,8 +266,10 @@ button.delete:focus {
 	display: inline;
 }
 
-.dropdown-menu {
+.dropdown-menu,
+.attendeename {
 	width: 100%;
+	max-width: 470px;
 }
 
 .dropdown-menu li {
@@ -278,6 +280,9 @@ button.delete:focus {
 .dropdown-menu li a {
 	display: block;
 	padding: 5px;
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
 }
 
 .dropdown-menu li a:hover {
--- a/js/app/controllers/attendeecontroller.js
+++ b/js/app/controllers/attendeecontroller.js
@@ -77,7 +77,32 @@ app.controller('AttendeeController', function($scope, AutoCompletionService) {
 	};
 
 	$scope.search = function (value) {
-		return AutoCompletionService.searchAttendee(value);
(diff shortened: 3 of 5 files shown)


raghunayyar:
:+1:

jancborchardt:
@tcitworld is it maybe possible to ellipsize just using CSS and going with the width? Normal ellipsizing is achieved using:
```
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

tcitworld:
I'll try.

tcitworld:
Still the same issue on mobile, but at least it's okay on desktop view.

raghunayyar:
@nextcloud/designers anyone up for a review? :)

raghunayyar:
@georgehrke please review, its long pending. :)

jancborchardt:
@raghunayyar can you post a quick screenshot?

jancborchardt:
@raghunayyar can you post a quick screenshot?

tcitworld:
@jancborchardt 
Desktop

Mobile

(Don't mind the display of the button, it's a Firefox Test Pilot issue)

jancborchardt:
So as per the requirement in #45:

> otherwise a user is not able to distinguish which email address is used.

it’s only shown when a name is duplicate, right? Because normally it’s not necessary to show it. :)

And yeah, it would be good to get the ellipsizing down, should be possible by restricting the width of the input. cc @nextcloud/designers anyone wants to help out here? :)

georgehrke:
@tcitworld Can you please only display the email address when there are multiple email addresses available for a user :)

gqgunhed:
@georgehrke @tcitworld What about starting with the basic functionality as provided here and adding visual enhancements later on?
- showing email addresses only when contact has multiple email entries?
- maybe just showing the email domain part? (but that would collapse again in specific cases)
- select behavior via a setting switch in the calendar app?
Etc.
As most of my contact entries have multiple email addresses (work and home) it's very hard to figure out which is the correct name when setting up event invitations.
So thanks @tcitworld as I was searching for that functionality, but my JS/PHP skills are to bad ;-)

georgehrke:
> maybe just showing the email domain part? (but that would collapse again in specific cases)

Then some emails will only have the domain and others will have the entire email address.
We should try to keep it consistent. Also I don't see a benefit in only showing the domain.

> select behavior via a setting switch in the calendar app?

I'm sorry, but this won't happen. We are trying to make things simple. Showing the email address when multiple are available and hiding it when only one is available covers all usecases.
Adding a setting doesn't solve anything imho and only clutters the user interface. :)

georgehrke:
Changed the behavior to the expected one:

calendar - nextcloud chromium today at 1 44 03 pm

please review @jancborchardt @tcitworld @eppfel

tcitworld:
Oh, I thought I had pushed the change.
LGTM 👍

skjnldsv:
Are you using the select2 function? This should have the same design as tags, admin//workflow or other selects on nextcloud :)

georgehrke:
Not yet.
Is there proper documentation on select2 and presumably select1/select?

skjnldsv:
I don't know! 😞 
But systemtags is using it. https://select2.github.io

PR #123, opened 2016-10-12
improve info.xml for new appstore

georgehrke:
please review @raghunayyar @BernhardPosselt @tcitworld


appinfo/info.xml
```diff
@@ -1,20 +1,33 @@
...
+	<description><![CDATA[The Nextcloud calendar app is a user interface for Nextcloud's CalDAV server.
+
+	It integrates with other apps, allows you to manage calendars and events, display external calendars and invite attendees to your events]]></description>
 	<version>1.4.0</version>
 	<licence>AGPL</licence>
-	<author>Georg Ehrke, Raghu Nayyar, Thomas Citharel</author>
+	<namespace>Calendar</namespace>
+	<author homepage="https://georg.coffee">Georg Ehrke</author>
+	<author homepage="http://raghunayyar.com">Raghu Nayyar</author>
+	<author>Thomas Citharel</author>
```
tcitworld: You Can put https://tcit.fr even if this is not live.

tcitworld:
:+1:

appinfo/info.xml
```diff
@@ -1,20 +1,33 @@
 <?xml version="1.0"?>
-<info>
+<info xmlns:xsi= "http://www.w3.org/2001/XMLSchema-instance"
+	  xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
 	<id>calendar</id>
 	<name>Calendar</name>
+	<summary>Calendar GUI for Nextcloud's CalDAV server</summary>
+	<description><![CDATA[The Nextcloud calendar app is a user interface for Nextcloud's CalDAV server.
```
BernhardPosselt: This will be rendered as markdown so indention will be treated as code blocks :)

appinfo/info.xml
```diff
@@ -1,20 +1,33 @@
...
+		<nextcloud min-version="9" max-version="11" />
 	</dependencies>
-	<description>Calendar GUI for Nextcloud's CalDAV server</description>
 	<documentation>
 		<user>https://docs.nextcloud.com/server/10/user_manual/pim/calendar.html</user>
 	</documentation>
 	<category>productivity</category>
 	<website>https://github.com/nextcloud/calendar/</website>
 	<bugs>https://github.com/nextcloud/calendar/issues</bugs>
+	<discussion>https://help.nextcloud.com/</discussion>
```
BernhardPosselt: This won't validate. The idea is to link to an open thread

georgehrke: Hm, but why would you link to one particular thread?

All people commenting with their issues, questions and feature requests in one thread seems wrong to me.
How would you mark single questions as answered?

I would totally agree on a calendar category in our discourse though.

appinfo/info.xml
georgehrke: good catch!
fixed it

appinfo/info.xml
BernhardPosselt: The one particular thread kinda mimics the old comment system on the old app store: it was just a continues stream of comments and answers and I think that's fine for some use cases. 

I like the idea for a separate category per app but I'd go for a convention like **https://help.nextcloud.com/c/app-calendar** (e.g. if there is no discussion xml element we change the link to a category). 

The issue with this however is that a moderator needs to create that category first and I suppose scrolling through categories when each app has a separate one gets kinda messy. Still a good idea.

georgehrke: We could use /c/app/calendar, so it displays as a folder of categories

BernhardPosselt: @jospoortvliet

georgehrke:
I ditched the discussion tag.

Issue #31, opened 2016-09-17
Auto-refresh 


georgehrke:
We should implement this in two steps:
- [ ] periodically check for updates of event in known calendars
- [ ] periodically check for new or updated calendars

georgehrke:
requires #209

georgehrke:
Let's implement https://github.com/nextcloud/calendar/issues/40 first.

Otherwise stuff like keeping events being edited locked will be a huge pain.

cl4x:
no news on this features? in 1.7.0 the autorefresh do not work

georgehrke:
> in 1.7.0 the autorefresh do not work

Not surprising, since this issue is still open 😉 

Please check the information on the right. It contains all the information you need:
![A1E77917-B021-414C-96EC-E55A044874C6](https://user-images.githubusercontent.com/1250540/61063111-b038a100-a3ff-11e9-92ab-2f81e165f4ae.png)

cl4x:
ok sorry, i have see this 
![immagine](https://user-images.githubusercontent.com/47188814/61063295-04dc1c00-a400-11e9-8884-f99324601f72.png)

vchrizz:
using following workaround until this issue is fixed:
in file core/templates/layout.user.php add after `<title>`
```
                <?php
                        // if loaded app is calendar, add html meta tag to refresh page every 60 seconds
                        if ($_['appid']=='calendar') {
                                print '<meta http-equiv="refresh" content="60">';
                        }
                ?>

```

georgehrke:
I would strongly advise anyone against using that workaround, because it just reloads the calendar app every 60 seconds, no matter whether you are just creating / editing an event or not. Definitely not a solution.

vchrizz:
correct, thanks for pointing that out, i should have mentioned that!
as i do not use the app via webbrowser, but manage my events via smartphone, this is not a problem in my case.
for explanation: i use a tablet on a wall to display the calendar and for that it is not practical to be forced to manually refresh the page to see the actual calendar status.

dev-zero:
Will the tasks app automatically profit from the implementation in the calendar app or should I open a separate feature request there if I want automatically refreshed task lists?

georgehrke:
@dev-zero No, this issue is only about the calendar app. The tasks app will need to implement this separately.

JonathanTreffler:
It would be nice if this feature would also be implemented for the Dashboard Widget of the Calendar App.

[...]

PR #132, opened 2016-10-17
refresh events every 5 minutes

georgehrke:
partial fix for https://github.com/nextcloud/calendar/issues/31

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -179,7 +179,7 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 						fcEvent.title = t('calendar', 'New event');
 						fcEvent.className.push(fcEventClass);
 						fcEvent.editable = false;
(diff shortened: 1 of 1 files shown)


tcitworld:
Why not PROPFIND too every 5 minutes ?

georgehrke:
Note to myself - there is one more issue:
- click existing event to edit it
- try to drag it -> won't work because its locked
- wait for refresh to kick in
- dragging now suddenly works -> that's not how it's supposed to be

Possible solution:
- take fcEvent that's supposed to be edited
- remove it from calendar
- use `renderEvent` with `stick=true` to render as individual event

georgehrke:
@tcitworld PROPFIND takes twice the requests.

Also when it comes to refreshing calendars, it'd like to keep the original object from the initial request and not replace it every 30 seconds ^^

tcitworld:
Rebased.

georgehrke:
Needs a proper implementation in the vue app + cdav-library

PR #135, opened 2016-10-18
add note to not use root or sudo

georgehrke:
please review @tcitworld @jancborchardt @MariusBluem


tcitworld:
:+1:

MariusBluem:
Good Point :+1:

PR #140, opened 2016-10-19
border styles for grid, highlight full-hour horizontal lines, fix #119

### LINKED ISSUE #119: Add more contrast to today's color
It's difficult to distinguish on some screens.

cc @raghunayyar @jancborchardt

jancborchardt:
Fix #119, please review @moy @tcitworld @nextcloud/calendar :)

Before:
capture du 2016-10-19 11-29-28

After (all full-hour grid lines are darker #eee, just like the row lines in Files, user management and elsewhere):
capture du 2016-10-19 11-28-06

### DIFF
--- a/css/app/calendarlist.css
+++ b/css/app/calendarlist.css
@@ -493,6 +493,13 @@ ul.dropdown-menu li > a:hover {
 	border-bottom: none;
 }
 
+/* border styles for grid, highlight full-hour horizontal lines */
+.fc-unthemed tr td {
+	border-top-color: #eee;
+}
+.fc-unthemed tr:nth-child(even) td {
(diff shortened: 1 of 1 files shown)


skjnldsv:
👍 okay!

georgehrke:
👍

moy:
Much better after than before indeed IMHO. It's still rather light: for example, I'm now seeing this from my laptop's screen which has a TN panel (=> gets different contrast depending on the viewing angle). Contrast is perfect when viewed with the right angle, but the grid still disappears quickly when looking at the screen from above.

IOW, If that was just me I'd use a slightly darker color, but I'm OK with the proposal too.

PR #147, opened 2016-10-20
fix dropping between grid and allday

### LINKED ISSUE #16: Make event allday by drag and drop doesn't work

georgehrke:
fixes #16 

This fixes a longstanding issue, that caused users to not be able to drag and drop events between the all-day grid and the time-grids in day and week view.

Test this by dragging and dropping events from all-day to time grid and vice versa.


georgehrke:
added another unit test to make sure you can repeat the action multiple times in a row

tcitworld:
LGTM

Issue #12, opened 2016-09-17
UNDO calendar deletion


georgehrke:

georgehrke:
use OC.Notification to display a yellow notification at top.

georgehrke:
This was implemented in #926

PR #148, opened 2016-10-21
add undo for calendar deletion

georgehrke:
implements / fixes https://github.com/nextcloud/calendar/issues/12

please review @nextcloud/calendar @jancborchardt

### DIFF
--- a/css/app/calendarlist.css
+++ b/css/app/calendarlist.css
@@ -212,6 +212,16 @@
 	box-sizing: border-box;
 }
 
+#app-navigation .app-navigation-list-item {
+	max-height: 1000px;
+	transition: 200ms cubic-bezier(.22, .61, .36, 1) all;
+}
+
+#app-navigation .app-navigation-list-item.deleted {
+	max-height: 0;
+	overflow: hidden;
+}
+
 #app-navigation .app-navigation-list-item .utils {
 	display: inline-block;
 }
--- a/js/app/models/calendarListItemModel.js
+++ b/js/app/models/calendarListItemModel.js
@@ -21,7 +21,7 @@
  *
  */
 
-app.factory('CalendarListItem', function(Calendar, WebCal, isSharingAPI) {
+app.factory('CalendarListItem', function($timeout, Calendar, WebCal, isSharingAPI) {
 	'use strict';
 
 	function CalendarListItem(calendar) {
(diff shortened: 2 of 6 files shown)


tests/js/unit/controllers/calendarlistcontrollerSpec.js
```diff
@@ -73,7 +73,7 @@ describe('CalendarListController', function() {
 		expect($scope.newCalendarColorVal).toBe('');
 	});
 
-	it ('should delete the selected calendar', function () {
+	/*it ('should delete the selected calendar', function () {
```
eppfel: Auskommentierter Code?

css/app/calendarlist.css
```diff
@@ -212,6 +212,16 @@
 	box-sizing: border-box;
 }
 
+#app-navigation .app-navigation-list-item {
+	max-height: 1000px;
```
georgehrke: This is needed for the transition. You can't define a transition from height:0 to height:auto

tcitworld:
Maybe we can add event deletion in the same PR ?

georgehrke:
> Maybe we can add event deletion in the same PR ?

It's a different feature, so I'd like to implement this in a dedicated PR, but definitely in time for 1.5 ;)

georgehrke:
added a unit test to CalendarListItem test suite

georgehrke:
One test in timezoneservice will fail, but that's unrelated and will be fixed with #149

css/app/calendarlist.css
georgehrke: Because I need those styles now and not with Nextcloud 11 ;)
But yes, eventually we should unify them in the server and add some developer instructions.

tcitworld:
Code looks good. Will test asap.

js/app/models/calendarListItemModel.js
```diff
@@ -133,6 +134,30 @@ app.factory('CalendarListItem', function(Calendar, WebCal) {
...
+			return new Promise(function(resolve, reject) {
+				context.isDeleted = true;
+
+				const timeout = $timeout(function() {
+					if (context.isDeleted) {
+						resolve();
+					}
+				}, 7500);
+
+				const elm = OC.Notification.showTemporary(t('calendar', 'Undo deletion'));
```
tcitworld: Same remark than for events : replace with "Calendar has been deleted. Undo ?"

jancborchardt: Yup. But with the actual calendar name:

> **Calendarname** has been deleted. **Undo?**

tcitworld: Maybe even
> Calendar **calendarname** has been deleted

georgehrke:
i changed the string and rebased, please check again :)

tcitworld:
Rebased.

georgehrke:
This approach is at a dead end thats only gonna end up in a huge chain of ugly hacks.
The proper solution is gonna be https://github.com/nextcloud/server/issues/1662

PR #149, opened 2016-10-22
Undo event deletion

### LINKED ISSUE #13: UNDO event deletion
<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38041667-undo-event-deletion?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
fix #13 

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -63,6 +63,9 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 			VEventService.create(calendar, data).then(function(vevent) {
 				if (calendar.enabled) {
 					fc.elm.fullCalendar('refetchEventSources', calendar.fcEventSource);
+					fc.elm.fullCalendar('removeEvents', function(fcEvent) {
+						return fcEvent.artificiallyRendered;
+					});
 				}
 			});
 		}
@@ -232,9 +235,25 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 							createAndRenderEvent(result.calendar, result.vevent.data, view.start, view.end, $scope.defaulttimezone);
 						}
 					}).catch(function(reason) {
-						if (reason === 'delete') {
-							deleteAndRemoveEvent(vevent, fcEvent);
+						if (reason !== 'delete') {
+							return true;
 						}
+
+						fc.elm.fullCalendar('removeEvents', fcEvent.id);
+						fcEvent.delete().then(function() {
+							return VEventService.delete(vevent).catch(function() {
+								throw new Error();
+							});
+						}).catch(function() {
+							const start = view.start.subtract(1, 'day');
+							const end = view.end.add(1, 'day');
+
+							const eventsToRender = vevent.getFcEvent(start, end, $scope.defaulttimezone);
+							angular.forEach(eventsToRender, function (event) {
+								event.artificiallyRendered = true;
+								fc.elm.fullCalendar('renderEvent', event);
+							});
+						});
 					});
 				},
 				eventResize: function (fcEvent, delta, revertFunc) {
--- a/js/app/models/fcEventModel.js
+++ b/js/app/models/fcEventModel.js
@@ -21,7 +21,7 @@
  *
  */
 
-app.factory('FcEvent', function(SimpleEvent) {
+app.factory('FcEvent', function($timeout, SimpleEvent) {
 	'use strict';
 
 	/**
@@ -33,6 +33,7 @@ app.factory('FcEvent', function(SimpleEvent) {
 	function FcEvent(vevent, event, start, end) {
 		const context = {vevent, event};
 		context.iCalEvent = new ICAL.Event(event);
+		context.isDeleted = false;
 
 		let id = context.vevent.uri;
 		if (event.hasProperty('recurrence-id')) {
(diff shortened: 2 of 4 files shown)


tcitworld:
Will have a look asap.

js/app/models/fcEventModel.js
```diff
@@ -135,17 +136,27 @@ app.factory('FcEvent', function(SimpleEvent) {
...
+			return new Promise(function(resolve, reject) {
+				context.isDeleted = true;
+
+				const timeout = $timeout(function() {
+					if (context.isDeleted) {
+						resolve();
+					}
+				}, 7500);
+
+				const elm = OC.Notification.showTemporary(t('calendar', 'Undo deletion'));
```
tcitworld: I'll prefer "The event has been deleted. Undo ?"

js/app/models/fcEventModel.js
```diff
@@ -135,17 +136,27 @@ app.factory('FcEvent', function(SimpleEvent) {
...
-			context.lock = false;
+		iface.delete = function() {
+			return new Promise(function(resolve, reject) {
+				context.isDeleted = true;
+
+				const timeout = $timeout(function() {
+					if (context.isDeleted) {
+						resolve();
+					}
+				}, 7500);
```
tcitworld: 7.5 seconds is too high IMHO. The user has got plenty of time to switch to another app, resulting in the deletion action not executing (correct ?).

I don't know if there's a way to execute queue actions when we're leaving the app. Leaving the browser could be handled though, right ?

georgehrke: (https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload)

js/app/models/fcEventModel.js
georgehrke: @jancborchardt ^

js/app/models/fcEventModel.js
tcitworld: Seems good. So undo actions should be in a queue that gets emptied on timeout or on the `beforeunload` event.

js/app/models/fcEventModel.js
jancborchardt: Yes, for sure. The pattern is:
> **eventname** has been deleted. **Undo?**

jancborchardt:
(@georgehrke just btw: is it possible to reduce the Coveralls comments or not have them at all? We have the summary on the bottom with all the other checks, so the comments are just noise.)

georgehrke:
Coverall is only supposed to send one comment per commit pushed. :/
It's going crazy lately. 
I'll disable the comments

jancborchardt:
(Yeah, as said disabling is better than commenting because we have the summary here on the bottom anyway. Thanks!)

js/app/controllers/calcontroller.js
```diff
@@ -233,9 +233,24 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
...
+							});
+						}).catch(function() {
+							const start = view.start.subtract(1, 'day');
+							const end = view.end.add(1, 'day');
+
+							const eventsToRender = vevent.getFcEvent(start, end, $scope.defaulttimezone);
+							angular.forEach(eventsToRender, function (event) {
+								fc.elm.fullCalendar('renderEvent', event);
+							});
+						});
```
wrexroad: Rerendering undeleted events like this causes them to double render when another event is created. I know that is a confusing way to word it so I took a shot video (sorry about the screen tearing...)
Double render

Fortunately, this is simply a rendering artifact and resolves itself on reload. 

If you rerender the events like this:
`fc.elm.fullCalendar('refetchEventSources', vevent.calendar.fcEventSource);`
it does not appear to cause this problem.

georgehrke: Good catch!

I tried using refetchEventSources, but I think it's a bit problematic. The event takes at least one second to reappear, because the browser has to send a CalDAV request to the server.
It would be way better, if the event would reappear immediately.

georgehrke:
I rebased, changed the string and fixed the issue mentioned by @wrexroad 
please check again :)

georgehrke:
The mysql and postgres builds failed, because it couldn't reach the database.
The sqlite tests passed fine.

wrexroad:
Two thoughts:
1. It looks like the double rendering is still an issue, just in a different way. If you undelete an event, then hide the calendar to which it belongs, the event is not hidden. Then, when you unhide the calendar the event gets double rendered.

2. It looks like the onbeforeunload event listener was never added.

georgehrke:
> It looks like the onbeforeunload event listener was never added.

Yes, I want to solve that at another point. We can't send requests in onbefureonload, because the requests are async. The browser won't wait for them to finish but just cancel them.
Making them synchronous is an absolute no-go, because it would block any user interaction with the app and we would also have to patch one library we use.

wrexroad:
I think ensuring events actually get deleted is really important. Think of the situation where someone knows that the just need to quickly load up the page, delete an event, and close the browser. If they don't know that they have to wait 7 seconds before closing the browser it will create confusion.

I feel like the worst part of it is the event instantly disappears, giving a visual cue that the event is, in fact, gone. How would you feel about changing the notification to be a countdown. Something like "<event name will be deleted in <3,2,1>s. Click to cancel." Only when the timer runs out will you remove the event from the calendar and delete it from the server. This also solves the double rendering issue.

Another idea is as soon as the delete button is clicked dump the event properties to temporary location then immediately delete the event from the server and clear the event from the calendar. Then if the undo button is clicked you can create a new event using the properties that were saved. If the button is not clicked you dump the temporary storage.

georgehrke:
> I think ensuring events actually get deleted is really important. Think of the situation where someone knows that the just need to quickly load up the page, delete an event, and close the browser. If they don't know that they have to wait 7 seconds before closing the browser it will create confusion.

The only thing we can do with onbeforeunload is to ask the user if he really wants to leave the page.
And the problem is that modern browsers don't allow custom messages anymore :/ 

> Another idea is as soon as the delete button is clicked dump the event properties to temporary location then immediately delete the event from the server and clear the event from the calendar. Then if the undo button is clicked you can create a new event using the properties that were saved. If the button is not clicked you dump the temporary storage.

AFAIK all invitations will be send out again in this case.
That's also not really the expected behavior :/

wrexroad:
>AFAIK all invitations will be send out again in this case.
Yikes that's no good at all. I hadn't even considered invitations...

georgehrke:
@jancborchardt What's your opinion about the leaving the page thing? 

I checked what files does, but the undo is completely gone even if files_trashbin is disabled.

eppfel:
It would be awesome if you could cue jobs on the server, like "delete event x in 7 seconds, if you don't here a cancel message" Just a thought...

georgehrke:
> It would be awesome if you could cue jobs on the server, like "delete event x in 7 seconds, if you don't here a cancel message" Just a thought...

Hm, what if you are traveling in a train? I wouldn't exactly trust the cancel request to make it's way to the server (in time) ;)

Regarding the server: I'm also working on a retention feature: https://github.com/nextcloud/server/issues/1662

georgehrke:
This approach is at a dead end thats only gonna end up in a huge chain of ugly hacks.
The proper solution is gonna be https://github.com/nextcloud/server/issues/1662

PR #175, opened 2016-11-11
don't save new view in publicMode

georgehrke:
don't save new view in publicMode, because the request will fail anyway

please review @tcitworld @raghunayyar @MariusBluem


tcitworld:
:+1:

PR #181, opened 2016-11-14
save 301 responses from webcal subscriptions

### LINKED ISSUE #42: Update webcal url stored in caldav when it is a 301

georgehrke:
fixes #42

### DIFF
--- a/controller/proxycontroller.php
+++ b/controller/proxycontroller.php
@@ -77,13 +77,57 @@ class ProxyController extends Controller {
 	public function proxy($url) {
 		$client = $this->client->newClient();
 		try {
-			$clientResponse = $client->get($url, [
-				'stream' => true,
-			]);
-			$response = new StreamResponse($clientResponse->getBody());
-			$response->setHeaders([
-				'Content-Type' => 'text/calendar',
-			]);
+			$queryUrl = $url;
+			$allow_redirects = false; // try to handle redirects manually at first
+			$redirect_count = 0;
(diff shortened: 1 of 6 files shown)


controller/proxycontroller.php
```diff
@@ -76,14 +76,60 @@ public function __construct($appName, IRequest $request,
 	 */
 	public function proxy($url) {
 		$client = $this->client->newClient();
+		//$history = new GuzzleHttp\
```
georgehrke: - [ ] remove

controller/proxycontroller.php
```diff
@@ -76,14 +76,60 @@ public function __construct($appName, IRequest $request,
...
+				} else {
+					$done = true;
+				}
+			} while(!$done && $redirect_count <= $max_redirects);
+
+			if ($redirect_count > 0 && $redirect_count <= $max_redirects) {
+				$response = new JSONResponse([
+					'proxy_code' => -4,
+					'new_url' => $queryUrl,
+				], Http::STATUS_IM_A_TEAPOT);
```
tcitworld: 303 should do it

georgehrke: Nope, I thought about using a 3xx code. But the browser will expect a Location header and just forward there.

tcitworld: 400 can handle this situation.

georgehrke:
please review @tcitworld

PR #214, opened 2016-12-06
app should be listed in the office category too

ChristophWurst:
Didn't find the app on NC11 because I looked into the 'Office & text' category. IMO it should be listed there too like contacts https://github.com/nextcloud/contacts/blob/master/appinfo/info.xml and mail https://github.com/nextcloud/mail/blob/master/appinfo/info.xml


tcitworld:
👍

georgehrke:
👍

Issue #21, opened 2016-09-17
Link with anchor to event or day


georgehrke:

nickvergessen:
Subscribing for activity

georgehrke:
lets split this up into two different PRs:
- [ ] link to day 
- [ ] link to event

georgehrke:
requires the changes made to hash service in https://github.com/nextcloud/calendar/pull/256

blizzz:
📢 #256 is already merged. :D

tcitworld:
Will probably wait for https://github.com/nextcloud/calendar/issues/540 and vue-router.

georgehrke:
Yes, let's wait for #540 here. Doesn't make much sense to implement it twice.

caugner:
I'm not sure if you were already considering this, but the router value should maybe be empty if the current day/week/month is shown (to prevent newly added bookmarks from pointing to the day the bookmark was set).

georgehrke:
@jancborchardt @skjnldsv Quick usability / design question.

When linking to an event that's recurring (e.g. from the activity app),
which occurrence should be opened?
1️⃣ The first one
2️⃣ The closest one in the future or if that's not available the most recent one

I would personally go for 2️⃣

skjnldsv:
I would also say 2

nickvergessen:
Definetly 2. otherwise you end up on a year(s) old date because someone just modified a recent event entry

vaandefanel:
I prefer option 2. Can you add the possibility to specify an occurence ?

georgehrke:
@vaandefanel Yes, there will be the option to link to a specific occurrence. The question was only what to do when linking to a recurring event when no specific  occurrence was given. :)

jancborchardt:
Agree with everyone: Of course it should be the next upcoming one. :) (It’s already confusing that in invite mails sometimes the first occurence of a repeating event was listed. :D)

PR #217, opened 2016-12-07
Enhancement/noid/hash service

georgehrke:
implements a hash_service

needed for:
https://github.com/nextcloud/calendar/issues/21
https://github.com/nextcloud/calendar/issues/41

### DIFF
--- a/js/app/service/hashService.js
+++ b/js/app/service/hashService.js
@@ -1,0 +1,90 @@
+/**
+ * Calendar App
+ *
+ * @author Georg Ehrke
+ * @copyright 2016 Georg Ehrke <oc.list@georgehrke.com>
+ *
+ * This library is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


georgehrke:
fixed the missing semicolon

js/Gruntfile.js
```diff
@@ -136,6 +136,7 @@ module.exports = function(grunt) {
 		babel: {
 			options: {
 				presets: ['es2015'],
+				plugins: ["array-includes"],
```
tcitworld: Hmm, seems like `String.prototype.includes` (and same with array) is already available for our requirements : https://kangax.github.io/compat-table/es6/#test-String.prototype.contains

georgehrke: Good catch!
Will remove it.

georgehrke:
removed the babel plugin, let's wait for the tests

Issue #72, opened 2016-09-20
Small UX improvements


eppfel:
I skimmed the calendar and found a lot of small things, we can fix to make the UI more consistent and less crammed.

~~\- [ ] Remove the background/border Box around the label for the selection of month/week/day, because it is not an interactive element~~
<img width="263" alt="bildschirmfoto 2016-09-20 um 08 35 22" src="https://cloud.githubusercontent.com/assets/1899308/18667811/5d8e8b12-7f32-11e6-8f5e-d10d5a6ceb57.png">  
- [x] If there is only one calendar, remove calendar selection from event forms #89 

<img width="491" alt="bildschirmfoto 2016-09-20 um 08 36 48" src="https://cloud.githubusercontent.com/assets/1899308/18668822/052dd756-7f38-11e6-8d55-5ab66aa7a783.png">  

- [x] Do not (/ do only visually) disable the time field (enable it on click, toggle `all day event`), so the user is not blocked. 
- [ ] Place time in the same line as date in details view
- [x] Make Location and description fields single line with autosize for multiline support (See nextcloud/server#1407)

Maybe a bigger thing to do in a separate window, but wanted to start the discussion:
- [x] Make some event details optional ( in settings?) and provide a `more options` button to save space in the details view #88 

@nextcloud/designers Gimme thoughts on this?

@georgehrke If you point me to the right places, i can fiddle with it myself.

georgehrke:
@nextcloud/calendar

ccoenen:
I could swear, the "july 2016" used to be clickable just a few days ago? It did bring up a tiny calendar. Maybe it's just currently broken?

georgehrke:
Yes, it is ...
https://github.com/nextcloud/calendar/issues/48

ccoenen:
also bring `all day event` up in the form field. If you're entering stuff via keyboard, currently you have to

* title <kbd>Tab</kbd>
* calendar selection <kbd>Tab</kbd>
* start date <kbd>Tab</kbd> end date <kbd>Tab</kbd>
* now at all day event <kbd>spacebar</kbd> (uncheck all day event)
* <kbd>Shift-Tab</kbd> <kbd>Shift-Tab</kbd> <kbd>Shift-Tab</kbd> (get back to start time)
...

georgehrke:
@ccoenen there is a dedicated issue https://github.com/nextcloud/calendar/issues/25

ccoenen:
Thanks, I moved my comment over there!

[...]

PR #226, opened 2016-12-10
disable all-day when clicking on all-day

georgehrke:
please review @nextcloud/calendar @eppfel

### DIFF
--- a/css/app/eventdialog.css
+++ b/css/app/eventdialog.css
@@ -116,8 +116,17 @@
 	font-size: 20px;
 }
 
+.advanced .events--time--wrapper,
+.events .events--time--wrapper {
+	display: inline !important;
+}
+
 .events .events--time {
 	width: 60px;
+}
+
+.events .events--time[disabled] {
+	pointer-events: none;
 }
 
 .events--textarea,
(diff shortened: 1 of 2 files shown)


georgehrke:
ref #72

css/app/eventdialog.css
```diff
@@ -120,6 +120,10 @@
 	width: 60px;
 }
 
+.events .events--time[disabled] {
```
georgehrke: https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

click doesn't work on disabled inputs. With `pointer-events:none` the clicks on the input do straight to the parent div

georgehrke:
@eppfel please check again :)

eppfel:
Still no text cursor, but with the at least with the mouse no second click is needed...

eppfel:
Oh, but the time input is now in a new line, did this happen in master or is this regression?

georgehrke:
this also happens in master

georgehrke:
but this might be related to the span, I'll check

georgehrke:
fixed that :)

eppfel:
I approve once more 👍

PR #227, opened 2016-12-10
make title, description and location autoresizing textareas

georgehrke:
please review @nextcloud/calendar @eppfel


georgehrke:
ref #72

PR #256, opened 2016-12-21
add web protocol handler for webcal

### LINKED ISSUE #41: add web-based protocol handlers for webcal

georgehrke:
~~tests don't pass locally, no clue why 😕~~
Jasmine does weird stuff sometimes

fixes #41

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -127,7 +127,7 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 		});
 
 		if (!isPublic) {
-			CalendarService.getAll().then(function (calendars) {
+			$scope.calendarsPromise = CalendarService.getAll().then(function (calendars) {
 				$scope.calendars = calendars;
 				is.loading = false;
 				// TODO - scope.apply should not be necessary here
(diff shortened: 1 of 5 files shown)


georgehrke:
please review @tcitworld

tcitworld:
Tested with this url : `webcal://ical.mac.com/ical/US32Holidays.ics`
The app is opened and the url correctly set, the button « Create » is locked, but it took **ages** for the server to download the file, and the logs show an HTTP 422 error code, though the calendar is finally correctly added, could you have a look ?

tcitworld:
Hmm, in fact, this url takes ages to load even when added as a subscription. Quite odd, since it downloads itself in half a second though a `wget`.

georgehrke:
There are reasons why this is a bit slow:
- first of all, we need wait for the calendar list to be loaded, so we no which urls are taken already
- then it tries to connect to https://ical.mac.com/ical/US32Holidays.ics
 - in this case https doesn't work, an in fact Apple was unable to configure their server correctly and https://ical.mac.com/ical/US32Holidays.ics eventually times out for me ...
- then it ties http://ical.mac.com/ical/US32Holidays.ics which succeeds 
- and finally it creates the subscription

georgehrke:
Let's merge this for now. I will add an issue about better feedback.

Redsandro:
This is merged but the handler is not registered in my Nextcloud instance. Is it removed? Or am I confusing something else?

I'm looking for that handler that shows up for example the first time you open google calendar: https://github.com/nextcloud/calendar/issues/290#issuecomment-849942177

tcitworld:
It was removed in the vue rewrite, follow https://github.com/nextcloud/calendar/issues/3211 for updates.

Issue #136, opened 2016-10-18
Add printing styles


georgehrke:

eppfel:
👎 for the print button, it's the browsers functionality.

tcitworld:
> 👎 for the print button, it's the browsers functionality.

A number of users don't know about this.

Gaming4LifeDE:
When is this going to be implemented?

ChristophWurst:
The is no ETA at the moment. This feature is not planned nor did anyone point out that they are working on it.

IanJSaul:
Would be very useful - thanks.

kaffeeringe:
I would also take an csv-export, so I can build my own view in Calc or Excel.

rokejulianlockhart:
@eppfel and @tcitworld, per https://github.com/nextcloud/calendar/issues/136#issuecomment-279331332, does a standard API exist for a website to invoke the browser's print dialog?

terenc3:
> @eppfel and @tcitworld, per [#136 (comment)](https://github.com/nextcloud/calendar/issues/136#issuecomment-279331332), does a standard API exist for a website to invoke the browser's print dialog?

Yes https://developer.mozilla.org/en-US/docs/Web/API/Window/print

rokejulianlockhart:
Consequently, the print button could at least for now merely invoke that, and we could consider a custom interface later. Does seem a little useless though, so I think a custom interface is probably the better option since Menu > Print is all that most browsers need to print a page.

localhorst:
What is the status here? Are there any print styles available? Maybe externally, or someone created custom styles for that?

PR #279, opened 2017-01-05
hide sidepanel on print view

tcitworld:
Hides the sidebar on print view.

We also could consider bundling https://github.com/fullcalendar/fullcalendar/blob/master/src/common/print.css

### DIFF
--- a/css/app/print.css
+++ b/css/app/print.css
@@ -1,0 +1,5 @@
+@media print {
+	#app-navigation {
+		display: none;
+	}
+}


tcitworld:
We should also change the title of the window to match the current view so that it's printed on the headers : https://github.com/nextcloud/calendar/issues/280

georgehrke:
The PR is working as described in the PR. But you won't be able to see which color belongs to which calendar. 
This is definitely better then the current solution, but we need a proper print styles in the long run.
I will create an issue.

georgehrke:
there already is: https://github.com/nextcloud/calendar/issues/136

alexanderdd:
@georgehrke hmm, left panel is not hidden for me in print view in NC 18.0.6 calendar 2.0.3 firefox 79.0
also the calendars in the sidebar have no colors next to them so you dont know which calendar belongs to which events

PR #296, opened 2017-01-17
don't toggle allday when clicking on disabled time field when editor in readonly mode

georgehrke:


georgehrke:
@tcitworld Can you review this please :)

PR #317, opened 2017-01-24
change WebDAV to iCal wording, fix #233

### LINKED ISSUE #233: Phrasing "Settings" on public page is confusing
Hi,

### Steps to reproduce
1. Add a calendar, share it publicly.
2. On the public URL of the calendar, try to access to the matching ICS file, to feed it into some external tool such as Google Agenda.

### Expected behaviour

There is no way to directly access (easily) the URL of the ICS file. I would have expected to have something like an input field with the URL of the ICS file somewhere, besides the "Download" button.

### Actual behaviour

The only solution I could find is to actually download the file, and rely on my browser to get the source URL of the ICS file, which is highly unpractical, and especially tricky for non-tech users.

### Server configuration

Latest version of NextCloud and calendar app as of today. The rest does not seem to be relevant for this issue.

Thanks!

jancborchardt:
Fix https://github.com/nextcloud/calendar/issues/233, please review @nextcloud/calendar

### DIFF
--- a/templates/part.publicinformations.php
+++ b/templates/part.publicinformations.php
@@ -54,7 +54,7 @@
 					<div class="davbuttons">
 						<div class="btn-group">
 							<button class="button first" ng-model="$parent.publicdav" uib-btn-radio="'CalDAV'">CalDAV</button>
-							<button class="button last" ng-model="$parent.publicdav" uib-btn-radio="'WebDAV'">WebDAV</button>
+							<button class="button last" ng-model="$parent.publicdav" uib-btn-radio="'iCal feed'">iCal feed</button>
 						</div>
 					</div>
 					<label>{{ $parent.publicdavdesc }}</label>
--- a/templates/part.calendarlist.item.php
+++ b/templates/part.calendarlist.item.php
@@ -91,7 +91,7 @@
 		<li ng-show="item.isWebCal()">
 			<button	ng-click="item.showWebCalUrl()">
 				<span class="icon-link svg"></span>
-				<span><?php p($l->t('WebCal URL')); ?></span>
+				<span><?php p($l->t('iCal link')); ?></span>
 			</button>
 		</li>
 		<li>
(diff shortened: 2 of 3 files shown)


jancborchardt:
@georgehrke if you prefer »iCal feed« then please update this and merge it. :) But the current state is not acceptable

jancborchardt:
@nextcloud/calendar please review

@georgehrke I changed it to »iCal feed«. If anything else needs to be changed please just do it. We should merge it soon because the current thing is just wrong.

georgehrke:
calendar - nextcloud chromium today at 12 39 03 pm

We are calling it Webcal within the calendar. Do you want me to rename it there?

jancborchardt:
Yes, so it’s consistent. To »iCal link«, right?

skjnldsv:
iCal seems like an apple product :D 

I won't mind WebCal or iCal imho. But without the 'url'.

georgehrke:
> iCal seems like an apple product :D

The entire standard is called iCalendar :P

---
Will change the wording to »iCal link«

eppfel:
If it is a url, it should be called url. Link is semantically incorrect. The only thing other than URL I can think of is address.

But not a big deal, just my 2 cents.

skjnldsv:
To be faire I was thinking without url or link.
Just "WebCal" ^^'

templates/part.publicinformations.php
```diff
@@ -54,7 +54,7 @@ class="settings-button"
 					<div class="davbuttons">
 						<div class="btn-group">
 							<button class="button first" ng-model="$parent.publicdav" uib-btn-radio="'CalDAV'">CalDAV</button>
-							<button class="button last" ng-model="$parent.publicdav" uib-btn-radio="'WebDAV'">WebDAV</button>
+							<button class="button last" ng-model="$parent.publicdav" uib-btn-radio="'iCal feed'">iCal feed</button>
```
jancborchardt: @georgehrke shouldn't we call it"iCal link" here too then?

georgehrke: Hm, maybe.

PR #331, opened 2017-02-04
Flex on app-navigation buttons

### LINKED ISSUE #312: Today button/link not working when the translation for it is too long
Today button/link is not working / is in the wrong row when the translation for it is too long. Please see the attached image of the current behaviour.

### Steps to reproduce
1. Fresh install of Nextclud 11.0.2
2. Fresh install of the default calendar app
3. User using language where the "Today" word is longer then 5 characters like in finnish "Tänään" which is 6 characters long
capture2

### Expected behaviour
The link / button should be it is default row/line with other links / buttons

### Actual behaviour
Today button falls one row/line below other and it not working, when you click the button it only shows add new calendar dialog.

### Server configuration
Debian Jessie
Apache 2.4
Mysql 5.7
PHP 5.6

Nextclud 11.0.2 (Fresh install)
Calendar 1.5.0 (Just activated the app from the apps page)
Admin page passes all checks ok.

skjnldsv:
Fix #312
- [x] Add flex to the select date buttons
- [x] Add flex to the calendar view buttons
- [ ] Fix the overall design of the app-navigation section since it doesn't fit the nextcloud css guidelines
- [x] Fix the datepicker

### DIFF
--- a/css/app/calendarlist.css
+++ b/css/app/calendarlist.css
@@ -322,7 +322,7 @@
 	overflow-y: auto;
 	overflow-x: hidden;
 	position: absolute;
-	top: 85px;
+	top: 126px;
 	bottom: 44px;
 	width: 250px;
 	transition: 200ms margin ease-in-out;
--- a/css/app/datepicker.css
+++ b/css/app/datepicker.css
@@ -32,10 +32,6 @@
 	color: #111;
 }
 
-#app-navigation .togglebuttons .button.today {
-	float: right;
-}
-
 #app-navigation .togglebuttons .button.today:focus {
 	background-color: rgba(240, 240, 240, .9);
 }
@@ -72,6 +68,7 @@
 
 .datepicker-heading {
 	margin: 0 5px;
+	display: flex;
 }
 
 .datepicker-heading .btn-arrow {
@@ -82,9 +79,9 @@
 }
 
 .datepicker-heading .btn-date {
-	width: 181px;
 	height: 36px;
 	margin-left: 4px;
+	flex-grow: 1;
 }
 
 #datepicker table thead tr {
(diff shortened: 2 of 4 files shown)


georgehrke:
@skjnldsv Any progress on this? :)

skjnldsv:
@georgehrke I'm far too busy lately! I'm so sorry :(
There isn't much left to do, if someone wants to take over? :)

georgehrke:
@jancborchardt @skjnldsv @eppfel  Do we want to use icons as discussed in https://github.com/nextcloud/calendar/issues/312#issuecomment-278090752 or stick to the text labels?

eppfel:
> @jancborchardt @skjnldsv @eppfel Do we want to use icons as discussed in #312 (comment) or stick to the text labels?
I'm in favor of icons.

To avoid mystery meat we should use labels, but that will break the layout I guess.
Maybe something for another PR...

Another thing: we should switch Today and day/week/month. The date picker/arrows and today actually move through time and therefore are more related to each other then day/week/month, as they just switch the layout.

georgehrke:
fix for the date picker is in #449

jancborchardt:
@eppfel where do you see mystery meat at the moment? The arrows are directly next to the month name so it is reasonably related to what they do.

georgehrke:
> @eppfel where do you see mystery meat at the moment? The arrows are directly next to the month name so it is reasonably related to what they do.

We were talking about replacing the labels "day" "week" and "month" with icons, not the arrows :)

jancborchardt:
Ah ok - well then of course have icons and the text underneath. :)

eppfel:
> Ah ok - well then of course have icons and the text underneath. :)
Could work...

jancborchardt:
Maybe better, just modifying the view to move the »Today« button to full width as @georgehrke demonstrated in https://github.com/nextcloud/calendar/issues/312#issuecomment-277270283 – then we can add the icons left of the text for each of the buttons.

georgehrke:
I will move the today button below the views now. We can add the icons in a later iteration.

georgehrke:
calendar - nextcloud chromium today at 11 44 14 am

jancborchardt:
Looks cool! :) What do you think about also visually connecting the arrow left and right buttons to the month name as the Day/Week/Month radio buttons are? Would look much nicer that way.

georgehrke:
@jancborchardt I'm not a big fan of that.
1. I'm looking forward to add a fourth view, agenda (https://github.com/nextcloud/calendar/issues/402)
Then there won't be enough space left.

2. the view buttons are for changing how you see your events.
The arrows are for changing the visible time range.
So is the date picker in the top. So the arrows should definitely remain there imho.

raimund-schluessler:
I think @jancborchardt means to get rid of the space between the arrow buttons and the month name.

georgehrke:
Ahhhhhhh, now i got it.
Yes, definitely.

georgehrke:
calendar - nextcloud chromium today at 11 13 17 am

georgehrke:
ping @jancborchardt ^

skjnldsv:
👍

PR #346, opened 2017-02-20
show proper am/pm labels next to timepicker

### LINKED ISSUE #345: Timepicker ununderstandable
I think of myself as a person that is able to understand the concept of time. But how the hell am I supposed to create an event for 14:00 here? And why are there numbers twice?

screen shot 2017-02-20 at 10 31 26

cc @georgehrke On c.nc.c

georgehrke:
fixes #345 

please review @tcitworld @skjnldsv 

before:

after:
calendar - nextcloud chromium today at 11 11 57 am

### DIFF
--- a/js/app/directives/datetimepickerDirective.js
+++ b/js/app/directives/datetimepickerDirective.js
@@ -88,7 +88,7 @@ app.directive('ocdatetimepicker', function($compile, $timeout) {
 			}
 			function initTimepicker() {
 				element.find('.events--time').timepicker({
-					showPeriodLabels: false,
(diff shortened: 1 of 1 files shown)


georgehrke:
PS: this will only show up if you use the 12h format. When you select a language like german (with 24h), you won't see these labels.

skjnldsv:
👍 nice.

How can I know which language comes with a h24? Can I use en with h24 format?

georgehrke:
> Can I use en with h24 format?

No, not yet.
See https://github.com/nextcloud/server/issues/1781

PR #354, opened 2017-03-01
testing branch for server pr

georgehrke:
branch for testing https://github.com/nextcloud/server/pull/3620

Don't merge ;)


schiessle:
So now that the server part was merged we need something like this on the calendar, maybe with a Nextcloud server version / capabilities check

georgehrke:
> So now that the server part was merged we need something like this on the calendar, maybe with a Nextcloud server version / capabilities check

Yes, I will open an issue for that :)

georgehrke:
https://github.com/nextcloud/calendar/issues/406

PR #362, opened 2017-03-10
debug travis

georgehrke:


georgehrke:
travis loads phpunit 6 in some cases.
our tests don't run with phpunit 6, we need to enforce phpunit 5

PR #371, opened 2017-03-18
make sure user can not set dtend to something earlier than dtstart

### LINKED ISSUE #11: make sure user can not set dtend to something earlier than dtstart

georgehrke:
fixes #11


georgehrke:
please review @nextcloud/calendar @tcitworld :)

PR #372, opened 2017-03-20
Public sharing improvements

### LINKED ISSUE #232: Optically highlight which link is selected
georg cloud share

The current selection (either CalDAV or WebCal) should be highlighted just like the current view.

### LINKED ISSUE #240: Show proper emptycontent view for non-existing public calendar links
Currently when you go to https://cloud.nextcloud.com/apps/calendar/public/non-existing-calendar there’s just an empty calendar. This calendar does not actually exist and there should be indication of that. Instead of the calendar view on the right, there should be an emptycontent-like view similar to the Files app for empty folders. :)

Saying something like:

> [icon-calendar]
> **Calendar does not exist**
> Maybe the link is wrong?

cc @nextcloud/calendar

### LINKED ISSUE #243: Calendar shared by link: overall design & left navigation
Currently a calendar shared by link looks like this:
capture du 2016-12-15 02-36-40

Some thoughts:
- There’s no blue bar on top which clearly states »Nextcloud«. It’s not only a branding thing but also to know where you are. We do that everywhere else when sharing – Files, Spreed, Gallery etc.
- The left navigation looks a bit lost. We could have the calendar name a bit bigger like a h2 maybe? Similar to the »Files drop« feature in the server
- Also similar to the »Files drop«, it would be cool to not only show the user id in parentheses like `(jan)` but actually show a separate line with avatar and full name (fallback to username)
- Bug: the right border of the left navigation is not going through the whole way. The lower half (except next to settings) is basically missing.

What do you think @nextcloud/calendar @nextcloud/designers?

### LINKED ISSUE #248: Mobile menu position incorrect
Seems related to calendar in public mode only.
screenshot_20161215-133931
screenshot_20161215-133939

georgehrke:
fixes #232
fixes #240
fixes #243 
fixes #248


georgehrke:
requires https://github.com/nextcloud/calendar/pull/373 for the avatar directive

georgehrke:
Just added the empty content warning:
nextcloud chromium today at 4 52 47 pm

georgehrke:
I would suggest the following changes
- I moved download and subscribe (iCal Feed) to the upper right corner. I don't think we actually need the CalDAV link. Everywhere on the web where you can subscribe to a calendar it's webcal. I've never seen CalDAV before. It's probably more confusing than beneficial.
- If we decide to remove the CalDAV link, rename "Subscribe & embed" to "Embed and change the icon to something like "</>"
- Hide the embed section completely when viewing the calendar that is embedded itself.

nextcloud chromium today at 5 04 21 pm

(Additionally, I will add the user avatar above the calendar name)

tcitworld:
> Maybe you got the wrong link

Or the calendar got unshared.

tcitworld:
> Maybe remove Embed from the public sharing site completely and show it inside the calendar app

Why not, but I guess we already have issues to place things in the calendar app itself, so...

I agree with the rest. Nice work !

georgehrke:
i changed the way we generate the public sharing url.
Reminder to myself:
- [ ] test with a nextcloud instance in a subfolder
- [ ] test with calendar app in a second appfolder

georgehrke:
The only thing missing here is the avatar.
I need the avatar directive from #373 for that :)

tcitworld:
So let's merge #373 for now and then open another PR for global work on the sidebar ?

georgehrke:
Hm. Yes. I think that would be the best solution :)

georgehrke:
New navigation bar of public sharing site:
nextcloud chromium today at 12 04 45 pm

tcitworld:
Oh. Maybe put ~~your face~~ the avatar a bit smaller ? :)

georgehrke:
Hm. yes.

It's 128px, just like the public file sharing, at the moment, but there is also much more whitespace on the public file sharing site.
Resizing it down to 96 ...

templates/part.publicinformations.php
```diff
@@ -22,51 +22,30 @@
...
-	<span class="action permanent displayname">{{ item.calendar.displayname }}</span>
-
-	<span class="icon-download svg public-ics-download"
-		ng-click="download(item)">
-		<?php p($l->t('Download')); ?>
-	</span>
+<div id="scollable" class="settings-fieldset-interior public-left-side" ng-repeat="item in calendarListItems">
+	<div class="avatardiv" data-user="{{ item.calendar.owner }}" data-size="96" avatar></div>
+	<h3 class="action permanent displayname" ngCloak>
+		<?php print_unescaped($l->t('%s shared the calendar <strong>%s</strong> with you', ['{{ item.getOwnerName() }}', '{{ item.getPublicDisplayname() }}'])); ?>
```
georgehrke: @LukasReschke Is this safe?

georgehrke:
Jan had issues with Unknown timezone, caused by changed $rootScope.baseUrl

jancborchardt:
Also, clicking the link icon in the sharing section opens a new tab of the files app instead of the shared calendar.

And instead of the calendar favicon, for the emptycontent please use `.icon-calendar-dark` :)

georgehrke:
@jancborchardt The unknown timezone issue should be gone now

georgehrke:
> Also, clicking the link icon in the sharing section opens a new tab of the files app instead of the shared calendar.

This was caused by the same issue that caused the "Unknown timezone" thing.
Should be fixed now as well

georgehrke:
please review @tcitworld @raimund-schluessler @jancborchardt @eppfel :)

templates/main.php
```diff
@@ -56,6 +56,16 @@
...
+<style>
+	@media only screen and (max-width: 768px) {
+		#app-navigation, #app-content {
+			top: 45px !important;
+		}
+	}
+</style>
+<?php endif; ?>
+
 <?php if($_['isEmbedded']): ?>
```
tcitworld: Maybe add $_['isPublic'] here too.

georgehrke: isEmbedded will only ever be treu when isPublic is true too

georgehrke:
nextcloud chromium today at 4 39 30 pm

That's the new 404 screen with the grey calendar icon

georgehrke:
please test this branch again to make sure the timezone error is gone :)

jancborchardt:
Great, the timezone issue is fixed. :)

Check please review too @nextcloud/calendar

georgehrke:
@jancborchardt Fixed the flashing text issue with my latest commit :)

templates/main.php
georgehrke: @tcitworld Are you fine with not adding isPublic here?

tcitworld:
Maybe it's related to changes in core, but the users pic in the sidebar resolves for me to `/index.php/avatar//96`, which is of course not found.

georgehrke:
The data-user attr is empty. I don't think it's a core issue, but rather related to the latest ng update

georgehrke:
@tcitworld please pull latest master and check again :)

jancborchardt:
Ok, so is this good to go? :) @raghunayyar @tcitworld did you review? :)

georgehrke:
We will have to make use of https://github.com/nextcloud/server/pull/4401 for master and Nextcloud 12, but I would like to address that in a follow-up PR.

georgehrke:
ping @tcitworld :)

georgehrke:
I will go ahead and merge this tomorrow evening if no one objects

PR #373, opened 2017-03-20
add avatar to sharing dropdown

### LINKED ISSUE #207: Add avatar to sharing list
Currently the search dropdown and the list of shares don't use the displayname and the avatar. Instead it only shows the id (ugly on ldap). Both should be fixed.

georgehrke:
fixes #207

### DIFF
--- a/css/app/calendarlist.css
+++ b/css/app/calendarlist.css
@@ -459,17 +459,14 @@ li.calendar-share-item span {
 }
 
 ul.dropdown-menu {
-	max-height: 100px;
+	max-height: 200px;
 	overflow-y: scroll;
+	overflow-x: hidden;
 	border: 1px #ddd solid;
-	border-radius: 0 0 3px 3px;
+	border-radius: 0 0 4px 4px;
 	margin-top: -7px;
 	margin-left: 6px;
-	width: 94%;
-}
-
-ul.dropdown-menu li {
-	height: 25px;
+	width: 89%;
 }
 
 ul.dropdown-menu li:last-child {
@@ -477,10 +474,7 @@ ul.dropdown-menu li:last-child {
 }
 
 #app-navigation ul.dropdown-menu a {
-	min-height: 22px;
-	max-height: 22px;
-	padding: 0 10px;
-	line-height: 2;
+	padding: 6px;
 }
 
 ul.dropdown-menu li > a:hover {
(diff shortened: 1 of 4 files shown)


georgehrke:
calendar - nextcloud chromium today at 11 01 01 pm

georgehrke:
The avatars should be added to the sharing list as well, but I'm generally not too happy about the current design of the "shared-with" list.

calendar - nextcloud chromium today at 11 02 38 pm

cc @jancborchardt @eppfel @nextcloud/designers

georgehrke:
I would prefer some design like this: (just like the files app)
files - nextcloud chromium today at 11 08 24 pm

but we have too little space in the navigation bar :/

georgehrke:
cc @nickvergessen who reported the original avatar issue

georgehrke:
And there is no real visual indication whether the expanded area belongs to `test 2` or `test 3`
calendar - nextcloud chromium today at 11 15 00 pm

tcitworld:
> And there is no real visual indication whether the expanded area belongs to test 2 or test 3

Maybe the whole area could have the calendar color instead of just the dot ?

georgehrke:
calendar - nextcloud chromium today at 12 02 31 am

That's with alpha .2

Hm. It could work, but there is no other place in Nextcloud doing it this way.
I'd prefer to use something we already use in other places of Nextcloud.

tcitworld:
> Hm. It could work, but there is no other place in Nextcloud doing it this way.

Yeah well I can't think of any other part of Nextcloud where the sidebar has to be reduced this small to show the calendar aside, so whatever we choose, I guess it wouldn't be just like the files app. :(

georgehrke:
I created an issue about the "shared-with" list: https://github.com/nextcloud/calendar/issues/377

tcitworld:
Works correctly. Just a small issue on my side : 

selection_030

georgehrke:
@tcitworld What browser / os is that? :)

tcitworld:
Firefox 52 on gnu/linux. Chrome/ium doesn't have this issue.

georgehrke:
calendar - nextcloud firefox today at 5 08 32 pm

Looks fine with Firefox on macOS

tcitworld:
Seems to be a resolution issue. If I use the adaptive view of Firefox to change the resolution (including the one I actually use), the issue is gone.

jancborchardt:
By the way, see how the share dropdown is nice and white in the files sidebar – can you take the same styles from there? :)

Issue #83, opened 2016-09-23
automate timezone updates


georgehrke:
We can probably reuse this script with some minor changes
http://hg.mozilla.org/comm-central/file/tip/calendar/timezones/update-zones.py

Issue #82, opened 2016-09-23
Wrong timezone for Europe/Moscow


blshadow:
### Steps to reproduce
1. Set Europe/Moscow time zone on client PC
2. Create new calendar event
3. Press More..., then press time zone button
### Expected behaviour

Moscow time zone is appears
### Actual behaviour

Minsk time zone is appears
### Server configuration

**Operating system**:
Gentoo Linux

**Web server:**
Apache 2.4.23

**Database:**
MariaDB 10.0.27

**PHP version:**
5.6.25

**Server version:** (see your admin page)
10.0.0

**Calendar version:** (see the apps page)
1.4.0

**Updated from an older installed version or fresh install:**
Updated from Owncloud 9.1.0

**Signing status (ownCloud/Nextcloud 9.0 and above):**

```
No errors have been found.
```

**List of activated apps:**

```
Enabled:
  - activity: 2.3.2
  - admin_audit: 1.0.0
  - calendar: true
  - comments: 1.0.0
  - contacts: 1.4.0.0
  - dav: 1.0.0
  - documents: 0.13.1
  - federatedfilesharing: 1.0.1
  - federation: 1.0.1
  - files: 1.5.2
  - files_accesscontrol: 1.1.2
  - files_automatedtagging: 1.1.1
  - files_external: 1.0.2
  - files_pdfviewer: 0.8.1
  - files_retention: 1.0.0
  - files_sharing: 1.0.0
  - files_texteditor: 2.1
  - files_trashbin: 1.0.0
  - files_versions: 1.3.0
  - files_videoplayer: 0.9.8
  - firstrunwizard: 1.1
  - gallery: 15.0.0
  - notifications: 0.3.0
  - password_policy: 1.0.0
  - provisioning_api: 1.0.0
  - serverinfo: 1.1.1
  - systemtags: 1.0.2
  - tasks: 0.9.3
  - templateeditor: 0.1
  - theming: 1.0.1
  - updatenotification: 1.0.1
  - workflowengine: 1.0.1
Disabled:
  - encryption
  - external
  - survey_client
  - user_external
  - user_ldap
  - user_saml
```

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "ocgcqtvbj5tn",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "cloud.blshadow.org.ru"
        ],
[...]

PR #397, opened 2017-04-01
replace server side timezone querying logic

georgehrke:
fixes https://github.com/nextcloud/calendar/issues/82
fixes https://github.com/nextcloud/calendar/issues/83


georgehrke:
ToDo: 
- [ ] add proper readme to `js/timezones` about license MPL v2
- [ ] update zones.json to latest 2017b

js/package.json
```diff
@@ -49,5 +49,8 @@
   },
   "engines": {
     "node": ">=6"
+  },
+  "dependencies": {
+    "gulp-insert": "^0.5.0"
```
tcitworld: Why not in devdependencies ?

georgehrke: Sorry, that was not intended. Will fix it later today

georgehrke: done :)

PR #405, opened 2017-04-04
update jstz detect and fullcalendar

georgehrke:


georgehrke:
Regarding jstz:
The one in bower is this unofficial one: https://github.com/HenningM/jstimezonedetect

And the maintainer only updated the jstz.min.js, but not the jstz.js
I reported that months ago, no reaction so far ...
https://github.com/HenningM/jstimezonedetect/issues/6

tcitworld:
Hmm, the installation of `mercurial-bower-resolver` worked for me on Node v6, but it seems to fail on v7.7.3.

georgehrke:
works fine for me with v7.8.0

PR #407, opened 2017-04-05
fix date timepicker layout in events editors

### LINKED ISSUE #72: Small UX improvements

georgehrke:
fixes #72 

Before = Blue
After = green

calendar - nextcloud chromium today at 10 42 21 pm
calendar - nextcloud chromium today at 10 42 34 pm
calendar - nextcloud chromium today at 10 43 07 pm
calendar - nextcloud chromium today at 10 43 18 pm
calendar - nextcloud chromium today at 10 43 37 pm
calendar - nextcloud chromium today at 10 44 16 pm

### DIFF
--- a/css/app/eventdialog.css
+++ b/css/app/eventdialog.css
@@ -95,12 +95,26 @@
 }
 
 .events .events--fieldset select {
-	width: 99.5%;
+	width: 100%;
 	margin-right: 0;
 }
 
+.events .event-time-interior {
+	width: 170px;
+}
+
 .events .events--date {
-	width: 70px;
+	width: 55%;
+	border-right-width: 0;
+	border-top-right-radius: 0;
+	border-bottom-right-radius: 0;
+	margin-right: 0;
+}
+
+.events .events--date:focus,
+.events .events--date:hover {
+	border-right-width: 1px;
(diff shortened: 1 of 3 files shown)


tcitworld:
Very nice !

PR #408, opened 2017-04-05
increase font-size of fullcalendar labels

### LINKED ISSUE #166: Fonts are too small
picture1

A client just told me about an interesting behavior of the calendar on bigger screens 😁 If you look at this screenshot - I think you will agree that the whole "date, hour and day" stuff is not readable for everybody who hasn't eyes of an 😼

cc @nextcloud/accessibility @nextcloud/designers

georgehrke:
fixes #166 

before:
calendar - nextcloud chromium today at 11 08 36 pm

after:
calendar - nextcloud chromium today at 11 07 31 pm

### DIFF
--- a/css/app/calendar.css
+++ b/css/app/calendar.css
@@ -43,3 +43,9 @@
 	max-width: 0 !important;
 	overflow: hidden !important;
 }
+
+#fullcalendar .fc-axis,
+#fullcalendar .fc-day-header {
+	font-size: 100%;
+	opacity: .5;
+}


MariusBluem:
Thanks :)

jancborchardt:
Can we reduce the opacity a bit though? Right now it overpowers the action elements on the left, and the event labels themselves.

georgehrke:
@jancborchardt It's already down to .5, not sure less is still readable.

Do you mind sending a pull request?

PR #417, opened 2017-04-12
Fix AdBlock blocking share icon

jancborchardt:
Same as https://github.com/nextcloud/server/pull/4337, but for the share icons in the Calendar.

Please review @kumzugloom @Pommesumdreher @MariusBluem @Bullnados @GetBoz @georgehrke @MariusBluem :)

### DIFF
--- a/templates/part.calendarlist.item.php
+++ b/templates/part.calendarlist.item.php
@@ -49,7 +49,7 @@
 		  ng-class="{'withitems': item.calendar.isShared() || item.calendar.isPublished() }">
 		<span
 			class="calendarlist-icon share permanent"
-			ng-class="{'icon-shared': item.calendar.isShared() && !item.calendar.isPublished(), 'icon-public': item.calendar.isPublished(), 'icon-share': !item.calendar.isShared() && !item.calendar.isPublished()}"
+			ng-class="{'icon-shared': item.calendar.isShared() && !item.calendar.isPublished(), 'icon-public': item.calendar.isPublished(), 'icon-shared': !item.calendar.isShared() && !item.calendar.isPublished()}"
 			ng-click="item.toggleEditingShares()"
 			ng-if="item.showSharingIcon()"
 			title="<?php p($l->t('Share Calendar')) ?>"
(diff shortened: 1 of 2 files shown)


ryanbr:
Given the domains for this tool changes, no way to cover all instances. If nextcloud has changed the element from icon-share to icon-shared? does that fix it?

templates/part.calendarlist.item.php
```diff
@@ -49,7 +49,7 @@
 		  ng-class="{'withitems': item.calendar.isShared() || item.calendar.isPublished() }">
 		<span
 			class="calendarlist-icon share permanent"
-			ng-class="{'icon-shared': item.calendar.isShared() && !item.calendar.isPublished(), 'icon-public': item.calendar.isPublished(), 'icon-share': !item.calendar.isShared() && !item.calendar.isPublished()}"
+			ng-class="{'icon-shared': item.calendar.isShared() && !item.calendar.isPublished(), 'icon-public': item.calendar.isPublished(), 'icon-shared': !item.calendar.isShared() && !item.calendar.isPublished()}"
```
nickvergessen: now the key exists twice

jancborchardt: @georgehrke could you please look into this? The images share.svg and shared.svg are exactly the same. They if shared or not should only differ in opacity of the icon, like in Files.

jancborchardt:
@georgehrke @raghunayyar can you check this out so it’s ready for Nextcloud 12? :)

jancborchardt:
Any review please @kumzugloom @Pommesumdreher @MariusBluem @Bullnados @GetBoz @georgehrke @MariusBluem?

georgehrke:
Jan, stay calm :)

The calendar is not affected by today's feature freeze. I will merge this before we release the next calendar version 

Please excuse my brevity and typos.

jancborchardt:
I know, still it’s good practice do do this as early as possible. ;)

georgehrke:
Fixed the styles:
calendar - nextcloud chromium today at 4 42 14 pm

PR #443, opened 2017-05-11
set ending time one hour later also when clicking on disabled time input (in order to disable allDay)

### LINKED ISSUE #438: Default ending time should be 1 hour later then start
### Steps
1. Enable "skip simple editor"
2. Create a new event
3. Click in the time field and set a time

### Expected
Ending 1 hour later

### Actual
Ending at start, makes no sense and might cause problems....

georgehrke:
fixes #438


nickvergessen:
For me it now sets the ending time only, instead of start+end when clicking in the disabled start time and selecting anything.

georgehrke:
this is weird oO

georgehrke:
@nickvergessen Can you please check again? :)

nickvergessen:
Works good, although still some minor flickering of the time (it shows 01:00 until you click anywhere and the time selector is closed)

PR #450, opened 2017-05-15
use owner-displayname if available

georgehrke:
make use of https://github.com/nextcloud/server/pull/4401


georgehrke:
ping @tcitworld :)

Issue #10, opened 2016-09-17
New concept for entering recurrence rules / full-featured recurrence settings [$155]


georgehrke:
The recurrence editor in the sidebar is currently being implemented.
It will follow the example of other calendar applications.

Although the current concept is pretty wide-spread, it's not really intuitive and we should create a concept for something more user-friendly.

This is definitely something to discuss at the next hackathon or conference.

georgehrke:
![2016_09_22_12_58_57 copy](https://cloud.githubusercontent.com/assets/1250540/18958515/23cd2f22-8663-11e6-854e-ea10342bb640.png)

Mockups created at Nextcloud conf

nickvergessen:
👍

cyphar:
Looks like this is the nextcloud version of the request owncloud/calendar#770 (which I opened before I switched to Nextcloud).

tcitworld:
Gnome Calendar just released a new UI for recurring events. https://www.youtube.com/watch?v=DoHLVmTvjFE

georgehrke:
@tcitworld Thx a lot for the link!

I really like having "Monday - Friday". We should probably have "Saturday & Sunday" too :) 
The rest is pretty similar to our concept

slideup-benni:
Why we won't support all of the RRULES defined in the ical spec?
http://www.kanzaki.com/docs/ical/rrule.html

For reference the google calendar interfaces:

![cal4](https://user-images.githubusercontent.com/12446028/29277432-ebdcc17c-8111-11e7-887a-0cacb1d12095.PNG)
![cal3](https://user-images.githubusercontent.com/12446028/29277434-ebe14e36-8111-11e7-9443-68c9abd12c67.PNG)
![cal2](https://user-images.githubusercontent.com/12446028/29277435-ebe36554-8111-11e7-9714-d2687a10e256.PNG)
![cal1](https://user-images.githubusercontent.com/12446028/29277433-ebddd3be-8111-11e7-851c-0ed5a085b70d.PNG)
![cal7](https://user-images.githubusercontent.com/12446028/29277437-ebf0d9fa-8111-11e7-9e27-7178fde9287a.PNG)
![cal6](https://user-images.githubusercontent.com/12446028/29277436-ebe9ec8a-8111-11e7-9f66-268bf5d99b2e.PNG)
![cal5](https://user-images.githubusercontent.com/12446028/29277438-ebf7b612-8111-11e7-907e-523912ecf2f6.PNG)

[...]

PR #454, opened 2017-05-17
ask whether to edit a single occurrence or all occurrences

### LINKED ISSUE #7: special dialog when exigent repeating events
When editing a repeating event the calendar needs a special dialog:

`Do you want to change only this or this and all upcoming occurrences?`

georgehrke:
fixes #7

### DIFF
--- a/js/app/models/veventModel.js
+++ b/js/app/models/veventModel.js
@@ -135,6 +135,95 @@ app.factory('VEvent', function(TimezoneService, FcEvent, SimpleEvent, ICalFactor
 			return missingTimezones;
 		};
 
+		/**
+		 * find a single VEvent subcomponent based on its recurrenceId
+		 * @param {ICAL.Time|null} recurrenceId
+		 * @throws {Error} if no matching subcomponent was found
+		 * @return {ICAL.Component|null}
+		 */
+		iface.findComponentByRecurrenceId = function(recurrenceId) {
+			const vEvents = context.comp.getAllSubcomponents('vevent');
+
+			return vEvents.find((vEvent) => {
+				const hasRecurrenceId = vEvent.hasProperty('recurrence-id');
+
+				if (hasRecurrenceId === false) {
+					return recurrenceId === null;
+				} else {
+					let vEventsRecurrenceId = vEvent.getFirstPropertyValue('recurrence-id');
+					return vEventsRecurrenceId.compare(recurrenceId) === 0;
+				}
+			}) || null;
+		};
+
+		/**
+		 * check if certain component of VEvent has more than one RRule property
+		 * @param {ICAL.Time|null} recurrenceId
+		 */
+		context.hasMultipleRRules = function(recurrenceId) {
+			const comp = context.findComponentByRecurrenceId(recurrenceId);
+			// getAllProperties always returns an array, so its safe to use length directly
+			return comp.getAllProperties('rrule').length > 1;
+		};
+
+		/**
+		 * Clones the event and sets new URI and UID and adds event
+		 * @param {ICAL.Component} event
+		 */
+		iface.fork = function(event) {
+			const uid = StringUtility.uid();
+			const uri = StringUtility.uid('Nextcloud', 'ics');
+
+			const ics = context.comp.toString();
+			const clone = VEvent.fromRawICS(context.calendar, ics, uri);
+
+			clone.comp.removeAllSubcomponents('vevent');
+			clone.comp.addSubcomponent(event);
+
+			event.updatePropertyWithValue('uid', uid);
+
+			return clone;
+		};
+
+		/**
+		 * search ExDates / RDates and list categorize in relation to another time
+		 * @param {ICAL.Time} recurrenceId
+		 * @param {ICAL.Time} delimiterRecurrenceId
+		 * @param {String} type
+		 */
+		context.sortRecurrenceDates = (recurrenceId, delimiterRecurrenceId, type) => {
+			const vEvent = context.findComponentByRecurrenceId(recurrenceId);
+			if (!vEvent) {
+				throw new Error('WTF?');
+			}
+
+			const dates = vEvent.getAllProperties(type);
+			const categorizedDates = {
+				'before': [],
+				'equal': [],
+				'after': []
+			};
+
+			dates.forEach((date) => {
+				const comparison = date.compare(delimiterRecurrenceId);
+				switch(comparison) {
+					case -1:
+						categorizedDates.before.push(date);
+						break;
+
+					case 0:
+						categorizedDates.equal.push(date);
+						break;
+
+					case 1:
+						categorizedDates.after.push(date);
+						break;
+				}
+			});
+
+			return categorizedDates;
+		};
+
 		Object.defineProperties(iface, {
 			calendar: {
 				get: function() {
@@ -224,9 +313,10 @@ app.factory('VEvent', function(TimezoneService, FcEvent, SimpleEvent, ICalFactor
 							dtstart: rawDtstart
 						});
 
-						let next;
+						let next, i=0, lastFcEvent;
 						while ((next = iterator.next())) {
 							const occurrence = iCalEvent.getOccurrenceDetails(next);
+							i++;
 
 							if (occurrence.endDate.compare(iCalStart) < 0) {
 								continue;
@@ -235,11 +325,25 @@ app.factory('VEvent', function(TimezoneService, FcEvent, SimpleEvent, ICalFactor
 								break;
 							}
 
+							console.log(occurrence);
+
 							const dtstart = context.convertTz(occurrence.startDate, timezone.jCal);
 							const dtend = context.convertTz(occurrence.endDate, timezone.jCal);
-							const fcEvent = FcEvent(iface, occurrence.item.component, dtstart, dtend);
+							const fcEvent = FcEvent(iface, occurrence.item.component, dtstart, dtend, {
+								recurring: occurrence.item.isRecurring(),
+								recurrenceId: occurrence.recurrenceId,
+								count: i + iterator.exDateInc - iterator.ruleDateInc,
+								firstOccurrence: (i === 1),
+								lastOccurrence: false
+							});
+							lastFcEvent = fcEvent;
 
 							fcEvents.push(fcEvent);
+						}
+
+						iterator.next();
+						if (iterator.complete) {
+							lastFcEvent.recurrenceDetails.lastOccurrence = true;
 						}
 					} else {
 						if (vevent) {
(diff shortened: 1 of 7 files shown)


templates/editor.recurexception-question.php
```diff
@@ -0,0 +1,18 @@
+<div>
+	<form class="events" ng-submit="">
+		<fieldset class="events--fieldset">
+			<h2><?php p($l->t('Do you want to change only this or this and all future occurrences?')); ?></h2>
```
andreasjacobsen93: What about 'Do you want to change this only or this and all future occurrences'?
Think the 'this or this', is a little confusing when reading it fast.

georgehrke: good point, let's change that :)

slideup-benni:
I updated my calendar-app to this pr, but there is no question. Neither after drag&drop nor when you click on the event and save changes. Is this PR WIP?

georgehrke:
@slideup-benni Yes, see the 2 - developing tag

a14a:
thanks for finally releasing the alpha @georgehrke ! editing a single event seems to work, but the dialog does not appear when dragging an event - instead the entire series is moved without asking, which seems counter-intuitive. to move a single event, one has to open the editor dialog for it.

georgehrke:
> but the dialog does not appear when dragging an event - instead the entire series is moved without asking, which seems counter-intuitive. To move a single event, one has to open the editor dialog for it.

Yes, it only works for the editor as of now. But the way its implemented it should be rather easy to reuse the same code for drag & drop, but one step after the other :)

a14a:
Okay, i think i've hit a couple of bugs :)

**Issue 1**
*Steps to reproduce:*
1. Create a recurring event

2. Modify a single event in the series
image

3. Move the rest of the series
image

4. Refresh the page to reveal the issue: upon refresh, the exception is no longer rendered, replaced by a standard occurrence instead
image

5. Moving the series back to original time reveals the exception again:
image


Vevents attached - the exception's DTSTART and DTEND are saved correctly, but apparently ignored.

**Issue 2**
*Steps to reproduce:*
Attempting to remove the above series produces a leftover event in the database, interface freezes, impossible to remove event; upon page reload, the calendar loads forever not unlike #495 . Had to remove manually from the database to regain access.

[vevent2.txt](https://github.com/nextcloud/calendar/files/1800352/vevent2.txt)
[vevent3.txt](https://github.com/nextcloud/calendar/files/1800353/vevent3.txt)
[vevent_leftover.txt](https://github.com/nextcloud/calendar/files/1800354/vevent_leftover.txt)

georgehrke:
Thx for testing! 

Especially issue 2 is confusing me, should have been fixed by https://github.com/nextcloud/calendar/pull/754 already. I will look into it :)

js/app/models/veventModel.js
```diff
@@ -255,13 +359,22 @@ app.factory('VEvent', function(TimezoneService, FcEvent, SimpleEvent, ICalFactor
 							const rawDtstart = dtstartProp.getFirstValue('dtstart');
 							const rawDtend = context.calculateDTEnd(singleVEvent);
 
-							const dtstart = context.convertTz(rawDtstart, timezone.jCal);
-							const dtend = context.convertTz(rawDtend, timezone.jCal);
-							const fcEvent = FcEvent(iface, singleVEvent, dtstart, dtend);
+                            const dtstart = context.convertTz(rawDtstart, timezone.jCal);
+                            const dtend = context.convertTz(rawDtend, timezone.jCal);
+                            const fcEvent = FcEvent(iface, vevent, dtstart, dtend, {
```
georgehrke: This is supposed to be `const fcEvent = FcEvent(iface, vevent, dtstart, dtend, {` instead of `const fcEvent = FcEvent(iface, singleVEvent, dtstart, dtend);`

Mistake while rebasing

georgehrke: When this is fixed, Issue 2 of @a14a should be fixed

georgehrke:
The solution for issue 1 should be pretty straight forward:
- when moving an event by drag and drop, also apply the given delta to all RecurrenceIDs of recurrence exceptions for the current series.
- The same has to be done when editing the series in the editor.

ZZMajor:
Hi, for now it is just working for editing, which seems to work pretty good for me.
I wanted to delete one event out of a series and hit "This only" and then clicked onto "Delete" but this event was not deleted, instead the series just was as before.

Another "problem" is with recurring events made by other software like Thunderbird. If I make a series in Thunderbird with the recurrence let's say every third Wednesday and click on it in the Nextcloud calendar, I got asked if I want to open All, only this or this and future, but inside the event there is just "Die Wiederholungsregel dieses Ereignisses wird noch nicht unterstützt." (German, in english something like "The rule for this repeating event ist not yet implemented). Will this be a problem for future releases, as I sometimes need such events and until now created them one by one, or is it just a problem for now?
I'm just someone who wants to have a clean calendar for the past an delete events which where yesterday or before. :-)

Nevertheless, thanks for this alpha which seems to fix some parts of the calendar I hardly missed for a long time (years). Well done!!

Greets, Jürgen

georgehrke:
> I wanted to delete one event out of a series and hit "This only" and then clicked onto "Delete" but this event was not deleted, instead the series just was as before.

That's actually supposed to work. Was it an existing event or did you create the event in Nextcloud just for testing? 

> Will this be a problem for future releases, as I sometimes need such events and until now created them one by one, or is it just a problem for now?

Yes, that's related to #10 :)

ZZMajor:
OK, got it. It was an existing recurring event which didn't work. Just made a new recurring event and tried to delete one out of the series and no problem. Seems that I have to reenter some of my events, but thats no problem. Maybe it is an event which I created outside of the calendar app with Thunderbird oder Davdroid and there was something wrong. I even tried o create an repeating event in Thunderbird and then deleted one out of the series in the calendar without problem, so all is good.

For the second part thats not that big problem. Until now I had to create single events for such events so there is no change for me and I can livee with that. Maybe I will create weekly events and then delete those who are "out of the row" via webif and all is good for now. :-)

Thanks a lot.

georgehrke:
@ZZMajor Is there any chance you can share that old event with me? This feature should obviously work on any recurring event, no matter whether its an old or a new event.

heikojansen:
Finally got around today to give this a try. 

Unfortunately I cannot report much on recurring events, because of some odd error:

- Replaced the existing up-to-date calendar app with the one downloaded from https://github.com/nextcloud/calendar/releases/download/v1.6.2-alpha-recurrence-question/calendar.tar.gz (first stopping and afterwards restarting nginx, php-fpm and mysql)
- Cleared browser cache, logged in to Nextcloud, shift-reloaded calendar app
- Having 5 calendars - holidays etc., personal one (both shared with others) and two shared by other family members; plus contact birthdays - 4 of those were loaded without problems but one shared by a family member never finished loading: spinning icon in the left sidebar never got replaced by the colored dot and the events from that calendar never showed up
- Apparently you cannot edit or add new events while a calendar resource is still considered loading, so could not test that (with recurring events)

The browser developer tools showed that (some?) data for the corresponding "REPORT" request was loaded and I could not spot any difference there to the other calendars. There 

Stopped nginx, php-fpm, mysql; replaced app with previous version, started everything, went back to the browser, reloaded - et voilà - everything back to normal: the problematic calendar showed up just fine.

Tried the whole process once again: same problem. Then back again to the previous app version and everything working fine again.

Browser console is overflowing with data in both cases. Only notable (to me) difference is that 1.6.2-ALPHA additionally showed ad ```TypeError: d is undefined```. Source maps did not work but after replacing the content of ```app.min.js``` with the un-minified code that translated to ```TypeError: lastFcEvent is undefined``` at ```app.min.js:4990:8```:

```javascript
if (iterator.complete) {
    lastFcEvent.recurrenceDetails.lastOccurrence = true;
}
```

Hope that helps in tracking down the problem.

derBobby:
One more bug?:
- 1+2: Create event Mon, 19th recurring daily for five times
- 3: Drag and drop one week further
- 4: Details for "This only" are set to the initial value
- 5: Reload of the site brings the real new value

1createlikethis
2correctincal
3draganddropped
4wrongdetails
5afterreload

derBobby:
To Issue 1 of  @a14a :
LAST-MODIFIED of the exception is not updated but the only difference. So I would guess that updating the modified timestamp for the exception should be the solution.

ZZMajor:
@georgehrke Sorry, I deleted that event totally and created a new series in the calendar directly. Since then no problem with it, but the old one is gone, no backup.
From my remembering it was an event series with past events I deleted with Thunderbird so the calendar before 1.6.1 just had the message that this rule was not implemented as told before. Maybe this was the problem.

DrMartinus:
I just downloaded it, and added a new event, made it repeat 5 times. Then I realised that I had set the ending time and wanted to change it. So I did - on the question I chose "This and all in the future", I believe (sorry...) and changed the edning time, after that the dialog whether this, all or all in the future persists on the page. It just doesn't go away.

NoobyNiceDev:
Hey guys.

I was looking for the mentioned function on the web, find this pull and tried it. Not very successfully.

I expected the bug, that you can't delete one out of some dates. But after syncing the calendar with a smartphone, the (not) deleted entries were gone on the Android Phone. Strange stuff. At Nextcloud, it's still there. May the modification is correct, but the handling of multiple occurances from the app/nextcloud-core is maybe wrong?

In Addition:
I also tried to modify the occurance of all (to set a new end of the intervall or modify the intervall) but the calendar app says something like "this intervall-rule isn't supported yet. The support will be added in further versions." Not very nice, because it's just a "one per week" intervall for araund 1.5 years... There is a second intervall, every second week but the same "this intervall-rule isn't supported yet. The support will be added in further versions." The occurances were shown well.

How could your pull be able to modify single/all occurances when the NC-core can't interpret the rule correctly?

Peterede:
The prompt to modify the event should be shown when the user clicks on update not when the editor is opened as they may not know at the time of opening if they want to change all or just one event.  The user may just want to view the details of the event so having to click this prompt each time is not ideal.

gr22:
I can confirm. I just tested and I was thankful for this essential functionality being added. Just I find it really uncomfortable that every time I want to preview a repeating event I have to face the dialog.

dsbrooks:
This doesn't seem to be working anymore after updating to  the latest NC release. I've tried replacing the calendar with these files again and nothing.

When I click "This Only" and try to delete it, the event is still there.

vahem2lu:
Any news on this one?

virtualdxs:
Is this compatible with Nextcloud 15?

georgehrke:
I will close because this pull-request is against Calendar v1 (angular), which will no longer receive new features.
This issue will be properly addressed in #926.

In #926, when clicking an event it won't open the first occurrence but properly open the occurrence you clicked. Instead of being asked upfront, you can first modify the event and will be asked on saving whether to edit this or this and all future events.

PR #457, opened 2017-05-19
Spelling: No "EMail"

comradekingu:


georgehrke:
cc @jancborchardt

juliushaertl:
I'd prefer to use "email" in english. 

See https://en.wikipedia.org/wiki/Email#Terminology on how different terms are used.

georgehrke:
It's also `Email` in the personal settings. So we should use a lowercase m for consistency reasons.

@comradekingu Can you please change it to `Email` respectively `email` :)

comradekingu:
@juliushaertl The preference for "email" is valid, as is "e-mail". However for every other language i know, abbreviation into the equivalent of "email" is in error, whereas "e-mail never is. Adherence to stricter defaults means fewer errors made in translation, as those unaware will generally just employ whatever conventions are in use. Right now Nextcloud does both.

I know it doesn't seem like much of an issue, but in the languages where it is, "email" reflects negatively upon the quality of the project and the people involved.

georgehrke:
If there is an issue with other languages (that I wasn't aware of so far), we have to communicate it with the translators. Can you specify which languages you are referring to here?

We should stick to `Email` / `email` here, because it's the correct version in english (which happens to be the source language) and it's used in all the other places of Nextcloud.
It's `Email` in:
- the personal settings
- the the sharing sidebar for files
- the contacts app
- the user's management
- the email settings in the admin panel

Everywhere, where I found `email` in the Nextcloud server and contacts app, it was `Email` / `email` and not `EMail` nor `E-Mail`. Therefore the calendar should use `Email` / `email` too for the sake of consistency.

georgehrke:
Other places:
- https://github.com/nextcloud/calendar/blob/master/js/app/filters/reminderfilter.js#L30
- https://github.com/nextcloud/calendar/blob/master/js/app/controllers/valarmcontroller.js#L55
- https://github.com/nextcloud/calendar/blob/master/templates/part.eventsattendees.php#L27

Do you mind changing those too? :)

comradekingu:
@georgehrke 

"EMail" is always wrong. I changed it from that, to "E-mail" and "e-mail" respectively.

The breakdown of the 146 strings featuring any variant, is otherwise something like
50% "email" 40% "e-mail" and 10% "mail"

For a variation over the theme valid "Email could not be sent. Check your mail server log" uses two different ones in the same string, which is assuredly wrong.

"Mail" I suppose you can get away with in non-formal English. The conjunction, if you pardon the pun, is lost on languages where e-mail and snail-mail aren't thought to be, or aren't, similar words.

It isn't far fetched for someone to translate "Invalid mail address" (which is an actual string) into "postal address"

Or just the ones that say "Mail", (which is too,) leaving the world of nouns, for a possible verb. You then have n^n errors. Going through an extra layer of possible guesswork when a user sees for example a contact field, and tries to guess which "post" to use for two similarly non-contextual fields.

As for feedback-loops, the idea of these albeit random changes, is to spur some debate where people are, and clear up some of the many internal bug-reports filed on Transifex. It is sadly not the only way it is broken as a communication-platform, and seeing as one default has to be chosen anyway, going with the one that hasn't ever caused any trouble, "e-mail" seems a safe bet.

Being asked to change a lot of strings that only differ in how they are conjugated, should sound off an alarm with anyone for whom the "-" would potentially be a problem.
There is however no accounting for that which cannot be attributed to malice…

The communication should feature when strings are frozen, and a summery of where the project is at with regards to translations, where to help out, what to look for, etc.

georgehrke:
> "EMail" is always wrong.

Neither me nor @juliushaertl are asking for `EMail`. We are asking for`Email`, which also happens to be the term required by the IETF: https://www.rfc-editor.org/materials/terms-online.txt

> It isn't far fetched for someone to translate "Invalid mail address" (which is an actual string) into "postal address"

I'm sorry, but our translators are not stupid. They are perfectly aware that Nextcloud is not able to automatically send out letters and that references to mail address is always about email.

Where ever we refer to postal address, we say so and don't use the term mail:
personal - nextcloud chromium today at 11 23 53 am

Also the contacts app never uses the term mail in context of postal address:
contacts - nextcloud chromium today at 11 26 01 am

And I can only repeat myself:
Can you please name a language where this has been an issue yet? 

I don't know of any. That's why I don't see a reason to use a wrong english version.

If there is really a need to give hints to translators, we can simple add comments to the source:
https://docs.nextcloud.com/server/9/developer_manual/app/l10n.html#hints

comradekingu:
@georgehrke 

It is only for RFCs

These:

https://tools.ietf.org/html/rfc1711
Classifications in E-mail Routing

https://tools.ietf.org/html/rfc2110
MIME E-mail Encapsulation of Aggregate Documents, such as HTML (MHTML)

Simple Mail Transfer Protocol
https://tools.ietf.org/html/rfc821
https://tools.ietf.org/html/rfc2821

https://tools.ietf.org/html/rfc4405
SMTP Service Extension for Indicating the Responsible Submitter of an E-Mail Message

https://tools.ietf.org/html/rfc4406
Sender ID: Authenticating E-Mail

https://tools.ietf.org/html/rfc4407
Purported Responsible Address in E-Mail Messages

https://tools.ietf.org/html/rfc4021
Registration of Mail and MIME Header Fields

https://tools.ietf.org/html/rfc6109
La Posta Elettronica Certificata - Italian Certified Electronic Mail

And on and on… Consistency is not easy.

And to the effect of being a laudable and expressed goal, the resource has this in it:

"o As used in"
"o as it appears in"

What about the IETF page itself?

Here is the electronic mailing list they use:
https://www.ietf.org/mailman/listinfo/6lo
Maybe that is all Mailman, but: 

"email" 
"e-mail"

So is it consistent within a page?
https://www.ietf.org/registration/ietf99/sendconfirmno.py

No.

Maybe IETF 100 will be the one.

So what about the actual RFC styleguide for RFCs:

https://www.rfc-editor.org/rfc/rfc7322.txt

"email"
"EMail"

I give this effort 0/1 soups as a resource for how to write "e-mail".

I shall not be condencended upon by such high-horsery as to not be called stupid,
by someone who does not understand my stupidity as well as me.

As a translator, (aha!) I found myself flagging issues, only to be asked not to.
And when looking for someone to fix them, I noticed that person was me.
Luckily it was all uphill from there.


https://github.com/nextcloud/server/blob/9d8e9b1efbc210319ce467927085405a7219a2c2/apps/sharebymail/l10n/pl.json

I went through the first two, out of 19 pages containing "mail" alone.
The real number is 164 * 61 languages, with 19 more requested.

By 746 different translators.

https://github.com/nextcloud/server/blob/9d8e9b1efbc210319ce467927085405a7219a2c2/apps/sharebymail/l10n/pl.json

"Password for mail share sent to {email}" : "Hasło do współdzielenia przez mail zostało wysłane do {email}",

If you wanted to do "mail" you would do mejl. If the translator thinks "mail" will do, where would one find people not
aware of this connection? People that need Polish translations…


https://github.com/nextcloud/server/blob/9d8e9b1efbc210319ce467927085405a7219a2c2/apps/sharebymail/l10n/pt_BR.json

"Password for mail share sent to {email}" : "Senha para o correio compartilhado enviado para {email}",

(Password for the post share sent via)  Curiously this is a part of the world where sneakernets are popular.


https://github.com/nextcloud/server/blob/16091f4424b44d9d8ab6076f1cf9218b1d25da4d/apps/sharebymail/l10n/it.js

"You shared %1$s with %2$s by mail" : "Hai condiviso %1$s con %2$s tramite posta",
"You shared {file} with {email} by mail" : "Hai condiviso {file} con {user} tramite posta",
"%3$s shared %1$s with %2$s by mail" : "%3$s ha condiviso %1$s con %2$s tramite email",

Willing to bet this is the work of two different translators.


https://github.com/nextcloud/server/blob/9d8e9b1efbc210319ce467927085405a7219a2c2/apps/sharebymail/l10n/fr.json

"Password for mail share sent to {email}" : "Le mot de passe pour le partage par email a été envoyé à {email}",

This is wrong on many levels. It is not french, and if it was, it wouldn't be "email".


https://github.com/nextcloud/server/blob/b1bd9ff0b95af68c9f5875de00806e37ef65a8b8/apps/sharebymail/l10n/ast.json

 "Password for mail share sent to {email}" : "Unvióse a {email} la contraseña pa la compartición per corréu",

Same word as "correo" (post) as spanish. This is activly saying the password is sent in the mail.

The alternatives were Corréu electrónicu, or "e-mail"


So the word for "mail" isn't there in other indo-european languages. If you take it to mean just "post", without anything else, you have written: "post".

In English just "mail" is (now) both, but without any context, if and when is does "mailbox" or "mailing address" relates to electronic or regular mail?

Yes, mailbox is both, but mailing address isn't.

And once you have vague input, you can try to leave it at that, or nail it down.

The unclear "mailbox", for reasons one could happen upon the wrong one, would distinctly be "letterbox". Norwegian Bokmål (postkasse)

A word like "Post-office box" is clearly only postal, but that doesn't help you, because lets say you didn't know, and didn't care to check.

It is a semi-easy mistake to make. Mailbox - ~Postbox  Norwegian Bokmål (postboks), German (postfach).

Norway doesn't really have post-offices anymore, nor do regular people have a PO-box.

So if box is "boks" when talking about a "Post-office box", what do you add onto post to get "Mailbox" as in letterbox?

That would be a "postkasse". Or a "post"-"crate". Because we have the standalone boxen, not the door-hatch type contraptions.

Lets say you envisioned one of those, seeing as it was an English resource. It would be brev-sprekke.

If you send it through the hatch (the part you actually open it with) to get it into the letter-box, you will find yourself right at home with "sprekke" being crack, as in slit.

Go from German "fach" to swedish "ficka", and it could not only be a slot/pouch, but a mail type bag alone.

As an example, it shows that with enough vagueness, you encourage vagueness, with possible or resulting misconceptions, or flat out errors.

With multi-tiered translations, and every user there is, what could have been easy, is now ugly, wrong, misleading, or a combination thereof.

jancborchardt:
Folks, let's please not waste our time with rather less relevant stuff like this.

**Email** is the widely-used spelling as also used by Wikipedia. E-Mail/e-mail is outdated, and EMail (capital M) simply looks ugly.

Please let's use it accordingly and then work on more fulfilling tasks than a looong discussion on spelling. ;)

comradekingu:
@jancborchardt 
First, [no](https://books.google.com/ngrams/graph?content=email%2C+e-mail&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cemail%3B%2Cc0%3B.t1%3B%2Ce%20-%20mail%3B%2Cc0).

I would advice you to look at how e-mail is spelt on Wikipedia-resources other than English.
My point is you end up with errors in translation if you employ "email".

The point gets across in other languages, but not without a severe penalty to the esteem of the source.

jancborchardt:
Here's a comparison of search volume, which I consider common usage: https://trends.google.com/trends/explore?q=e-mail,email

Since it's an invented word anyway, the usage which is more common and easy should be employed. Which is Email. Chicago Manual of Style also updated recently, which was also referred in the Grammar Girl article you linked: http://www.quickanddirtytips.com/education/grammar/chicago-updates-stop-capitalizing-internet-and-hyphenating-email

comradekingu:
@jancborchardt Having a special sloppy exception is not easy. Which is why the predilection for which, makes English a complicated and inconsistent language.

Consider
https://trends.google.com/trends/explore?q=G-string,gstring
https://trends.google.com/trends/explore?geo=US&q=G-string,gstring
https://trends.google.com/trends/explore?geo=GB&q=G-string,gstring

I don't write "e-mail" when searching for things, because I don't have a reason not to.

https://trends.google.com/trends/explore?date=all&q=e-mailed,mailed,emailed

So that is where we are headed with English. "Mailpile" has already done away with the "e" in both in title and content.

Meanwhile, I do prefer "e-mail" for use in software, because that is where translations come from.

georgehrke:
> Meanwhile, I do prefer "e-mail" for use in software, because that is where translations come from.

To repeat myself a third time for now. Has there been actual issues with translations yet? I'm not aware of any. Please tell me the affected languages.

And even if we had issues with translations so far, we can give hints to our translators as already mentioned in https://github.com/nextcloud/calendar/pull/457#issuecomment-302925790

georgehrke:
superseded by https://github.com/nextcloud/calendar/pull/469

comradekingu:
@georgehrke 

This isn't my idea of something that would be really fun to make up. Yes.

Failed to send share by E-mail
Feilet når delingen skulle sendes på epost
1022
Failed to create the E-mail
Feilet ved opprettelse av epost
1840
To send a password link to the user an email address is required.
Krever epost for å sende link til bruker.
1939
Contact us by email
Kontakt oss på epost
2752
Failed to send Email: %@
Sending av epost feilet: %@
3041
Author's e-mail
Forfatterens epostadresse
3588
Please add your email address in the personal settings in order to add attendees.
Legg inn epost-adressen din i personlige innstillinger for å kunne legge til deltakere.

Last one here is actually me. I messed up the same way this person did:

https://github.com/nextcloud/server/blob/b4f36d44c43aac0efdc6c70ff8e46473341a9bfe/apps/user_ldap/l10n/de_AT.json
"Email Field" : "Email-Feld",

A mistake easily made:

https://github.com/nextcloud/server/blob/d41fd7471c05ccaff14d13ed1a8d721dc363b8b0/settings/l10n/nn_NO.js
 "Your email address" : "Di epost-adresse",

Not sure if this is a compound-error, but still:

https://github.com/nextcloud/server/blob/b4f36d44c43aac0efdc6c70ff8e46473341a9bfe/apps/user_ldap/l10n/eu.json
"Email Field" : "Eposta eremua",

**This can be avoided:**

https://github.com/nextcloud/server/blob/9c8a8386240d440e8f7363d85dd3ead2207b7a60/core/l10n/sv.js
"Share with users, groups or by mail..." : "Dela med användare, grupper eller via epost",

https://github.com/nextcloud/server/blob/6b490f45fd938d631fc524e0c04af43182056bb7/apps/user_ldap/l10n/pt_PT.json
"Email Field" : "Campo de email",

https://github.com/nextcloud/server/blob/d41fd7471c05ccaff14d13ed1a8d721dc363b8b0/settings/l10n/bs.json
"You need to set your user email before being able to send test emails." : "Prije nego li ste u mogućnosti slati testnu email trebate postaviti svoj korisnički email.",

https://github.com/nextcloud/server/blob/7d1f362aa030b0b279067da3a05105890bd5d194/apps/user_ldap/l10n/sq.json
 "LDAP / AD Email Address:" : "Adresë Email LDAP / AD:",

https://github.com/nextcloud/server/blob/462f12611c059b619881fe972e06d0eb175c1af2/apps/user_ldap/l10n/pl.json
"LDAP / AD Email Address:" : "Adres email LDAP/AD:",


Maybe this is helpful, but it isn't correct:

https://github.com/nextcloud/server/blob/a442c86a8465a27fab84837d2fc40f2211de06b8/core/l10n/ro.json
Please contact your administrator." : "Nu a putut fi trimis un email pentru resetare deoarece nu există o adresă email pentru acest utilizator. Contactează-ți administratorul.",

**Wrong type of mail, could be avoided:**

https://github.com/nextcloud/server/blob/462f12611c059b619881fe972e06d0eb175c1af2/settings/l10n/sl.js
"mail" : "pošta",

https://github.com/nextcloud/server/blob/462f12611c059b619881fe972e06d0eb175c1af2/settings/l10n/sl.json
"mail" : "pošta",

Maybe this person would remember the dash. If you read this with a screen-reader, it becomes "enamel"

https://github.com/nextcloud/server/blob/7d1f362aa030b0b279067da3a05105890bd5d194/lib/l10n/de_DE.json
"This is an automatically sent email, please do not reply." : "Dies ist eine automatisch versendete EMail, bitte nicht antworten.",

This is just wrong, but maybe dumping a lot of strings requiring corrections would help:

https://github.com/nextcloud/server/blob/7d1f362aa030b0b279067da3a05105890bd5d194/lib/l10n/es.json
"This is an automatically sent email, please do not reply." : "Este es un correo enviado automáticamente, por favor no responda.",

The first 7 strings have been corrected. I don't have access to touch the rest on the translation platform. This is a very small portion of the total amount of errors induced by "email" in English, as opposed to "e-mail".

comradekingu:
So the question remains, what is the tally for cases where the source says "e-mail": 2

Sample-group: All 22 pages of 
https://github.com/nextcloud/server/search?p=22&q=e-mail&type=&utf8=%E2%9C%93

https://github.com/nextcloud/server/blob/d41fd7471c05ccaff14d13ed1a8d721dc363b8b0/settings/l10n/sq.json
E-Mail" : "Email",

https://github.com/nextcloud/server/blob/d41fd7471c05ccaff14d13ed1a8d721dc363b8b0/settings/l10n/sq.js
 "E-Mail" : "Email",

At-least it is consistently (from what i can tell, email, in this resource)

georgehrke:
> Dies ist eine automatisch versendete EMail, bitte nicht antworten.

This is perfectly fine. What it's supposed to be in your opinion?

"Dies ist eine automatisch versendete Netzpost, bitte nicht antworten."? ;)

comradekingu:
@georgehrke That would just be quirky.
"E-mail"
https://de.wikipedia.org/wiki/E-Mail#Artikel_und_Schreibweisen

> 

Gemäß Duden, Wahrig und dem amtlichen Wörterverzeichnis der reformierten deutschen Rechtschreibung[4] ist E-Mail die einzig richtige Schreibweise.[5][6] Daneben sind jedoch auch die amtlich inkorrekten Schreibweisen Email (Verwechslungsgefahr mit dem Schmelzüberzug Email/Emaille!), email sowie eMail, e-Mail, e-mail, E-mail oder einfach Mail bzw. mail verbreitet.

georgehrke:
It might be E-mail in german, but it's email in english and english is the source language. 
If you want translators to use E-mail in the german translation, tell them, but let's not break the english version. Report an issue on transifex.

Let's end this discussion here. 

The term email affects almost all apps + the server. If you want to properly discuss this, please go to https://help.nextcloud.com/c/dev.
Calendar is the wrong place for this discussion

comradekingu:
@georgehrke "e-mail" is not wrong in English, it is correct. And by that I don't mean to say that "email" isn't too.

I never envisioned it could be this hard to get some leverage on changing it.

comradekingu:
As requested, issue raised on https://help.nextcloud.com/c/dev

https://help.nextcloud.com/t/e-mail-or-email/13051

PR #472, opened 2017-05-25
Language: Removed redundant "successfully"

comradekingu:


jancborchardt:
Good call – but it would also be good to add context, like what was imported:

> Imported calendar **Personal**

or

> Imported event **Open Source Design Summit**

comradekingu:
@jancborchardt Add that if you want. I would approve, albeit maybe not truly ideal, the potential for error would be on the side of caution by specifying.

jancborchardt:
@georgehrke or someone of @nextcloud/calendar can you add the calendar/event name in this feedback message?

georgehrke:
I can take care, but this is not as trivial as it might sound. 
Something for 1.7.0

georgehrke:
Needs a proper implementation in the vue app + cdav-library

PR #473, opened 2017-05-25
Spelling: iframe

comradekingu:
It used to be IFRAME https://www.w3.org/TR/1999/REC-html401-19991224/present/frames.html#h-16.5
for reasons unknown, but now it is "iframe"
https://www.w3.org/TR/2014/REC-html5-20141028/embedded-content-0.html#the-iframe-element
Added something to start of sentence to avoid issue with Iframe or iframe.

### DIFF
--- a/templates/part.publicinformations.php
+++ b/templates/part.publicinformations.php
@@ -45,7 +45,7 @@
 		<fieldset class="settings-fieldset">
 			<ul class="settings-fieldset-interior">
 				<li class="settings-fieldset-interior-item">
-					<label><?php p($l->t('IFrame to integrate')); ?></label>
+					<label><?php p($l->t('Iframe to integrate')); ?></label>
 			    <textarea class="integration-code"
 			      type="text"
 			      ng-value="integration(item)"


georgehrke:
> https://www.w3.org/TR/2014/REC-html5-20141028/embedded-content-0.html#the-iframe-element

That refers to the tag name of iframes in HTML, not the spelling in everyday use.
The correct form (imho) is iFrame. As it is the first word of the sentence, we capitalize the first letter 
-> IFrame

georgehrke:
👎

comradekingu:
@georgehrke That is what I was trying to reference, without going to the length of adding -tags.
I don't understand where the "F" comes from.
https://www.w3schools.com/html/html_iframe.asp

tcitworld:
Same opinion. :-1: 

> The correct form (imho) is iFrame.

Does it work only on Safari ? :p 

> https://www.w3schools.com/html/html_iframe.asp

You really shouldn't use this website as a reference.

jancborchardt:
https://en.wikipedia.org/wiki/HTML_element#Frames – »iframe« is used there in the body text too. :)

georgehrke:
@jancborchardt So make it »Iframe to integrate«?

The »put in« seems weird to me

comradekingu:
@tcitworld Mistook it for w3c.org
I don't know. "iframe" is possibly correct, it may be "Iframe", but certainly not "IFrame".
@georgehrke Enclose in iframe tags/element maybe?

georgehrke:
>  "iframe" is possibly correct, it may be "Iframe",

As we capitalize the first letter because its at the start of the sentence it will be `Iframe` in both cases.

> Enclose in iframe tags/element maybe?

I still don't get why you want to add something. `Iframe to integrate` would be perfectly fine in my opinion. Can you elaborate please?

comradekingu:
@georgehrke I didn't think of that. Will do.

georgehrke:
let me squash and rebase

PR #474, opened 2017-05-25
Language: Shouting at user (!) removed

comradekingu:


georgehrke:
It shows an error and we can't proceed when the user doesn't select a calendar. I don't see why an exclamation mark is wrong here. 

cc @jancborchardt

georgehrke:
exclamation mark is frankly is not shouting.
»SELECT A GODDAMN CALENDAR ALREADY!!11« would be

tcitworld:
More seriously though, let's ask @jancborchardt or someone else to write something on phrasing in Nextcloud so we can refer to it.

comradekingu:
@georgehrke While I agree, that is not English.
Can you device a method of shouting that states the calender should be selected, without swearing, omitting the use of an exclamation mark?

Since this is difficult, or requires additional wording, doing away with the "!" in favour of "." is an easy fix that can not at the very least be mistaken for anything else.

jancborchardt:
The removal of the exclamation mark is good. :) Much more friendly. Not sure about changing »select« to »pick« though, because we use »select« everywhere else I think.

georgehrke:
@jancborchardt What about https://github.com/nextcloud/calendar/pull/474/files#diff-1a8e50adcc0728c0a7a104dab9e4179aL127 and https://github.com/nextcloud/calendar/pull/474/files#diff-1a8e50adcc0728c0a7a104dab9e4179aL135 ?

comradekingu:
@jancborchardt I imagined it to be a pick-and-done type of thing, rather than a select-between-many-and-do-something-else ordeal.

jancborchardt:
@comradekingu ok, sounds good. Can you fix the two other occurences @georgehrke listed?

georgehrke:
@comradekingu Can you please squash the commits?

js/app/controllers/editorcontroller.js
```diff
@@ -124,15 +124,15 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
...
 				error = true;
 			}
 			if ($scope.calendar === null || typeof $scope.calendar === 'undefined') {
-				OC.Notification.showTemporary(t('calendar', 'Please select a calendar!'));
+				OC.Notification.showTemporary(t('calendar', 'Please pick a calendar.'));
 				error = true;
 			}
 			if (!$scope.properties.checkDtStartBeforeDtEnd()) {
-				OC.Notification.showTemporary(t('calendar', 'The event ends before it starts!'));
+				OC.Notification.showTemporary(t('calendar', 'The event can not end before it starts.'));
```
georgehrke: What about `may not end`? 

@jancborchardt

js/app/controllers/editorcontroller.js
```diff
@@ -124,15 +124,15 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
 		$scope.validate = function() {
 			var error = false;
 			if ($scope.properties.summary === null || $scope.properties.summary.value.trim() === '') {
-				OC.Notification.showTemporary(t('calendar', 'Please add a title!'));
+				OC.Notification.showTemporary(t('calendar', 'Please add a title.'));
 				error = true;
 			}
 			if ($scope.calendar === null || typeof $scope.calendar === 'undefined') {
-				OC.Notification.showTemporary(t('calendar', 'Please select a calendar!'));
+				OC.Notification.showTemporary(t('calendar', 'Please pick a calendar.'));
```
georgehrke: Please revert this to `select`

js/app/controllers/editorcontroller.js
comradekingu: I think it is a, while polite, way that makes it harder for people whom are not English to understand it.
Entertain the idea that "may not" could mean it is not 100% clear cut that the event ends before it starts. As in maybe. And yes, into context you could derive that this is not the intention, but it makes it possible to get into that sort of thinking.

jancborchardt: Yeah, "can not" seems better.

Even better would be though if the dates would always be adjusted automatically. When you adjust the start date beyond the end date, move the end date accordingly. And vice versa, when you modify the end date to be before the start date, adjist the start date. This message should never ever need to be shown. :)

comradekingu: If it just lights up in red whenever it is not a valid span, it would be more intuitive.

georgehrke: > Yeah, "can not" seems better.

Hm. ok. I still think @comradekingu underestimates our translators ;) 

> When you adjust the start date beyond the end date, move the end date accordingly.

The end date is currently always adjusted when changing the start date :)

So about changing the end date?
What if a user changes the end to something before start?
1. change the start time to the end time?
or
2. change the start time to the end time minus one hour?

georgehrke:
Let's merge after the tests passed

js/app/controllers/editorcontroller.js
jancborchardt: Yep, change the start time to end time minus one hour. Or for a full-day event of course make it a one-day event.

georgehrke: Okay :)

Will create an issue for it

georgehrke: https://github.com/nextcloud/calendar/issues/497

PR #475, opened 2017-05-26
show displayname instead of userid after creating a new user share

### LINKED ISSUE #459: When creating a new share, it shows userid instead of display name in list
refresh is needed to fix it.

georgehrke:
fixes #459 

(also empties the search field after creating new share)


brantje:
:+1: Looks good. If you like i can approve the PR.

PR #476, opened 2017-05-26
use NC12 IEMailTemplate for fancier emails

### LINKED ISSUE #427: Better designed emails when sharing calendar
Today we use standard templates that contains mostly text. It works, but isn't that pretty. What I suggest is to improve the design so that it looks more professional. I've seen some great work by @Espina2 the past weeks, and I'd really like to see some suggestions regarding an improvment of the mails that are sent to users.

As an external user of Nextcloud (someone that uses Nextcloud sends a link to someone that doesn't) the first impression is the mail they get with either a calendar invitavion or a shared link. To attract more users to use Nextcloud and make a switch from Google or some other provider I belive it's crucial that our soulution looks better.

Google Calendar invitation:
markering_002

The server emails are merged by @MorrisJobke and now it's only the calendar emails left.. It currently looks like this in Thunderbird:
calendar

It would be great if that also was merged before 12 is release so that we can close this issue.

Thank you! :)

Copy pasted from here: https://github.com/nextcloud/server/issues/2071

georgehrke:
fixes #427 

email

I didn't find a way to make the calendar name bold yet :/

### DIFF
--- a/controller/emailcontroller.php
+++ b/controller/emailcontroller.php
@@ -85,16 +85,56 @@ class EmailController extends Controller {
 	 * @NoAdminRequired
 	 */
 	public function sendEmailPublicLink($to, $url, $name) {
-
 		$user = $this->userSession->getUser();
 		$username = $user->getDisplayName();
 
 		$subject = $this->l10n->t('%s has published the calendar "%s"', [$username, $name]);
 
-		$emailTemplateHTML = new TemplateResponse('calendar', 'mail.publication.html', ['subject' => $subject, 'username' => $username, 'calendarname' => $name, 'calendarurl' => $url, 'defaults' => $this->defaults], 'public');
-		$bodyHTML = $emailTemplateHTML->render();
-		$emailTemplateText = new TemplateResponse('calendar', 'mail.publication.text', ['subject' => $subject, 'username' => $username, 'calendarname' => $name, 'calendarurl' => $url], 'blank');
-		$textBody = $emailTemplateText->render();
+		$serverVersion = $this->config->getSystemValue('version');
+		if (version_compare($serverVersion, '12', '>=')) {
+			$emailTemplate = $this->mailer->createEMailTemplate();
+
+			$emailTemplate->addHeader();
+			$emailTemplate->addHeading($this->l10n->t('%s has published the calendar %s', [$username, $name]));
+
+			$emailTemplate->addBodyText($this->l10n->t('Hello,'));
+
+			$htmlText = str_replace(
+				['{boldstart}', '{boldend}'],
+				['<b>', '</b>'],
+				$this->l10n->t('We wanted to inform you that %s has published the calendar {boldstart}%s{boldend}.', [$username, $name])
+			);
+
+			$plainText = $this->l10n->t('We wanted to inform you that %s has published the calendar %s.', [$username, $name]);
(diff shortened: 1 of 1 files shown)


georgehrke:
I'm thinking about adding »« like in the files app

inbox ehrke email 5 238 messages mail today at 5 17 46 pm

tcitworld:
> I'm thinking about adding »« like in the files app

Makes it quite strange IMHO, as a French, we're not used to see this since we use french quotes « citation » just like regular quotes. Maybe we can leave this up to the translators to choose.

enoch85:
Cc @espina2

enoch85:
What I miss when invited to an event:
Accept | Decline | Maybe (if maybe is an option?)

Also; information about who invited you, which calendar it is (name), and place of the event.

The Google pics describes it pretty good in this issue: https://github.com/nextcloud/calendar/issues/427 (Even though the red is not so nice ;) )

georgehrke:
@enoch85 Server issue ;)

enoch85:
@georgehrke >  Server issue ;)

So what you mean is that the DAV endpoint doesn't support it? :open_mouth:

georgehrke:
> So what you mean is that the DAV endpoint doesn't support it?

The dav end point the part of the server ;)

enoch85:
I may be mistaken, but it should be possible to extract the info that you can type in, because you can add info about:
- Attendee
- Time 
- Place
- Who invited you is just about showing the name of the sender which could be extacted from the Display name (right?)

Or do you mean that `Accept | Decline | Maybe (if maybe is an option?)` is not possible?

georgehrke:
Yes, that's possible. But not the calendar app but the server sends out the invitation emails ;)

enoch85:
Aah, I should go to bed now :)

enoch85:
Design wise :+1:

Espina2:
@georgehrke @enoch85 if you want to highlight something just use a plain bold. It should be enough.

georgehrke:
Hm, sorry for this stupid question, but what do you mean with plain bold? 

`<b> </b>` doesn't work

enoch85:
@Espina2 

> Hm, sorry for this stupid question, but what do you mean with plain bold?

Ping

Espina2:
```<strong>``` is the html default markup for bold

georgehrke:
`<strong>` will be escaped just like `<b>`

georgehrke:
check the first screenshot. It shows `<b>Geburtstage von Freunden</b>`.

If I use `<strong>`, it will just be `<strong>Geburtstage von Freunden</strong>`.

georgehrke:
I don't seem to find a way to add bold text looking thru the [EMailTemplate.php](https://github.com/nextcloud/server/blob/master/lib/private/Mail/EMailTemplate.php)

Let's go with »« for now. Translators will have the chance to replace that with "" or any other local standard.

enoch85:
> Let's go with »« for now. Translators will have the chance to replace that with "" or any other local standard.

Let's do this! :rocket:

georgehrke:
No No No 🙈

enoch85:
Oops :D

georgehrke:
@enoch85 Please don't merge stuff with "2 - developing" ;)

enoch85:
Aiai sir :+1:

georgehrke:
I will just open a follow up PR

PR #480, opened 2017-05-29
fix nighly builds

georgehrke:


georgehrke:
- We removed autosize when we dropped support for nextcloud 9 and 10
- timezones folder was replaced by timezones.json

PR #483, opened 2017-05-30
nc12 emailtemplate follow up

georgehrke:


georgehrke:
follow up for https://github.com/nextcloud/calendar/pull/476

georgehrke:
screenshot

enoch85:
:+1:

jancborchardt:
The body text completely duplicates the title though … this is not really necessary, and we should simplify it.

georgehrke:
@jancborchardt Can you suggest a different body text? :) 

I'd be happy to change it

jancborchardt:
I would just show the header and the button, nothing more. :)

Issue #509, opened 2017-06-15
NC11: header right buttons on public shared calender are shown on the left


michag86:
### Steps to reproduce
1. share a calender by link
2. open the shared calendar

![grafik](https://user-images.githubusercontent.com/6691165/27181547-47f5d3ea-51d8-11e7-9f69-4ed03af2f4f1.png)

Only NC11 is affected. If you add this, it works:
![grafik](https://user-images.githubusercontent.com/6691165/27181621-a1030048-51d8-11e7-83b4-c60498ef041f.png)

### Expected behaviour
Subscribe and Download button are shown right

### Actual behaviour
Subscribe and Download button are shown left. Both are not usable!

### Server configuration
**Operating system**: Linux ekaptest.thlv.de 3.10.0-514.6.1.el7.x86_64 #1 SMP Wed Jan 18 13:06:36 UTC 2017 x86_64

**Web server:** Apache/2.4.6 (CentOS) OpenSSL/1.0.1e-fips PHP/7.0.10 (apache2handler)

**Database:** mysql 5.5.52

**PHP version:** 7.0.10
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, session, standard, apache2handler, bz2, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, json, ldap, exif, mysqlnd, PDO, Phar, posix, shmop, SimpleXML, sockets, sqlite3, sysvmsg, sysvsem, sysvshm, tokenizer, xml, xmlwriter, xsl, zip, mysqli, pdo_mysql, pdo_sqlite, wddx, xmlreader

**Nextcloud version:** 11.0.3 (stable) - 11.0.3.2

**Updated from an older Nextcloud/ownCloud or fresh install:**

**Where did you install Nextcloud from:** 

**List of activated apps:**
<details>
	<summary>App list</summary>
```
Enabled:
 - activity: 2.4.1
 - calendar: 1.5.3
 - comments: 1.1.0
 - dav: 1.1.1
 - deck: 0.1.4
 - federatedfilesharing: 1.1.1
 - federation: 1.1.1
 - files: 1.6.1
 - files_pdfviewer: 1.0.1
 - files_sharing: 1.1.1
 - files_texteditor: 2.2
 - files_trashbin: 1.1.0
 - files_versions: 1.4.0
 - files_videoplayer: 1.0.0
 - gallery: 16.0.0
[...]

PR #510, opened 2017-06-15
Fix for nc11 (public share header buttons)

michag86:
Fix for public shared calender (header buttons on the right side) #509


michag86:
Tested on NC 11 and NC12

georgehrke:
Thx for the pull request! :)

tcitworld:
@georgehrke Maybe release this as 1.5.4 ?

georgehrke:
yes, we should do a 1.5.4 indeed
I don't see 1.6.0 before end of September :/ 

I will prepare a release in the next couple of days.

Issue #513, opened 2017-06-19
Highlight current day in small calendar in left navigation bar too


jancborchardt:
The small calendar in the left navigation bar should have the current day highlighted too, and ideally the weekend too.

![screenshot from 2017-06-19 19-41-46](https://user-images.githubusercontent.com/925062/27298055-7ea62a3c-5527-11e7-9641-956139311ced.png)

Should be done the same way as for the main calendar: light yellow for current day, grey for weekend.

cc @nextcloud/calendar

georgehrke:
Was implemented with #533

PR #533, opened 2017-07-10
Add colors datepicker, focus day when navigating

tbille:
**Proposition for this issue:** https://github.com/nextcloud/calendar/issues/513

Add grey color to the week-ends and yellow color to the current day (today).

selection_008

**Bug fix:** when changing month the selected day was the same as the current day. For example if the current day is 10th of July, when I changed month it was focusing the 10th of August.

This is fixed, now the first day of the month and the week is selected.

**Tests:** This pull-requests adds the missing tests on the DatepickerController.

### DIFF
--- a/css/app/datepicker.css
+++ b/css/app/datepicker.css
@@ -89,12 +89,13 @@
 	border-left: 0;
 }
 
-#datepicker table tbody tr.uib-weeks td:last-child button {
-	border-right: none;
+.highlight-today {
+	border-radius: 50%;
+	background-color: #e6e6e6;
 }
 
-#datepicker table tbody tr.uib-weeks:last-child td button {
-	border-bottom: none;
+#datepicker table tbody tr td button {
+	border: none;
 }
 
 #datepicker table tbody button {
@@ -102,7 +103,10 @@
 }
 
 #datepicker table tbody button.active {
-	background-color: rgba(240, 240, 240, .9);
+	border-radius: 50%;
+	background-color: #0082c9;
+	color: #fff;
+	font-weight: bold;
 }
 
 #datepicker table tbody button span {
@@ -110,6 +114,10 @@
 }
 
 #datepicker .text-muted {
+	color: #c7c7c7;
+}
+
+.highlight-weekend button {
 	color: #aaa;
 }
 
@@ -122,6 +130,12 @@
 	display: flex;
 	margin: 0 5px;
 	width: calc(100% - 10px);
+}
+
+#app-navigation .datepicker-heading .button {
+	background-color: #fff;
+	border: none;
+	font-weight: normal;
 }
 
 #app-navigation .togglebuttons .button {
--- a/js/app/controllers/datepickercontroller.js
+++ b/js/app/controllers/datepickercontroller.js
@@ -29,22 +29,22 @@ app.controller('DatePickerController', ['$scope', 'fc', 'uibDatepickerConfig', '
 	function ($scope, fc, uibDatepickerConfig, constants) {
 		'use strict';
 
-		$scope.datepickerOptions = {
-			formatDay: 'd'
-		};
-
-		function getStepSizeFromView() {
-			switch($scope.selectedView) {
-				case 'agendaDay':
-					return 'day';
-
-				case 'agendaWeek':
-					return 'week';
-
-				case 'month':
-					return 'month';
+		function getDayClass(data) {
+			if (moment(data.date).isSame(new Date(), 'day')) {
+				return 'highlight-today';
 			}
+
+			if (data.date.getDay() === 0 || data.date.getDay() === 6) {
+				return 'highlight-weekend';
+			}
+
+			return '';
 		}
+
+		$scope.datepickerOptions = {
+			formatDay: 'd',
+			customClass: getDayClass
+		};
 
 		$scope.dt = new Date();
 		$scope.visibility = false;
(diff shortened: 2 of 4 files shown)


georgehrke:
@tbille Thx a lot for this pull request! :) 

Looks really cool! 🎉 :) 

> Add grey color to the week-ends and yellow color to the current day (today).

👍 As discussed in the issue

> when changing month the selected day was the same as the current day.

This was actually the expected behavior for me, but let's open this up for discussion.
cc @jancborchardt @tcitworld 

> This pull-requests adds the missing tests on the DatepickerController.

Awesome! Much appreciated.

css/app/datepicker.css
```diff
@@ -89,7 +89,16 @@
...
+.todayHilight {
+	background-color: rgb(255, 255, 170);
+}
+
+#datepicker table tbody tr.uib-weeks td:first-child button {
+	background-color: rgba(240, 240, 240, .9);
+}
+
 #datepicker table tbody tr.uib-weeks td:last-child button {
+	background-color: rgba(240, 240, 240, .9);
```
georgehrke: I didn't test this yet, but I think this will only work for US. What if you set your locale e.g. to german, where the week starts on monday? I guess Monday and Sunday will be shown a little darker

tbille: true, my work environment is set in US. This should be changed to work also with an European environment.

georgehrke: I'm not sure this approach will work for both locales.

Maybe we have to add another class to the date picker element, that says whether the week starts on Sunday or Monday.

tbille: ok it can be done in the function `getDayClass` of the datepickerController

tbille: done, now a class is added for everyday weekend days: `weekendHilight`

jancborchardt:
> This was actually the expected behavior for me, but let's open this up for discussion.

Yeah, I’d also say it’s expected behavior.

Great work @tbille! :)

jancborchardt:
Just one point of feedback: In that datepicker there instead of giving the weekend full grey background, I would rather grey out the numbers. Otherwise it’s way too much visual load. :)

css/app/datepicker.css
```diff
@@ -89,6 +89,14 @@
 	border-left: 0;
 }
 
+.todayHilight {
```
jancborchardt: Also please use class naming like: `.highlight-today` and `.highlight-weekend`. :)

jancborchardt:
I was working on a new datepicker design and since I made sure it works with the calendar datepicker too, how about we use the same design there?

See https://github.com/nextcloud/server/issues/5691#issuecomment-314796241
datepicker (circles should be round of course, it’s a rough mockup ;)

tbille:
@jancborchardt I applied your mockup on the datepicker, here how it looks like: 

When not visible:
selection_009

When visible:
selection_010

Shall we apply the same style for the buttons month/weeks/day?

jancborchardt:
Very nice! Looks good and I'd merge it. Cc @pixelipo since he just works on the date picker in core.

The week/month/today buttons should stay like this. :)

pixelipo:
Looks great! I've only got two comments:
- header should not be bold (same as core version)
- today date would look better (and have better contrast) with a white font

jancborchardt:
Right, agree with @pixelipo. Also, the selected date font should be bold. :)

tbille:
@jancborchardt @pixelipo WDYT?

- [x] header should not be bold (same as core version)
- [x] today date would look better (and have better contrast) with a white font
- [x] the selected date font should be bold

selection_011

css/app/datepicker.css
```diff
@@ -89,27 +89,35 @@
 	border-left: 0;
 }
 
-#datepicker table tbody tr.uib-weeks td:last-child button {
-	border-right: none;
+.today-hilight {
```
jancborchardt: Just one thing: Can you use the class exactly as I said in my previous comment:

> .highlight-today

(Proper english and also the correct order, as we use .highlight-weekend too)

css/app/datepicker.css
```diff
@@ -89,27 +89,35 @@
...
 
 #datepicker table tbody button span {
 	cursor: pointer;
 }
 
 #datepicker .text-muted {
+	color: #c7c7c7;
+}
+
+.weekend-hilight button {
```
jancborchardt: Same here, please call the class

> .highlight-weekend

(Proper english and also the correct order, as we use .highlight-today too)

css/app/datepicker.css
```diff
@@ -89,27 +89,35 @@
...
 }
 
 #datepicker table tbody button {
 	max-width: 34px; /* Hack for Firefox for overflowing tables */
 }
 
 #datepicker table tbody button.active {
-	background-color: rgba(240, 240, 240, .9);
+	border-radius: 50%;
+	background-color: #0082c9;
```
georgehrke: @skjnldsv Can apps use scss too?

raimund-schluessler: Yes, they can.

css/app/datepicker.css
tbille: done

css/app/datepicker.css
tbille: done

css/app/datepicker.css
tbille: they should :) , but it should be in another issue.

It would be much easier to 'debug css' and to adapt the apps to the nextcloud themes

pixelipo: @georgehrke we migrated Deck to SCSS recently: nextcloud/deck#219

css/app/datepicker.css
```diff
@@ -89,27 +89,35 @@
 	border-left: 0;
 }
 
-#datepicker table tbody tr.uib-weeks td:last-child button {
-	border-right: none;
+.hilight-today {
```
pixelipo: As mentioned previously by @jancborchardt - this should be `.highlight-today` (spelling)

css/app/datepicker.css
```diff
@@ -89,27 +89,35 @@
...
 
 #datepicker table tbody button span {
 	cursor: pointer;
 }
 
 #datepicker .text-muted {
+	color: #c7c7c7;
+}
+
+.hilight-weekend button {
```
pixelipo: As mentioned previously by @jancborchardt - this should be `.highlight-weekend` (incorrect spelling)

css/app/datepicker.css
tbille: small mistake.
Done

css/app/datepicker.css
tbille: Done

css/app/datepicker.css
georgehrke: > @georgehrke we migrated Deck to SCSS recently: nextcloud/deck#219

This looks wrong, because it doesn't use the scss engine from the Nextcloud server. Hence it won't update the CSS when the admin changes the theme

georgehrke: > they should :) , but it should be in another issue.

Yes :)

I would vote to merge this now and fix the scss / primary color issue in a follow up PR :)

jancborchardt:
Great work everyone, our new datepicker is awesome! :)

PR #542, opened 2017-07-20
make sure not to include .DS_STORE in appstore bundles

### LINKED ISSUE #494: calendar.tar.gz contains unnecessary  img/.DS_Store file
The nextcloud calendar plugin contains a .DS_Store file, which is a metadata file from OS X finder. This shouldn't cause any problems, but it is unnecessary and should be removed.

The file is not part of the git repository and actually the .gitignore file already contains .DS_Store and thus tries to make sure such files don't get committed. So it seems this got introduced somewhere in the process from repository to tarball.

georgehrke:
fixes #494

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,7 @@ appstore:
 	"js/public" \
 	"COPYING" \
 	"CHANGELOG.md" \
+	".gitignore" \
 	$(appstore_build_directory)
 ifdef CAN_SIGN
 	mv $(configdir)/config.php $(configdir)/config-2.php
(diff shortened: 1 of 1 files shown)


georgehrke:
For crying out loud .........

`--exclude-vcs-ignores` was added with Gnu tar 1.28
The travis image we use has 1.27.1

georgehrke:
I think we should just update the Makefile here and make sure travis doesn't call `tar` with `--exclude-vcs-ignores`. 
Both `trusty` and `trusty-updates` have 1.27.1 and travis doesn't offer xenial or any other newer ubuntu version yet.

PR #568, opened 2017-08-21
Fix patch logic: Properly remove parameter if it no longer exists

leonklingele-work:
I'm committing this in behalf of @georgehrke, thanks for providing the fix.


leonklingele-work:
@georgehrke can this be merged or are we waiting for a third review?

PR #580, opened 2017-08-30
Fix checkbox styling

bergart:
This fixes the typical look of the DOM checkbox with the design of NC (flat and blue background).


georgehrke:
@bergart Can you please always provide before and after screenshots when making design related changes?

bergart:
@georgehrke sure, it should look like:

before:
selection_009

after:
selection_010

PR #590, opened 2017-09-06
Add language selection to attendee section

bergart:
This feature adds a language selection in the attendee section. 

attendeelanguage


js/app/models/simpleEventModel.js
```diff
@@ -51,7 +51,8 @@ app.factory('SimpleEvent', function () {
 		'cutype',
 		'cn',
 		'delegated-from',
-		'delegated-to'
+		'delegated-to',
+		'lang'
```
georgehrke: RFC 5545 defines a [language parameter](https://tools.ietf.org/html/rfc5545#section-3.2.10) which is also [applicable to the Attendee property](https://tools.ietf.org/html/rfc5545#section-3.8.4.1).

Please use officially defined parameter instead of a custom one :)

georgehrke:
Please extend the contacts API to combine it with the lang property if set, to automatically provide the correct language :)

https://tools.ietf.org/html/rfc6350#section-6.4.4

bergart:
>  Furthermore the language label is mispositioned. It looks like its related to [ ] Optional [ ] Does not attend

@georgehrke you are right! There should be a line break.

I will take a look at RFC 5545 & 6350 and adjust it to the notes

jancborchardt:
Before we add anything new there, the attendee list styling should be adjusted to be the same as in file sharing: https://github.com/nextcloud/server/pull/4136

That is:
- avatar/placeholder
- attendee name
- 3-dot menu which triggers a popover with all the options there in a nice list
- everything with proper padding

cc @nextcloud/designers @tcitworld @raghunayyar

controller/viewcontroller.php
```diff
@@ -98,6 +127,65 @@ public function index() {
 			$initialView = 'month';
 		}
 
+		// TODO(artur): Duplicate: Refactor Language --> NC:Server:PersonalInfo.php
+		$forceLanguage = $this->config->getSystemValue('force_language', false);
+		if($forceLanguage !== false) {
+			return [];
+		}
+
+		$userLang = $this->config->getUserValue($userId, 'core', 'lang', $this->l10nFactory->findLanguage());
```
leonklingele-work: Where does `$userId` come from? It's not injected into the controller.

leonklingele-work: Nevermind, it's inside the current scope. Everthing's fine! :)

georgehrke:
Please rebase and fix the broken unit tests

georgehrke:
> 3-dot menu which triggers a popover with all the options there in a nice list

@jancborchardt We should put option-select blocks into the 3 dot menu? oO

georgehrke:
> Before we add anything new there, the attendee list styling should be adjusted to be the same as in file sharing:

Generally: Yes, we have to refactor alarm and attendee to match that style, but this is out of scope for this PR

jancborchardt:
> We should put option-select blocks into the 3 dot menu? oO

To be honest I don’t know why this dropdown is necessary in the first place. Language should be determined by the user’s set system language. Is this for the invitation email?

Ideally the language for sharees should be determined from their address setting in contacts or the personal profile. But not in a dropdown in each app.

georgehrke:
> Is this for the invitation email?

Yes

> Ideally the language for sharees should be determined from their address setting in contacts or the personal profile. But not in a dropdown in each app.

Yes, that's why I asked @bergart to extend the calendar's contacts API to also include the `LANG` field, so it automatically chooses the right language. But I'm not even sure you can set the `LANG` field in the contacts app right now, so one should be able to correct it.

bergart:
@georgehrke broken unit tests are fixed, it ends with success. The `LANGUAGE` parameter is now shown in the iCal.

@jancborchardt  this is just a feature, the UI part should be opened in a new issue / pull request

jancborchardt:
> this is just a feature, the UI part should be opened in a new issue / pull request

The problem is that usually new features are added and then the polish is postponed and/or not done. As the interface in that part is already garbled enough I'd prefer we get it right _before_ adding anything new.

js/app/controllers/attendeecontroller.js
```diff
@@ -62,7 +62,8 @@ app.controller('AttendeeController', function($scope, AutoCompletionService) {
 					'role': 'REQ-PARTICIPANT',
 					'rsvp': 'TRUE',
 					'partstat': 'NEEDS-ACTION',
-					'cutype': 'INDIVIDUAL'
+					'cutype': 'INDIVIDUAL',
+					'language': OC.getLocale() // Use current user's timezone as a default value
```
Peterede: Instead of OC.getLocale() for the default value the default should be the value of the user value 'core' 'lang'.  When loading the page this can be added to the parameters by retrieving  $this->config->getUserValue($user->getUID(), 'core', 'lang'); in viewcontroller.php.

georgehrke:
Closing in favour of https://github.com/nextcloud/calendar/issues/1484

It should not be an input field, instead, this information should automatically be fetched from DAV / the address book.

Issue #571, opened 2017-08-22
Missing 'Nextcloud' logo when accessing shared calendar


leonklingele-work:
Steps to reproduce:

1. From the left sidebar of of the Calendar app, share the calendar publicly:
![contacts-share-publicly](https://user-images.githubusercontent.com/31063993/29574387-b04a9d68-8761-11e7-8f86-b972c51390e2.png)
2. Open the link, the header bar looks similar to this:
![contacts-share-publicly-missing-logo](https://user-images.githubusercontent.com/31063993/29574496-0846a8d6-8762-11e7-8818-cb131466049e.png)

Expected result:
- A Nextcloud logo on the top left

Actual result:
- Misplaced text instead of a logo

georgehrke:
Must be related to some changes in Nc 13 alpha

PR #613, opened 2017-09-23
Move to SCSS [was: App to css guidelines]

skjnldsv:
After https://github.com/nextcloud/server/pull/6399


georgehrke:
Will these changes be compatible with Nextcloud 11 and 12 too?

skjnldsv:
For now compatible 13 only. We'll have to decide if we want to implement a fix until the last supported version is 13. Or something else! :)

skjnldsv:
@georgehrke I added a lots of stuff!
@tcitworld @raghunayyar Could you test it please?
Please list anything you find and I will fix it or explain the choice I made! :)

Also: please test it under nc12. This pr will allow the app to only nc12 and higher, this is unfortunate but all the other apps are doing the same to enjoy the scss unification! :)

georgehrke:
@skjnldsv Is this ready to review?

skjnldsv:
@georgehrke sorry, forgot that :p 
Almost!

skjnldsv:
@georgehrke I think we're good here. :)

georgehrke:
I created a `stable1.5` branch, because this breaks compatibility with Nextcloud 11, which is still supported.

@skjnldsv You said this works with Nextcloud 12, right?

skjnldsv:
No, a specific patch will still be needed.

georgehrke:
- [ ] The new / edit dialogs in the sidebar look a little weird:

0afa7d3e-b0ca-4b9e-80b4-8356127e05a7
164b9d51-00a4-441d-b3e9-88e81b1614bd
360b2d9e-1842-41ba-9bb2-e16c87e4bff7

- [ ] Furthermore the color-picker overlaps with the first item in the calendar-list

georgehrke:
@skjnldsv Can you check this please :)

skjnldsv:
Hum, in the end, it could be a good thing to experiment with the color picker on a popover? :)

georgehrke:
Hm, would be good to get it fixed and merged first and then start experimenting ;)

skjnldsv:
Is this really an issue though? :D

georgehrke:
yes :P ;) 

just for the record:
- [ ] blue tick button misses border
- [ ] input field should span all the way to the left
- [ ] color picker of New Calendar overlaps first item in calendar list

georgehrke:
Also: The New Calendar tick on blue background is white but for New Subscription it's black. It should be white too :)

georgehrke:
Sorry to ask again, but any update on this? :)

georgehrke:
@skjnldsv ^

skjnldsv:
Okay, let's go now! :)

georgehrke:
I will test this asap

Thx for your work! :)

tcitworld:
The CSS seems perfecty but the js is broken (even after rebase).

selection_130

skjnldsv:
@tcitworld I don't have this issue. 
I barely touched the js on this pr, are you sure this is coming from here?

tcitworld:
>  I barely touched the js on this pr, are you sure this is coming from here?

Yes, it's because the message from `'<?php p($l->t('Some events in this calendar are broken. Please check the JS console for more info.')); ?>'` contains a `'` in french and therefore cuts the simple quote.

raimund-schluessler:
The colorpicker still overlaps the next calendar entry, because the `app-navigation-entry-edit` is placed with `display: absolute`. Is this a good idea? Shouldn't this use `relative` instead?

georgehrke:
- [ ] the spinner moved to the left, it used to be in the middle. I don't mind fixing this in a follow up PR though.

bc9f01d8-8765-4900-8037-773bb22a82a4

georgehrke:
@skjnldsv What exactly did you try to fix with your last commit? 

The primary button for New Calendar / New Subscription is pretty important imho.

templates/part.calendarlist.item.php
```diff
@@ -22,56 +22,45 @@
...
-	<span class="icon icon-error"
-		  ng-if="item.calendar.hasWarnings()"
-		  title="<?php p($l->t('Some events in this calendar are broken. Please check the JS console for more info.')); ?>">
-		&nbsp;&nbsp;&nbsp;&nbsp;
-	</span>
+   title="{{
+      item.calendar.hasWarnings()
+      ? '<?php p($l->t('Some events in this calendar are broken. Please check the JS console for more info.')); ?>'
+      :item.calendar.displayname
+   }}">
```
georgehrke: This was never an Error Message or Displayname.

Displayname was always displayed.
Also see @tcitworld's comment about this breaking for french e.g.

skjnldsv:
> The primary button for New Calendar / New Subscription is pretty important imho.

It far more easier to just remove it than create a full multiple fixes just for a border... :/
I dont' really think it add that much ! :)

> the spinner moved to the left, it used to be in the middle. I don't mind fixing this in a follow up PR though.

Where can I find this spinner? :)

skjnldsv:
> The colorpicker still overlaps the next calendar entry, because the app-navigation-entry-edit is placed with display: absolute. Is this a good idea? Shouldn't this use relative instead?

No, because it overlaps the whole entry with an animation (theoretically). There is a swipe-from-right overlapping :)

raimund-schluessler:
> No, because it overlaps the whole entry with an animation (theoretically). There is a swipe-from-right overlapping :)

I don't get it. It not only overlaps the edited entry, but also the next one. Though this is because the colorpicker leads to a larger height than a regular entry, it still should be solved, not?

skjnldsv:
@raimund-schluessler because the edit entry is not supposed to allow height higher than 44px. The guidelines does not allow this. This is why I suggested we put the color picker inside the edit menu.

raimund-schluessler:
> This is why I suggested we put the color picker inside the edit menu.

This sounds like a larger effort, probably better suited for a separate PR?

skjnldsv:
Agreed, the overlapping on the next item isn't an issue for me. It' s like the popover menu in the end. :)

georgehrke:
> It far more easier to just remove it than create a full multiple fixes just for a border... :/

But it's just a border, isn't it? 

> Where can I find this spinner? :)

https://github.com/nextcloud/calendar/blob/43bd60006c0f926ac8d9aae31899724e6fe278c1/templates/part.calendarlist.php#L25

skjnldsv:
> But it's just a border, isn't it?

Not exactly, the two buttons are superposed by 1px to share a common border. Which means changing the border color to primary will remove the left border of the second button... So either one or the other is not very convenient! :/

georgehrke:
I just noticed, that the input got shorter too. It should be full width.
f19d98b1-8484-4bf8-adb6-20d9cd7f8be6

skjnldsv:
<skjnldsv> @georgehrke, there is a last small issue for the css guidelines update
<skjnldsv> the "shared" text on a shared calendar is breaking stuff a bit
<skjnldsv> what do you think of just changing the opacity of the shared icon?
<skjnldsv> like we do in files?

skjnldsv:
All clear, please review! @georgehrke @tcitworld @raghunayyar

splitt3r:
@skjnldsv is checkmark, close the prewferred way? https://github.com/nextcloud/documentation/blob/91fffff7f444b8523c2f2e7bbd459194d0071bfa/developer_manual/design/navigation.rst#edit-entry does it the other way around?

skjnldsv:
@splitt3r I don't think there is a proper way for this, I would go for close then check, but that's just a personal preference :)

splitt3r:
Same here. I think it would be nice to have it streamlimed over alle apps using it.

georgehrke:
I second @splitt3r here. The position / order of Create and Exit / Cancel should be consistent for all new ... dialogs.

skjnldsv:
Let's go then! :)

georgehrke:
I'm just writing unit tests for a different PR. Will review this first thing after that :)

georgehrke:
Please fix the order for new subscription too ;)

skjnldsv:
🙈

skjnldsv:
Bump! Let's merge this! :)

georgehrke:
I’ve had quite a few angular / ja errors when testing this 3 days ago. Will post them later and look into them

templates/part.calendarlist.item.php
georgehrke: This issue was not resolved yet and will still cause issues with french locale

templates/main.php
```diff
@@ -84,20 +84,16 @@
 		</div>
 
 		<?php if(!$_['isPublic']): ?>
-		<div ng-controller="CalendarListController" id="calendarlistcontainer" ng-cloak>
-			<div id="scrollable">
```
georgehrke: @skjnldsv You can't remove `#scrollable` here, it served a very important purpose:

kapture 2017-12-19 at 15 20 34

The lower part of the list is not hidden behind the settings button

georgehrke:
9e4e736f-5411-42fb-8bde-1fdc6faefa7b

The dropdown menu is hidden behind the new subscription input

georgehrke:
- [ ] The sharing opacity thing doesn't work for me. All sharing icons are the same, shared or not shared

georgehrke:
@skjnldsv @tcitworld I fixed all the issues that i found. Can you double check? :)
I would like to merge this soon

tcitworld:
Will have a look this afternoon.

skjnldsv:
Amazing work @georgehrke Works perfectly! 😍 🎉

georgehrke:
@tcitworld Is it okay if i go ahead and merge this?

Given the [release schedule of the server](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule#nextcloud-13) i would like to get this in fairly soon ;)

tcitworld:
All I saw was this issue : 
The nc logo is not displayed on the public calendar page
selection_161

georgehrke:
That seems related to this: https://github.com/nextcloud/calendar/issues/571 :)

tcitworld:
Fixed. Good to go.

georgehrke:
awesome, let's get this in :)

tcitworld:
Did realize I didn't test this on nc12 though.

skjnldsv:
@tcitworld if we want to keep it ok for 12, it's an easy fix :)
@georgehrke do you both want to keep 12 + 13 ?

PR #622, opened 2017-10-05
Add support to auto-open the 'Create Event' modal

leonklingele-work:
This can be used to link to the Calendar which then should open the 'Create Event' modal.
Test this patch by opening `/apps/calendar/#!?newevent`

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -26,8 +26,8 @@
 * Description: The fullcalendar controller.
 */
 
-app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEventService', 'SettingsService', 'TimezoneService', 'VEvent', 'is', 'fc', 'EventsEditorDialogService', 'PopoverPositioningUtility', '$window', 'isPublic', 'constants',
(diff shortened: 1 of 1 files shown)


leonklingele-work:
Updated to conform with jshint style.

js/app/controllers/calcontroller.js
```diff
@@ -38,6 +38,18 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
 		$scope.eventModal = null;
 		var switcher = [];
 
+		$scope.$on('ready', () => {
+			var autoOpenNewEvent = $location.search().hasOwnProperty('newevent');
+			if (autoOpenNewEvent) {
+				$scope.fcConfig.select(
+					moment(),
+					moment(),
+					{}, // TODO(leon): Do we need this argument?
```
georgehrke: Yes, It's kind of important to determine the position of the element. 

Instead of calling `fcConfig.select` directly, I would emit a `click()` event on todays fullcalendar item.

georgehrke:
We should use vue-routing for this once we migrate to vue.

PR #628, opened 2017-10-11
autocompletion: Use 'POST' instead of 'GET' method to access API

leonklingele-work:
This prevents including the search term in webserver logs.
It will also break all clients using this API endpoint, which is
fine, as non-OCS APIs are meant to be private and shouldn't be used
by 3rdparty apps.


tcitworld:
Bower seems to have some issues.

leonklingele-work:
@tcitworld should be fine now, unit tests have been fixed.

tcitworld:
@leonklingele-work It's related to https://twitter.com/bower/status/918073147789889536

georgehrke:
@leonklingele-work Can u please squash the commits? :)

leonklingele-work:
@georgehrke done.

PR #634, opened 2017-10-16
Add support to open event by id

leonklingele-work:
This allows to directly open the 'Edit meeting' modal / sidebar by
accessing an URL like the following:

    /apps/calendar/#!?edit=Nextcloud-EVENTIDHERE.ics

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -26,8 +26,8 @@
 * Description: The fullcalendar controller.
 */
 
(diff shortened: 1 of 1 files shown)


georgehrke:
Please use the `HashService`.

Why did you go for `#!?` ?

leonklingele-work:
> Why did you go for #!? ?

`$location.search()` otherwise doesn't return anything :/

georgehrke:
See https://github.com/nextcloud/calendar/blob/00c67ab949ee6df6705a6fea57d24594e8229511/js/app/controllers/calendarlistcontroller.js#L335 for usage of `HashService`

georgehrke:
Generally using `clientEvents` won't work, because not all events are cached in fullcalendar. The calendar only loads events that are actually displayed to the user. 

If the event in question is in the previous month, it won't be in the fullcalendar cache.

georgehrke:
Closing

This approach is not working and also won't work in the future, because fullCalendar's `clientEvents` is never gonna contain all events a user owns.

PR #643, opened 2017-10-26
Add a public og:description meta tag

### LINKED ISSUE #609: Show proper metadata for when you post the link in social networks or chats
Currently you see stuff like this:
calendar link js
»This app requires Javascript…« instead of displaying the correct Calendar metadata, like the name and the person it was shared by, which should be in the website metadata (ogp or so).

cc @nextcloud/calendar

maximevaillancourt:
### What are you trying to achieve with this PR?
Fix #609. By adding the `og:description` meta tag on public templates, sites such as Riot.im (powered by [Synapse](https://github.com/matrix-org/synapse/)) now use the Open Graph content instead of scraping the text nodes on the page, thus offering a better experience for users who receive a shared calendar link.

### Why are you going with this approach?
I placed the `og:description` with the other Open Graph tags to keep them all together.

### Discussion
**Couldn't we show the shared calendar's detailed information?**
We could move `og:description` in [`part.publicinformations.php`](https://github.com/nextcloud/calendar/blob/master/templates/part.publicinformations.php) to offer detailed calendar information such as the calendar name, its owner, and the user who shared it. However, I could not get this approach to work since it requires string interpolation with Angular's `ng-repeat` directive, which (I think) has no easy way to integrate with NextCloud's string translation method (`$l->t(...)`).

**Okay, but couldn't we provide a better string instead of just "Public access"?**
Absolutely. It would require creating a new string in the [`l10n`](https://github.com/nextcloud/calendar/tree/master/l10n) files, something along those lines: "Someone has shared a public link to a calendar on `$theme->getName()`".

### DIFF
--- a/templates/main.php
+++ b/templates/main.php
@@ -25,6 +25,7 @@
  /* OpenGraph */
 if($_['isPublic']) {
 	OCP\Util::addHeader('meta', ['property' => "og:title", 'content' => $theme->getName() . ' - ' . $theme->getSlogan()]);
+	OCP\Util::addHeader('meta', ['property' => "og:description", 'content' => $l->t('Public access')]);
 	OCP\Util::addHeader('meta', ['property' => "og:site_name", 'content' => $theme->getName()]);
 	OCP\Util::addHeader('meta', ['property' => "og:url", 'content' => $_['shareURL']]);
 	OCP\Util::addHeader('meta', ['property' => "og:type", 'content' => "object"]);


tcitworld:
Let's use 
> Someone has shared a public link to a calendar on $theme->getName()

Instead then. You don't need to do anything with the l10n files, it will be handled automatically.

maximevaillancourt:
@tcitworld 

I'm not sure I understand — you're telling me that there is an automatic translation engine in place behind the `$l->t(...)` method that would allow me to simply update my PR as follows and have the string get translated automatically, even though `Someone has shared a public link to a calendar on %s` does not exist in any of the l10n files? If so, that's *awesome*. Otherwise, please explain further. Thanks for the help! :smiley: 

```diff
diff --git a/templates/main.php b/templates/main.php
 
 /* OpenGraph */
 if($_['isPublic']) {
 	OCP\Util::addHeader('meta', ['property' => "og:title", 'content' => $theme->getName() . ' - ' . $theme->getSlogan()]);
-	OCP\Util::addHeader('meta', ['property' => "og:description", 'content' => $l->t('Public access')]);
+	OCP\Util::addHeader('meta', ['property' => "og:description", 'content' => $l->t('Someone has shared a public link to a calendar on %s', [$theme->getName()])]);
 	OCP\Util::addHeader('meta', ['property' => "og:site_name", 'content' => $theme->getName()]);
 	OCP\Util::addHeader('meta', ['property' => "og:url", 'content' => $_['shareURL']]);
 	OCP\Util::addHeader('meta', ['property' => "og:type", 'content' => "object"]);
```

raimund-schluessler:
> I'm not sure I understand — you're telling me that there is an automatic translation engine in place behind the $l->t(...) method that would allow me to simply update my PR as follows and have the string get translated automatically, even though Someone has shared a public link to a calendar on %s does not exist in any of the l10n files? If so, that's awesome.

That is exactly what he meant ;) Nextcloud uses Transifex for handling translations, so strings to be translated are collected by transifex, translated there and pushed back to the repo. No need to translate anything here.

jancborchardt:
We can be more specific, no? Instead of the bland

> Someone has shared a public link to a calendar on Nextcloud

> Georg shared the calendar Release planning

The important things are the name of the sharer and the calendar name. :)

georgehrke:
> We can be more specific, no? 

I’m afraid this is not that easy at the moment. We only get the sharer name and display name of the calendar  after sending some CalDAV requests in JavaScript.
So we don’t know that value when rendering the template yet.

We will have to implement some API in the dav app for that

jancborchardt:
Then yeah, let's implement that to make it happen. ;) "Someone shared …" is really a bit strange.

maximevaillancourt:
Just a quick update: I didn't take the time to implement the above yet, but I just wanted to let you all know that I (partly) fixed the issue on the Synapse side of things (Riot.fm is powered by Synapse): https://github.com/matrix-org/synapse/pull/2576.

jancborchardt:
@maximevaillancourt do you need any help with this? Feel free to ask. :)

georgehrke:
Closing in favour of https://github.com/nextcloud/calendar/issues/609

See for updated information.

PR #659, opened 2017-11-14
Fix #269 - Escaping on translation's variables disabled

### LINKED ISSUE #269: Wrong single quote convertion
### Steps to reproduce
1. Select French language
2. Create event and add pop-up alarm in reminders
3. Single quote is transform to `&#039;`

image

Simounet:
Hello,
First of all, thanks for this great app I'm using everyday.

I fixed the issue with accents on reminders. The t() function is escaping translation's variables by default so I disabled this because we control inputs filled in. This is not dirty user datas.
I simplified the code but I'm don't know much about your dev rules. Maybe you prefer to keep translation keys without variables. Let me know.

### DIFF
--- a/js/app/filters/reminderfilter.js
+++ b/js/app/filters/reminderfilter.js
@@ -47,36 +47,27 @@ app.filter('simpleReminderDescription', function() {
 
 		var relative = alarm.trigger.type === 'duration';
 		var relatedToStart = alarm.trigger.related === 'start';
+		var translationEscapingDisabled = { escape: false };
+		var translationVars = { type: getActionName(alarm) };
 		if (relative) {
 			var timeString = moment.duration(Math.abs(alarm.trigger.value), 'seconds').humanize();
 			if (alarm.trigger.value < 0) {
-				if (relatedToStart) {
-					return t('calendar', '{type} {time} before the event starts', {type: getActionName(alarm), time: timeString});
-				} else {
-					return t('calendar', '{type} {time} before the event ends', {type: getActionName(alarm), time: timeString});
-				}
+				var translationKey = '{type} {time} before the event ' + ( relatedToStart ? 'starts' : 'ends' );
+				translationVars.time = timeString;
 			} else if (alarm.trigger.value > 0) {
-				if (relatedToStart) {
-					return t('calendar', '{type} {time} after the event starts', {type: getActionName(alarm), time: timeString});
-				} else {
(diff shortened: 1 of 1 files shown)


georgehrke:
I didn't test this yet, but the value of `type` is [directly user controlled](https://github.com/Simounet/calendar/blob/14f3f1205920feda2d8b6c023f320953ce92dac9/js/app/filters/reminderfilter.js#L39) and if u don't escape that, you create a huge XSS vulnerability.

And as u can share calendars, an attacker with an account could easily create an event with a malicious `VAlarm` block and share it with other users.

georgehrke:
This is also related to this core issue: https://github.com/nextcloud/server/issues/3692

georgehrke:
@Simounet I don't speak french, can you explain to me what exactly is double escaped here? 
Is it the `Audio alarm` / `Pop up` / `E-mail` part?

tcitworld:
@Simounet please provide a screenshot

georgehrke:
@tcitworld He is referring to #269

Simounet:
From the reminder sidebar:
nextcloud-calendar-encoding-issue

Simounet:
I get the security point of view but it seems strange to me that we can't store a secure version of the user input then display it nicely.

georgehrke:
@Simounet so what exactly is double escaped. I don’t speak french, please translate :)

Simounet:
`Fenêtre d&#039;alerte ` should be `Fenêtre d'alerte ` so it's the `{type}` var into the translation key.

georgehrke:
Okay, I guess we can fix this by not injecting the type dynamically but adding 4 hardcoded versions, one for each type.

`E-mail {time} before the event`
`Pop-up {time} before the event`
`Audio Alarm {time} before the event`
...

georgehrke:
I will send a pull-request to fix this

georgehrke:
https://github.com/nextcloud/calendar/pull/661

georgehrke:
> I get the security point of view but it seems strange to me that we can't store a secure version of the user input then display it nicely.

Relying on "the data is probably safe" is never a good approach when handling user data. Users are able to store any valid (as of RRC5545) calendar-data in Nextcloud's CalDAV server and that's the expected behavior. A CalDAV server is never supposed to alter user-data.

Furthermore think about a scenario where an attacker gains access to an SQL injection. The attacker would be able to modify the calendar data and hence also manipulate events and the included VALARM.

PR #661, opened 2017-11-15
fix escaping of alarm type names

georgehrke:
supersedes #659 

@tcitworld Can you verify that this works with french locale? 
Then i will go ahead and fix the unit tests


tcitworld:
Tested and works : 

nc_calendar_l10n

georgehrke:
fixed the tests

PR #667, opened 2017-11-28
add special treatment for duplicate errors when importing calendars

### LINKED ISSUE #589: Import only events whose UID does not already exist
### Steps to reproduce
1. Import ics.
2. Import same ics again.

### Expected behaviour
It will be only imported events, whose UID does not already exist in calender for the user.

### Actual behaviour
Every event is imported regardless if it already exist or not. Therefore it exists events with the same UID in the same calender, which is not allowed by CalDAV specification: https://tools.ietf.org/html/rfc4791.

> 
>  The UID property value of the calendar components contained in a
>    calendar object resource MUST be unique in the scope of the calendar
>    collection in which they are stored.
> 
>    Calendar components in a calendar collection that have different UID
>    property values MUST be stored in separate calendar object resources.
> 
>    Calendar components with the same UID property value, in a given
>    calendar collection, MUST be contained in the same calendar object
>    resource.  This ensures that all components in a recurrence "set" are
>    contained in the same calendar object resource.  It is possible for a
>    calendar object resource to just contain components that represent
>    "overridden" instances (ones that modify the behavior of a regular
>    instance, and thus include a RECURRENCE-ID property) without also
>    including the "master" recurring component (the one that defines the
>    recurrence "set" and does not contain any RECURRENCE-ID property).

I already found #53, which was closed earlier this year. However I confirm that it makes no sense to import already existing events. Therefore I would not ask the user about replacing / adding the event. Instead, the item to be imported should be ignored if another one already exists with the same UID in the calendar.

Independent of the CalDAV specification, here is an use case:

User deletes an event and needs to restore a backup. The backup does not contain all events of the current calendar but the event, which was deleted by accident. Therefore I cannot delete the whole calendar (because I would delete events, which are not in the backup), but I also cannot import the ics because of the duplicated events.

georgehrke:
fixes #589

### DIFF
--- a/js/app/filters/importerrorfilter.js
+++ b/js/app/filters/importerrorfilter.js
@@ -4,7 +4,7 @@
  * @author Raghu Nayyar
  * @author Georg Ehrke
  * @copyright 2016 Raghu Nayyar <hey@raghunayyar.com>
- * @copyright 2016 Georg Ehrke <oc.list@georgehrke.com>
+ * @copyright 2017 Georg Ehrke <oc.list@georgehrke.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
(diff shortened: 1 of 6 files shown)


js/app/filters/importerrorfilter.js
```diff
@@ -29,17 +29,29 @@ app.filter('importErrorFilter', function () {
...
-		switch(file.errors) {
-			case 0:
-				return t('calendar', 'Successfully imported');
-
-			case 1:
+		if (file.errors === 0) {
+			return t('calendar', 'Successfully imported');
+		} else if(file.errors === 1) {
+			if (file.duplicates === 1) {
+				return t('calendar', 'Partially imported, skipped 1 duplicate');
```
tcitworld: 1 -> one

js/app/filters/importerrorfilter.js
```diff
@@ -29,17 +29,29 @@ app.filter('importErrorFilter', function () {
 		}
 
 		//TODO - use n instead of t to use proper plurals in all translations
```
tcitworld: Maybe a good time to handle this ?

js/app/filters/importerrorfilter.js
```diff
@@ -29,17 +29,29 @@ app.filter('importErrorFilter', function () {
...
-		switch(file.errors) {
-			case 0:
-				return t('calendar', 'Successfully imported');
-
-			case 1:
+		if (file.errors === 0) {
+			return t('calendar', 'Successfully imported');
```
tcitworld: Is there a way we could bring in the number of imported events here ?

georgehrke: Good call, will add it

js/app/filters/importerrorfilter.js
georgehrke: I don't think we can have both, `n` and `one`

georgehrke:
@tcitworld I updated the strings, now it's more verbose

PR #669, opened 2017-12-01
Attendee enter fix

### LINKED ISSUE #502: Enter in new attendee field closes sidebar

platypro:
Fixes #502. The enter key should be a little more predictable. Enter by itself will never close the dialog or apply anything, but Ctrl+Enter does. Now when the enter key is pressed while adding an attendee, it adds it instead of closing (and applying) the event.


georgehrke:
@platypro Thx a lot for your pull-request! 
Can you please sign off your commits?

georgehrke:
@jancborchardt What do you think about the `Ctrl` + `Enter`?

georgehrke:
- [ ] Test if it works with `CMD` + `Enter` too (for macs)

jancborchardt:
Well, enter in the new attendee field should not close the sidebar / dialog at all either. :) In this case it seems that enter and ctrl+enter should do the same thing?

georgehrke:
> Well, enter in the new attendee field should not close the sidebar / dialog at all either. :) In this case it seems that enter and ctrl+enter should do the same thing?

@jancborchardt The idea here is that you can only close the dialog with CTRL+C / ⌘+C

Enter will only work for the field you are currently editing.

jancborchardt:
Ctrl-C is a shortcut for copying, not closing. :)

- Enter should save but not close (as does this pull request, correct @platypro?)
- Ctrl-Enter should save and closr
- Esc should close and not save
- Ctrl-C should not close ;)

georgehrke:
I'm sorry, i obviously didnt mean CTRL+C but CTRL+Enter

platypro:
@jancborchardt The idea was that enter should not save anything when the user was probably intending to do something else, such as adding a new attendee. If this pull request does otherwise then that was not what I intended.

jancborchardt:
@platypro well, enter does »save« in the sense that it adds a new attendee, right? I basically agree with what this pull request does design-wise. :) If it works then go ahead with it @georgehrke

platypro:
@jancborchardt If thats what you were implying then its working how I expected it to.

georgehrke:
@platypro Can you please modify the template of the events editor popover as well, so we have the same CTRL / CMD + ENTER behavior there as well :) 
THX!

georgehrke:
> Ctrl-Enter should save and close

That's not the case yet. It only saves the event, but doesn't add the attendee.

platypro:
I see what you mean. Would it also be acceptable for a notification to show if an attendee is added? It might not be obvious to everyone.

js/app/controllers/editorcontroller.js
```diff
@@ -121,6 +121,14 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
 			});
 		};
 
+		$scope.keypress = function(event) {
+			var code = event.keyCode || event.which;
+			if(event.ctrlKey===true && code === 13) {
```
georgehrke: Can you please use `event.metaKey`. Then it will also work on Macs. :)

https://api.jquery.com/event.metakey/

georgehrke:
And please remember to sign your commits :) 

Then this is ready to go

js/app/controllers/editorcontroller.js
```diff
@@ -123,8 +123,8 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
 
 		$scope.keypress = function(event) {
 			var code = event.keyCode || event.which;
-			if(event.ctrlKey===true && code === 13) {
-				$scope.$broadcast('save-contents');
+			if((event.metaKey===true || event.ctrlKey===true) && code === 13) {
```
georgehrke: `if(event.metaKey===true && code === 13) {` should be just fine.

Meta-Key is true for CTRL on Windows / Linux and CMD on Mac :)
https://api.jquery.com/event.metakey/

js/app/controllers/editorcontroller.js
```diff
@@ -123,8 +123,8 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
...
+			  $scope.$broadcast('save-contents');
```
georgehrke: Please fix the indentation back to tabs

js/app/controllers/editorcontroller.js
georgehrke: and please add spaces on the left side: `event.metaKey === true`, not `event.metaKey===true `

georgehrke: Sorry, it doesn't map to CTRL on Windows, it maps to the windows key.
My fault

georgehrke:
@platypro Thanks a lot for this pull-request! :)

PR #676, opened 2017-12-06
show warning when a user has email reminders

georgehrke:
because they are not implemented in the CalDAV server yet

cf7ff9f9-d760-4fb7-9109-96ccd8442dee


tcitworld:
What's the benefit over disabling email inside `reminderTypeSelect` ?

georgehrke:
People may import ics files from different sources or create events with email reminders in their mobile / desktop clients, so we should be able to display all reminders.

georgehrke:
You will see this message immediately when selecting email in the reminder editor.

PR #684, opened 2017-12-17
Spelling: Redundant "has been" and "while" removed

comradekingu:


js/app/controllers/calendarlistcontroller.js
```diff
@@ -148,9 +148,9 @@ app.controller('CalendarListController', ['$scope', '$rootScope', '$window', 'Ha
 			MailerService.sendMail(item.email, item.publicSharingURL, item.calendar.displayname).then(function (response) {
 				if (response.status === 200) {
 					item.email = '';
-					OC.Notification.showTemporary(t('calendar', 'Email has been sent.'));
+					OC.Notification.showTemporary(t('calendar', 'Email sent.'));
```
georgehrke: 👍

js/app/controllers/calendarlistcontroller.js
```diff
@@ -148,9 +148,9 @@ app.controller('CalendarListController', ['$scope', '$rootScope', '$window', 'Ha
...
 				} else {
-					OC.Notification.showTemporary(t('calendar', 'There was an issue while sending your email.'));
+					OC.Notification.showTemporary(t('calendar', 'There was an issue sending your email.'));
```
georgehrke: 👎 Please change this back.

`There was an issue while sending your email.` is grammatically correct and is also a more common phrase to use according to a native speaker i asked.

js/app/controllers/calendarlistcontroller.js
```diff
@@ -104,7 +104,7 @@ app.controller('CalendarListController', ['$scope', '$rootScope', '$window', 'Ha
 						$scope.subscription.newSubscriptionLocked = false;
 					})
 					.catch(function() {
-						OC.Notification.showTemporary(t('calendar', 'Error saving WebCal-calendar'));
```
georgehrke: Please change this back to `Error saving WebCal-calendar` (and `Error saving WebCal-calendar` only)

js/app/controllers/calendarlistcontroller.js
```diff
@@ -148,9 +148,9 @@ app.controller('CalendarListController', ['$scope', '$rootScope', '$window', 'Ha
 			MailerService.sendMail(item.email, item.publicSharingURL, item.calendar.displayname).then(function (response) {
 				if (response.status === 200) {
 					item.email = '';
-					OC.Notification.showTemporary(t('calendar', 'Email has been sent.'));
+					OC.Notification.showTemporary(t('calendar', 'Email sent.'));
 				} else {
-					OC.Notification.showTemporary(t('calendar', 'There was an issue while sending your email.'));
```
georgehrke: Please change it back to `There was an issue while sending your email.` and don't come up with new phrases for this very string please. There is nothing wrong about `There was an issue while sending your email.`

comradekingu: While true, in the greater scheme of things, there are 83 "Could not"-errors, and only 8 "There was"-errors. Two of which are the ones dealt with here.
My thoughts on the matter is "could not" is shorter, and "There was an issue" is ambiguous as to whether it is a failure or not. One could argue the email got sent, and that the issue was some headers were omitted or anything else that amounts to an issue.

georgehrke:
Thx for your pull-request. Generally we should not take changing strings no that easily though, because all translated strings will be lost and will have to be retranslated.

PR #712, opened 2018-01-31
fix displayname on public sharing site

### LINKED ISSUE #596: Displayname is missing on public sharing site
works on Nextcloud 11, is broken on Nextcloud 12 and master

It's supposed to show the displayname, but it shows the user id.
Might be a server-side issue. Subject to investigation.

847b51f4-baf0-473f-9271-57dc0db21b61

georgehrke:
fixes #596 

before:
before

after:
after

### DIFF
--- a/js/app/factory/calendarFactory.js
+++ b/js/app/factory/calendarFactory.js
@@ -19,7 +19,7 @@
  *
  */
 
-app.service('CalendarFactory', function($window, DavClient, Calendar, WebCal, constants) {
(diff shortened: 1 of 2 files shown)


Bellusterra:
looks like this issue got dropped in favor of changing the UI, but still appears to be an issue. Events within the calendar embedded or otherwise will still show the user id not the display name.  I also remember having issues using the calendar as a result of this because my user id had a space in it and required me to migrate all my data to a new profile without a space to function properly.

PR #715, opened 2018-02-04
fix submit buttons in event editor, broken by #669

georgehrke:


georgehrke:
Going ahead, because I need to package 1.6.0 tonight

PR #716, opened 2018-02-05
fix height of summary input in event editor popover

georgehrke:
before:
before

after:
after


georgehrke:
Going ahead, because I need to package 1.6.0 tonight

PR #737, opened 2018-02-23
fix ProxyControllerTest, issues caused by updated Guzzle lib in the server

georgehrke:
Update only affected some classes used for testing in the ProxyControllerTest, ProxyController is unaffected by updated Guzzle lib.


georgehrke:
change only affected tests and not actual code
fixes (among other things) nightly builds

PR #743, opened 2018-02-26
timezone select

### LINKED ISSUE #586: Let user decide their timezone
There are quite a few people having issues with our automatic timezone detection: https://github.com/nextcloud/calendar/issues/496

Furthermore there are scenarios where you want to see the calendar in a different timezone.
Let's say you are on a trip in New York and want to see your calendar for next week, where you gonna be in Germany again, then you will want to see the appointments for next week in the german timezone. Therefore a user should be able to overwrite the timezone setting.

Additionally, users accessing the public sharing site, should also be offered a timezone dropdown.

Where should we place the timezone dropdown? 
Settings in calendar or personal settings? 
(I was thinking about upper right corner for public sharing site)

Default should always be set to automatic detection.

cc @jancborchardt @leonklingele-work @tcitworld

georgehrke:
fixes #586


js/app/models/fcEventModel.js
```diff
@@ -46,8 +46,8 @@ app.factory('FcEvent', function(SimpleEvent) {
 			_isAFcEventObject: true,
 			id: id,
 			allDay: allDay,
-			start: start.toJSDate(),
-			end: end.toJSDate(),
+			start: moment(start.toString()),
+			end: moment(end.toString()),
```
georgehrke: This small change has actually a big impact. I'm considering doing a 1.6.1 beta before the final release.

tcitworld: I don't see right away what it does, I'll check later.

georgehrke: It all comes down to one difference: 
- the JS date object stores date and time in relation to an offset.
- the string representation just stores date and time.

Let's say your browser's UTC offset is one hour and so is the offset of the timezone you selected: both will work and the event will be displayed correctly.

But let's say your browser's UTC offset is one hour, but you selected New York as your timezone, then you will still see the event with the one hour UTC difference, although you would expect it to be displayed in New York timezone.

That's because when invoking the convertZone method on `ICAL.Time`, the offset of the js date object is updated as well.

If you have an js date object like this:
`2018-02-27 18:00 UTC+1`
converting to New York timezone results in 
`2018-02-27 12:00 UTC-5`

Now your browser wants to display it, but nows you are not in UTC-5 but in UTC+1, so it adds the 6 hours again and you end up with `2018-02-27 18:00`

tcitworld: Thx :)

PR #745, opened 2018-02-27
add better error handling for importing calendars

### LINKED ISSUE #598: Import gets stuck at Analyzing when presenting it an empty file
Not empty as in `VCALENDAR`, but no `VEVENT`
but actually empty

georgehrke:
fixes #598


georgehrke:
There are no tests for the `ImportController` nor the `ImportFileWrapperModel` at all at the moment.

And I would like to refactor import soon anyway, because the entire code is pretty ugly and hacky. Gonna write tests when refactoring.

tcitworld:
Yeah but you added fixtures anyway, that why I asked ;)

Issue #675, opened 2017-12-06
Double the height of the description textarea


tcitworld:
Otherwise people think it's a one-line input and don't fill multi-lines (and are afraid to hit backspace to send the form).

georgehrke:
Hey @Keerthana17,

great that you want to tackle this issue! :)
If you have any questions, please don't hesitate to ask

PR #751, opened 2018-03-03
Double the height of the description text area (issue #675)

zrwitter:
In response to issue #675, this doubles the size of the text area input for the event description by increasing the text area to two rows.

Before:
screenshot 1

After:
screenshot 2


tcitworld:
> @tcitworld Should we do the same for the location field?

Not sure about this one : it doesn't seems so much as important and we need the advanced editor trying to show up without scrollbars in most resolution.

Moreover, I guess at some point we could have some typeahead plugin with autocomplete (OSM, data from existing events, ...).

PR #764, opened 2018-03-11
Changes made to the calcontroller.js file

arunikayadav42:
I was working on the issue to make more than one calendars visible on the same page.Thus different tokens separated by "." were split using "." as the delimiter and each token was accessed individually.


georgehrke:
Closing in favor of #765

PR #818, opened 2018-05-05
Use clearer color to easier locate today.

pascalBokBok:


georgehrke:
@pascalBokBok Thx for your pull-request! 

Can you please provide before / after screenshots for pull-requests, that affect UI elements? :)
That makes reviewing them much easier. Thx

pascalBokBok:
Screenshots - all calendar events disabled for privacy reasons, but the chaos of having them there makes it more necessary to have a clearer color. 
Before:
image

After:
image

pascalBokBok:
@georgehrke was above what you meant?

georgehrke:
@pascalBokBok Yes, exactly Thx :)

I would probably prefer something slightly brighter, but then this is unfortunately something that heavily depends on the users screen / screen settings.

What do you think about `#FAF2C9`:
73e1a609-284a-4e20-bd9b-9584fa7f49ac-2

pascalBokBok:
I think there is a reason I submitted the change on a saturday. 
It looks fine on a tuesday but, the contrast of the brighter color is quite low in the grey area showing the weekend.
Changing the weekend block to #99cde9 (light blue taken from inactive app icons in top bar) is also a nice way of solving the problem.. All three colors look OK if the weekend is changed to a light blue #99cde9 ( who wants grey weekends anyway).
image
(The crazy sunday first can be ignored - it's from the nextcloud demo page, and I don't know how to change it)

georgehrke:
@skjnldsv @jancborchardt Any opinion on using `#FAF2C9` for today? (see my screenshot) Weekends should remain grey imho.

georgehrke:
037d1f18-1ac8-4fb1-bd19-ad0b636622c1-2

Another approach from FullCalendar: instead of coloring the day cell, just highlight the day number in the upper right corner.

skjnldsv:
I like this idea :)
Weekend should probably stay a little darker yes

raimund-schluessler:
> Another approach from Fantastical: instead of coloring the day cell, just highlight the day number in the upper right corner.

How does that look in the week view?

georgehrke:
@raimund-schluessler 
7824f5f6-96e2-4932-9f4b-ed9f10252c7f-2

raimund-schluessler:
I like the idea as well, but for the week view this seems to require quite some changes to the design.

pascalBokBok:
Highlighting the day number looks really good :+1:

georgehrke:
Something like this?

e61dfec7-c7b9-4372-814c-375f291b9095

pascalBokBok:
I think it's a great solution!
Please feel free to close this request when it's convenient.

jancborchardt:
Yup, highlighting the day number would have been by suggestion too. :) Background color is still important for highlighting, because the number is only in a corner. But as color we shouldn't use yellow, but either grey or a light shade of Nextcloud blue.

georgehrke:
@jancborchardt What if the theme color is white / too light?

jancborchardt:
@georgehrke then for highlighting, a grey should be used.

jancborchardt:
@pascalBokBok can you adjust the pull request to reflect the proposal? :)

pascalBokBok:
No, that goes beyond my CSS skillz, but clearly @georgehrke was able to get it working :-)

dugite-code:
for the impatient (aka my boss), using the custom css app this is what I'm using to add a circular background to the today number.

```css
.fc-today > .fc-day-number {
  background: #0082c9;
  color: #ffff;
  border-radius: 50%;
  padding-left: 6px;
  padding-right: 6px;
}
```

georgehrke:
@dugite-code Do you mind sending a pull-request please? :)

georgehrke:
superseded by #850

PR #822, opened 2018-05-09
 Scheduling Hotfix, Time Reporting, Export Rendering

jsc144:
Scheduling Hotfix- Allows users to share their calendar with another person. This other person can confirm their attendance to the event, causing it to autopopulate into the calendar layer that the user designates. Uses database along with a public write token to ensure secure write operations

Time Reporting - Added another button to the settings menu that allows a user to generate a csv file that contains the sum of all time spent at events for any number of user selected calendar layers.

Export Rendering- Added a button under share calendar that displays the relevant calendar layer  without sidebars or tedious UI elements. Can be printed or embedded in an iframe.


georgehrke:
The fact, that the data is stored in a special database table and not in CalDAV, is a reason not to merge it for us. All the data by the calendar should always be accessible by other caldav-clients as well. We don't want to provide a different view.

Changing it for this changes would require a complete rewrite of this patch.

Thanks

PR #859, opened 2018-07-02
update dependencies (both npm and bower)

georgehrke:


tcitworld:
You added a package-lock.json but we already have the yarn.lock file.

georgehrke:
removed the package-lock.json and disabled two style lint rules for now.

The scss need proper fixing and make use of proper nesting, but i didn't find any tool that would automatically do that. (`sass-convert`  can automatically nest from css to scss, but not reformat scss. And we already use scss features, so we can't just rename it back to css)

georgehrke:
@tcitworld can we get this in? :)

Issue #708, opened 2018-01-29
Feature request: Share multiple calendars in one link


Ich5003:
Hello,

I'd really love to see the following feature.
It would be awesome if you could share multiple calendars in one link.

For example I want to implement a calendar in an iframe on a website.
Let's say I have the two calendars "sport" and "meeting" with two different colors and I want to implement both of them on my website in one iframe, so the user sees all events with different colors.

Thank you,
Lucas

georgehrke:
As a first step we should just allow a user to pass multiple sharing tokens to the embed / public sharing site:

Right now it's like:
`http://nextcloud.local/index.php/apps/calendar/p/token1`
`http://nextcloud.local/index.php/apps/calendar/embed/token1`

The tokens are alphanumerical, so we can just use a dot (or a dash) as a separator and hand over a list of tokens like this:

`http://nextcloud.local/index.php/apps/calendar/p/token1.token2.token3`
`http://nextcloud.local/index.php/apps/calendar/embed/token1.token2.token3`

I'm not sure how to design a GUI that would allow a user to build these sharing URLs with multiple tokens (or whether that's necessary at all).

georgehrke:
If possible, we should do `http://nextcloud.local/index.php/apps/calendar/p/token1/token2/token3` but I'm not sure if that's possible with the Nextcloud routing.

Ich5003:
What about something like this?
![image](https://user-images.githubusercontent.com/1706621/35770037-abf88718-0914-11e8-98d2-1dba5e02569a.png)

georgehrke:
Well, I think the primary question should be how often this feature is gonna be used.

While I see a use-case for certain scenarios (mostly by advanced users) I don't see the average user ever use this feature. And if that's the case, we shouldn't add a new UI element as prominently as you suggested imho.

cc @nextcloud/designers

jancborchardt:
Yes, I agree @georgehrke. The url modification is already nice for sure. We can check how to better integrate this afterwards.

arunikayadav42:
Is somebody working on this or can I please  get started with this for my rgsoc 2018?

georgehrke:
@arunikayadav42 I'm not aware of anyone working on this, so please go ahead :) 

Don't hesitate to ask if you have a question or need any hint where to start!

arunikayadav42:
@georgehrke alright!thanks for approving the request!

arunikayadav42:
[...]

PR #865, opened 2018-07-11
Added possibility to share multiple calendars

Ich5003:
This PR implements a function requested in #708 
You can pass multiple share tokens split by a . to embed multiple calendars.
All calendars are shown in the default color at the moment. This may be impoved.


js/app/controllers/calcontroller.js
```diff
@@ -140,17 +140,28 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
...
+                        $scope.$apply();
+                    }
+                }).catch((reason) => {
+                    angular.element('#header-right').css('display', 'none');
+                    angular.element('#emptycontent-container').css('display', 'block');
+                });
+            });
 		}
 
+        $scope.integration = function () {
```
georgehrke: Why did you move this here?

js/app/controllers/calcontroller.js
```diff
@@ -140,17 +140,28 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
...
-				$scope.$apply();
-			}).catch((reason) => {
-				angular.element('#header-right').css('display', 'none');
-				angular.element('#emptycontent-container').css('display', 'block');
-			});
+		    $scope.calendars = [];
+		    var tokens = constants.publicSharingToken.split(".");
+		    var finished = 0;
+		    tokens.forEach(function(token){
+                $scope.calendarsPromise = CalendarService.getPublicCalendar(token).then(function(calendar) {
```
georgehrke: Please ident with tabs instead of spaces

js/app/controllers/calcontroller.js
```diff
@@ -140,17 +140,28 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
...
+                    finished++;
+                    $scope.calendars.push(calendar);
+                    if(finished >= tokens.length){
+                        is.loading = false;
+                        // TODO - scope.apply should not be necessary here
+                        $scope.$apply();
+                    }
+                }).catch((reason) => {
+                    angular.element('#header-right').css('display', 'none');
```
georgehrke: This error handling needs to be adapted.
You don't want to show the empty-container error page if one out of ten calendars fail. 
You would want some kind of Notification though. You can use `OC.Notification.show()` for that.

js/app/controllers/calcontroller.js
```diff
@@ -140,17 +140,28 @@ app.controller('CalController', ['$scope', 'Calendar', 'CalendarService', 'VEven
...
+                    angular.element('#emptycontent-container').css('display', 'block');
+                });
+            });
 		}
 
+        $scope.integration = function () {
+            return '<iframe width="400" height="215" src="' + $scope.$parent.root + 'embed/' + constants.publicSharingToken + '"></iframe>';
```
georgehrke: Not sure if angular handles escaping here.
What if you pass html as a parameter for `constants.publicSharingToken`?

This might result in an html injection.

Can you please check if its properly escaped?

js/app/controllers/calcontroller.js
Ich5003: Moved here, because otherwise the embed code will only contain one of the public sharing tokens.

js/app/controllers/calcontroller.js
Ich5003: Will check.

js/app/controllers/calcontroller.js
Ich5003: Will check

Ich5003:
I messed up commits, which automatically closed this PR.
Had to create a new one, see #870

PR #872, opened 2018-07-26
added language to simple event

Peterede:
Added language in simple event to store and retrieve the language from the event


georgehrke:
closing in favor of #883

PR #878, opened 2018-08-01
Federated calendar sharing

schiessle:
Implement federated calendar sharing

@georgehrke I started to add the needed methods and classes, but there is still some stuff to do to make it actually work. I tried to document it as good as possible. Maybe you can have a look and we can have a chat about it to clarify all remaining questions. Would be awesome if you would have some time to work on it, especially the UI and calendar specific part.

### DIFF
--- a/controller/federationcontroller.php
+++ b/controller/federationcontroller.php
@@ -1,0 +1,139 @@
+<?php
+/**
+ * @copyright Copyright (c) 2018 Bjoern Schiessle <bjoern@schiessle.org>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


controller/federationcontroller.php
```diff
@@ -0,0 +1,139 @@
...
+		$this->user = $userSession->getUser();
+	}
+
+	/**
+	 * @param string $url (should be the subscriber url, e.g. webcal://nextcloudserver/remote.php/dav/public-calendars/fkhskfhksf?export)
+	 * @param string $calendarName
+	 * @param string $recipient (the federated cloud ID of the recipient)
+	 * @return bool
+	 */
+	public function createFederatedShare(string $url, string $calendarName, string $recipient) {
```
georgehrke: This needs a dedicated route

georgehrke:
Just keeping open for reference.

This does not belong in the calendar repo, needs to go into dav app.

georgehrke:
Server issue: https://github.com/nextcloud/server/issues/1440

ChristophWurst:
I'm closing this PR as incomplete and stale. If the topic of federated calendar sharing is picked up again we can still take the code from the closed PR.

PR #882, opened 2018-08-03
Fix public layout rendering on Nextcloud 14

juliushaertl:
This fixes the public page layout for Nextcloud 14. Before this PR calendar was using public instead of base as layout value, which isn't equal with NC14 anymore.

I didn't move it to the newly introduced PublicTemplateResponse, since calendar still supports versions below 14.

Once 14 is a minimal requirement, we should move the public page with header to the new response.

Requires https://github.com/nextcloud/server/pull/10530

### DIFF
--- a/controller/viewcontroller.php
+++ b/controller/viewcontroller.php
@@ -155,7 +155,7 @@ class ViewController extends Controller {
 		$params = array_merge($templateParameters, $publicTemplateParameters);
 		$params['isEmbedded'] = true;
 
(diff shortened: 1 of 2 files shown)


rullzer:
https://github.com/nextcloud/server/pull/10530 is in

juliushaertl:
@tcitworld @georgehrke Anything blocking here?

georgehrke:
Not as far as i can tell. Will give this a review and merge tonight

PR #885, opened 2018-08-06
Quickfix for sidebar on NC14

### LINKED ISSUE #876: [Nc 14] Sidebar on adding a new event atop of calendar view
### Steps to reproduce
1. Have "skip simple event editor" enabled
2. be on 14 beta
3. add a new event

### Expected behaviour

The event sidebar  should appear on the right and decrease the width of the calendar view (as with Nc 13)

### Actual behaviour

The event sidebar is put on top, the calendar view is pushed down, below the fold.

### Server configuration

**Operating system**: Ubuntu 16.04

**Web server:** Apache

**Database:** MySQL 

**PHP version:** 7.0

**Server version:** 14.0.0.13

**Calendar version:** 1.6.1

**Updated from an older installed version or fresh install:** updated

**Signing status (ownCloud/Nextcloud 9.0 and above):** 👍

**Are you using an external user-backend, if yes which one:** LDAP

### Client configuration

**Browser:** FF 61.0.1

**Operating system:** Antergos

**CalDAV-clients:** WebUI

#### Browser log
```
Content Security Policy: The page’s settings blocked the loading of a resource at https://cloud.nextcloud.com/apps/files/ (“script-src”).

Deprecation warning: use moment.updateLocale(localeName, config) to change an existing locale. moment.defineLocale(localeName, config) should only be used for creating a new locale See http://momentjs.com/guides/#/warnings/define-locale/ for more info. sentry.js:1:7668

<script> source URI is not allowed in this document: “https://cloud.nextcloud.com/apps/notifications/js/notifications.js?v=59c93372-49”. calendar:44

Attempting to call a FullCalendar method on an element with no calendar.
sentry.js:1:7668
```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61544118-nc-14-sidebar-on-adding-a-new-event-atop-of-calendar-view?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tcitworld:
Closes #876

I didn't test on NC < 14, but it should be fine since it only puts back
old sidebar behaviour before nextcloud/server#10218

### DIFF
--- a/css/app/calendar.scss
+++ b/css/app/calendar.scss
@@ -26,6 +26,10 @@
 	width: 100%;
 }
 
+.modal-content #app-sidebar {
+	position: fixed;
+}
+
 #fullcalendar table {
 	white-space: inherit !important;
 }


tcitworld:
Maybe @skjnldsv can leave some input on how to handle `position: sticky;` better instead of returning to `position: fixed` as I did ? :)

skjnldsv:
@tcitworld sure, position sticky is the same as fixed instead in is based on the parent scrolling position.
Currently, the first scrollable parent is body, so you just need to make sure you don't add any overflow to any parent the sticky element have.

Nonetheless, since the calendar don't have any scrollable area? (right?)
Sticky shouldn't change anything I think. Let me have a look :)

skjnldsv:
Ah, the layout is not ok, the sidebar is supposed to be right next to the content
https://github.com/nextcloud/server/pull/9982

tcitworld:
> Nonetheless, since the calendar don't have any scrollable area? (right?)

Will have to check on that. :/

> Ah, the layout is not ok, the sidebar is supposed to be right next to the content

That's the issue I mentioned earlier with

>  The issue is that the editor panel (sidebar) is inside a popup over the content, not aside it.

I think it's because reducing the calendar's width makes it quite ugly that it's like that. Do you think the sidebar panel can be put aside and still get over the main content ?

I'll try to have a look on how other apps did with https://github.com/nextcloud/server/pull/9982.

skjnldsv:
@tcitworld the calendar is a bit aside from the others app :)
Maybe just add a manual fixed override like you did then.
Just comply to the first layout where only app-content is needed and not the #app one :)
See the contacts hack we did (since the new update won't have the same behavior, this is just a quick fix) https://github.com/nextcloud/contacts/blob/master/js/contacts-inject-14.js

tcitworld:
Need to review after https://github.com/nextcloud/calendar/pull/882 is merged.

georgehrke:
rebased

georgehrke:
@skjnldsv Do you know what change in the server caused this?
The behavior in the files app is correct. (if sidebar opens, width of filelist adapts)

georgehrke:
@skjnldsv Would you mind looking into this? It's probably just two or three lines of css.

skjnldsv:
> It's probably just two or three lines of css.

It's never that easy :grin: 
The app sidebar is not on the right location according to the standard, but seeing the overall layout, it would be far too hard to change everything.
This patch works fine, let's merge this for 14 and make sure the update to vue fit the new structure guidelines.

georgehrke:
Roger that!

PR #887, opened 2018-08-06
"Settings & Import" no longer covers bottom calendar.

### LINKED ISSUE #730: Scroll bar not long enough. Last calendar hidden from "Settings & Import"
### Steps to reproduce
A scroll bar is displayed for many calendars. In this case, the last subscribed calendar can no longer be reached via the scroll bar.

### Expected behaviour
Normally, the last calendar should appear when you scroll down.

### Actual behaviour
The last calendar is hidden by the link "Settings & Import".

platypro:
Fixes #730, tested with Firefox 61.0.1 and Chromium 68.0.3440.84 on Nextcloud 13

### DIFF
--- a/css/app/calendarlist.scss
+++ b/css/app/calendarlist.scss
@@ -320,7 +320,7 @@ $color-border: nc-darken($color-main-background, 8%) !default;
 }
 
 ul#calendarlistcontainer {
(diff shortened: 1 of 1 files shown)


georgehrke:
@platypro Thx for your pull-request!

PR #891, opened 2018-08-17
Feature/796/do not default new events

### LINKED ISSUE #796: use FullCalendar navLinks
All newly created events are set to "All day Event" by default.

I don't know how about you, but a vast majority of my calendar events are **not** All day events.

I'd like the All day event to be unchecked when creating a new event.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56932651-do-not-default-new-events-all-day-event?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

newhinton:
Fixes #796

This defaults the AllDay-State to True, but changes the state for new events to the last selected state. If it is toggled, the time is set to 12:00 to 13.00.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -34,6 +34,8 @@ return [
 		//Settings
 		['name' => 'settings#getConfig', 'url' => '/v1/config', 'verb' => 'GET'],
 		['name' => 'settings#setConfig', 'url' => '/v1/config', 'verb' => 'POST'],
+		['name' => 'settings#setLastChoosenDateState', 'url' => '/v1/allday/initial/set', 'verb' => 'GET'],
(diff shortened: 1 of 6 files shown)


newhinton:
@georgehrke I saw you are porting this app to vue, will you include this function directly or should i port this to vue later? Or should i fix this for angular?

georgehrke:
See https://github.com/nextcloud/calendar/issues/796#issuecomment-531193232

We will be doing it in a different way.
Thanks!

Issue #879, opened 2018-08-01
Show RSVP status of participants [$50]


dphi:
Related to #144 . If you can accept/ decline/ .. invitations, then it should also be possible to view the RSVP status of participants in the calendar app. 

Great would be to see the rsvp status of the event
- in the detail page of an event, and
- in the overview with a text hint like, eg,  '#accepts/#tentative/#declined/#not_responded' -> '1/2/0/5'

MisterY:
Adding some search keywords: attendees, participants, invitees ;)

pwFoo:
Yes, I would plan group events with nextcloud calendar and without a list of accepted / declined invites it isn't possible.
Also the notification (nextcloud app) should contain if the person accepted or declined the invite.

fert4342:
I would like to second this issue.

Without a way to see if attendees have accepted/declined an invitation, the calendar app is pretty much useless. Please fix this.

Thank you for your work

sebclemens:
Pleas support this. At least to send an email to the calendar owner to let him know about the response.

tbabej:
I agree this is critical, hence we put 50 USD bounty on this.

Upperholme:
We can't make use of Nextcloud's calendar if we can't track responses to event invitations. This looks to be a major omission.

bazzspeed:
Interestingly I can see the statuses in the Android Calendar App, so at least that's better than nothing.

georgehrke:
@bazzspeed Yes, the data from the user's response is obviously saved in the calendar data.
The only thing missing is showing that status in the web calendar. That's what this ticket is about

pwFoo:
Response is shown in my android calendar by email address (isn't a contact at the moment), but missing in web calendar.
Following here because it would be important for my use case.

georgehrke:
@pwFoo Please use the [Github Reactions](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) feature instead of commenting +1, 👍 etc. These kinds of comments just unnecessarily notify everyone subscribed to this issue. Thx! :)

PR #926, opened 2018-10-18
Calendar 2.0 - Vue

georgehrke:
# Components
## Sidebar
- [ ] collapseable Datepicker with buttons
- [ ] view buttons
- [ ] Today Button
- [ ] calendar-list
  - [ ] calendar-list-new
  - [ ] calendar-list-item
  - [ ] colorpicker
- [ ] subscription-list
  - [ ] subscription-list-new
  - [ ] subscription-list-item
  - [ ] colorpicker
- [ ] Settings
  - [ ] import-progressbar (similar to contacts)

## calendar grid
- [ ] fullcalendar

## editor 
- [ ] title
- [ ] timepicker
- [ ] details
- [ ] invitees-list
  - [ ] invitees-list-new (got some mockups, will upload later)
  - [ ] invitees-list-item (got some mockups, will upload later)
- [ ] alarm-list
  - [ ] alarm-list-new
  - [ ] alarm-list-item (got some mockups, will upload later)
- [ ] repeat
- [ ] how to allow other apps to provide their own module here? 

# Routes
- [ ] /{view}/{first-day-of-view}
- [ ] /{view}/{first-day-of-view}/edit/{popover|sidebar}/{calendar-uri}/{object-uri}/{recurrence-id-to-edit}
- [ ] /{view}/{first-day-of-view}/new/{popover|sidebar}/{recurrence-id-to-edit}

@tcitworld Do these routes make sense?


tcitworld:
What is `first-day-of-view` and why do we need it in our route ?
What would the values for `view` be ?

.eslintrc.js
```diff
@@ -0,0 +1,76 @@
...
+	},
+	parserOptions: {
+		parser: 'babel-eslint'
+	},
+	extends: [
+		'eslint:recommended',
+		'plugin:node/recommended',
+		'plugin:vue/essential',
+		'plugin:vue/recommended',
+		'standard'
```
tcitworld: I'm not so fond of `standard` (but no worries). Maybe it it should be - or not - inside Nextcloud coding style ? Do you want me to open an issue on server ?

georgehrke: I just copy pasted the basic stuff from the contacts app, happy to change it

skjnldsv: @tcitworld The overall edits I made on the standard really feels like what we usually encounter on the nextcloud apps :)
Could you take a look at the contacts app?

skjnldsv:
> How can we use webpack in dev mode since eval is now disabled ([nextcloud/server#11028](https://github.com/nextcloud/server/pull/11028)) ?
> Maybe `evalScriptAllowed` could be set to the `debug` system config.
> cc @rullzer @skjnldsv

Good question :)
Is there a dedicated webpack config for that?

tcitworld:
I should actually read articles instead of archiving them for later. Thanks ! 😀

.eslintrc.js
skjnldsv: @tcitworld btw, if you decide for a better settings that fit the way we usually use on our apps (we globally have a similar coding style between our apps :wink:) could you ping me so I can take a look and also update the contacts app to the same settings?

src/components/AppNavigation/ViewButtons.vue
```diff
@@ -1,11 +1,60 @@
...
 
 <script>
-    export default {
-        name: "ViewButtons"
-    }
+export default {
+	name: 'TodayButton',
+	computed: {
+		labelAgendaDay() {
+			return t('calendar', 'Day')
```
tcitworld: What do you think of a basic translation component ?

georgehrke: cc @skjnldsv ^

georgehrke:
First implementation of Settings area:

563d8395-2835-4530-995b-9557ba0d2ba5

@skjnldsv Some CSS love is always appreciated 😉

src/components/AppNavigation/ViewButtons.vue
skjnldsv: Well, we have the translate function used all over nextcloud and linked to the transifex component. I think it's working really well, so I don't think changing it is a good idea. It's easier to have the same solution all over nextcloud :)

tcitworld:
The renaming of `appinfo/application.php` to `appinfo/Application.php` gives me this :

```
{"Exception":"Error","Message":"Class 'OCA\\Calendar\\AppInfo\\Application' not found","Code":0,"Trace":[{"file":"server\/lib\/private\/legacy\/app.php","line":261,"function":"require_once"},{"file":"server\/lib\/private\/legacy\/app.php","line":154,"function":"requireAppFile","class":"OC_App","type":"::","args":["calendar"]},{"file":"server\/lib\/private\/legacy\/app.php","line":127,"function":"loadApp","class":"OC_App","type":"::","args":["calendar"]},{"file":"server\/lib\/base.php","line":972,"function":"loadApps","class":"OC_App","type":"::","args":[]},{"file":"server\/index.php","line":42,"function":"handleRequest","class":"OC","type":"::","args":[]}],"File":"server\/apps\/calendar\/appinfo\/app.php","Line":24,"CustomMessage":"--"}

```

skjnldsv:
> @skjnldsv Some CSS love is always appreciated

Ahah! Well, we still have no standard for the settings area, so I cannot really help.
I would say it looks ok now. Not too crowded, not too many options. Keep it like that :+1: 

On a side note, why is there two different caldav url for mac and others? Shall I do the same on contacts?

georgehrke:
@tcitworld @skjnldsv How should we store events in vuex?

Right now (in this PR) we have this `events` property on calendars, similarly to how the new Contacts app does it, but it doesn't really make sense imho. We can't possibly store all events a user has in there. It's not uncommon to have tens of thousands or even hundreds of thousands of events. Permanently storing everything in memory could cause real memory issues.

Even after importing 10000 events, it doesn't make sense to keep them all in memory if only 5 of them have to be displayed by the current view.

Use-cases where we need to access events:
1. display events from the current time-range. Required time-range is provided by full calendar.
  - we just need to query them and hand over well-formed objects to full calendar
  - theoretically there is no need to store them in vuex at all
  - it would still be good to store them (but where?), so we can access them faster for opening the editor
2. a user clicks an event to edit / view it
  - we need to get the full event data and display it in the editor
3. a user is forwarded to the calendar by a different app with a route to an event. In that case, if no recurrenceId is given, we need to figure out the next in future / closest occurrence in the past
  - we need to get the full event data and display it in the editor

I was thinking about having an editEvent store, that's usually empty and only stores the data of one `VEVENT` when it's being edited.

Does that make sense?
What's your opinion on this?

georgehrke:
> On a side note, why is there two different caldav url for mac and others?

If the server is not configured correctly (`.well-known/caldav` etc.), then you need to point macOS straight to the user principal address.

> Shall I do the same on contacts?

Yes :)

georgehrke:
> I was thinking about having an editEvent store, that's usually empty and only stores the data of one VEVENT when it's being edited.

> Does that make sense?
What's your opinion on this?

Also cc @raimund-schluessler 
How do you manage tasks? Do you load all at once?

skjnldsv:
Yes, contacts and events behave very differently. I don't think you should indeed keep all the events in the storage. Though it could be interesting to think about the most common use case. How often a user go back in time? How often the user checks the next month? And the Month after that?

I would say preloading the month with the full data and then only get the simplest data like day, hour, duration for the next incoming months is a nice idea. Then, like contacts, fetch the full data depending on how full the current view displays the events? I'm not the best qualified on VEVENTS and associated entries since it's not my field, but I'd say there is some optimisation to think about.
Though, storing 1k events for the sake of preloading is not a proper optimlisation, I'll agree with you on this :)

skjnldsv:
@georgehrke then the idea of preloading still stands, but not on the props but on the quantity  I guess :)

raimund-schluessler:
> How do you manage tasks? Do you load all at once?

For the Tasks app I will keep the strategy I currently use. On app load all VTODOs which are not completed are loaded like so
```
<c:calendar-query xmlns:c="urn:ietf:params:xml:ns:caldav" xmlns:d="DAV:" xmlns:a="http://apple.com/ns/ical/" xmlns:o="http://owncloud.org/ns">
   <d:prop><d:getetag/><c:calendar-data/></d:prop>
   <c:filter>
      <c:comp-filter name="VCALENDAR">
      <c:comp-filter name="VTODO"><c:prop-filter name="COMPLETED"><c:is-not-defined/></c:prop-filter>
      </c:comp-filter>
      </c:comp-filter>
   </c:filter>
</c:calendar-query>
```

This is necessary in order to calculate the number of tasks due today/this week/which are important. When you open the details of a task, all subtasks which are already completed are loaded as well:
```
<c:calendar-query xmlns:c="urn:ietf:params:xml:ns:caldav" xmlns:d="DAV:" xmlns:a="http://apple.com/ns/ical/" xmlns:o="http://owncloud.org/ns">
   <d:prop><d:getetag/><c:calendar-data/></d:prop>
   <c:filter>
      <c:comp-filter name="VCALENDAR">
      <c:comp-filter name="VTODO"><c:prop-filter name="COMPLETED"/><c:prop-filter name="RELATED-TO"><c:text-match>xx5bjzqwb9</c:text-match></c:prop-filter>
      </c:comp-filter>
      </c:comp-filter>
   </c:filter>
</c:calendar-query>
```

(On a side note: I haven't checked if this is easily possible with the cdav-library already. If not, would be good to add it.)

I store all tasks in the Vuex store. And I also save the subtasks of a task in the tasks properties (as of https://github.com/nextcloud/tasks/commit/ecdcb827ef4a27b6d0edab396b982864f028a3b6). Otherwise all tasks would be re-rendered when one changes. I think for Tasks, this is the best strategy. Loading all tasks would absolutely not work fast enough.

For calendar, loading all VEVENTs is not a good idea either, I agree. Your strategy sounds fine. Maybe preloading the events for the previous and next week could be an option to speed up navigation. Loading more is not necessary in my opinion. Storing them in the Vuex store would be ok, but maybe not necessary if you have another idea.

Side note 2:
Vue offers a very nice interface for checking the performance of an app. It can be enabled with
```
Vue.config.devtools = true
Vue.config.performance = true
```
on `main.js` (see https://github.com/nextcloud/tasks/commit/26001585d878b42d254ce119c6cefb1f99d959d9)
In Chrome you then see which components are rendered in which order and how long it took:

vueusertiming

This helped a lot in speeding up the Tasks app. Although rendering a list with 60 tasks still takes ~400 ms.

tcitworld:
> Maybe preloading the events for the previous and next week could be an option to speed up navigation.

Agreed with always fetching events for previous, current, and next period (same for days and months).

georgehrke:
I had an idea about the caching thing that i would like to get some feedback on:
- Along every event we store a time to live
- initial ttl is 10
- Whenever we request events from the server (no matter whether it's a single event or a bunch of events), the time to live is decremented.
- if the ttl is 0, the event is purged from the current cache
- if an event was already cached, but is part of the server response, the ttl is set back to 10
- events from this month, the previous month and the upcoming two months always have a ttl of -1 and are never removed from cache

Any opinion @tcitworld @skjnldsv @raimund-schluessler

skjnldsv:
Seems like a nice idea!!
The only thing I'm conflicted about is the ux. What if I see a view with an event I just deleted on my phone. How would the user know that this is not a bug? Could we at least fetch all of the events uid? To make ensure deletion is up-to date? What about fetching REV and UID, and only update the events that are changed?

georgehrke:
I decided to implement our own FullCalendar 4 Timezone implementation: https://github.com/nextcloud/calendar/pull/926/commits/8e9aa9e4c4d568c8364056dae6079f3bbe937d15#diff-51210ecedb0c511ac616572de6c0bf45

Moment Timezone with all timezones is 917KB and we already have all the timezone data inside our icalendar VTimezone provider, so it's great we can now live without moment timezone in FullCalendar 4.

georgehrke:
c343c866-1202-406a-8e5a-6caf1e6b2c9b

We now show the owner avatar next to a shared calendar

src/components/AppNavigation/CalendarListItemSharingPublishItem.vue
```diff
@@ -0,0 +1,264 @@
...
+					}, 2000)
+				})
+		},
+		copySubscriptionLink() {
+			// change to loading status
+			this.copyingSubscriptionLink = true
+			event.stopPropagation()
+
+			const rootURL = OC.linkToRemote('dav')
+			const url = new URL(this.calendar.publishURL + '?export', rootURL)
```
georgehrke: TODO: replace protocol with `webcal` respectively `webcals`

jancborchardt:
How to try this out? `make` doesn’t run through:
```
~/nextcloud/apps/calendar$ make 
rm -f js/calendar.js
rm -f js/calendar.js.map
rm -rf node_modules
npm install

> node-sass@4.11.0 install /home/jan/nextcloud/apps/calendar/node_modules/node-sass
> node scripts/install.js

Cached binary found at /home/jan/.npm/node-sass/4.11.0/linux-x64-67_binding.node

> node-sass@4.11.0 postinstall /home/jan/nextcloud/apps/calendar/node_modules/node-sass
> node scripts/build.js

Binary found at /home/jan/nextcloud/apps/calendar/node_modules/node-sass/vendor/linux-x64-67/binding.node
Testing binary
Binary is fine
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

added 1438 packages from 793 contributors and audited 26561 packages in 79.619s
found 0 vulnerabilities

npm run lint

> calendar@2.0.0-alpha1 lint /home/jan/nextcloud/apps/calendar
> eslint --ext .js,.vue src tests


Oops! Something went wrong! :(

ESLint: 5.10.0.
No files matching the pattern "tests" were found.
Please check for typing mistakes in the pattern.

npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! calendar@2.0.0-alpha1 lint: `eslint --ext .js,.vue src tests`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the calendar@2.0.0-alpha1 lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/jan/.npm/_logs/2018-12-20T20_52_48_602Z-debug.log
Makefile:45: recipe for target 'lint' failed
make: *** [lint] Error 2
```

tcitworld:
`npm run dev` or `npm run build` should do it in the meantime.

jancborchardt:
@tcitworld thanks!

Some initial issues:
- [ ] The personal and contact birthdays calendar is not enabled by default, should be
- [ ] Clicking the date opens the date picker, but clicking again does not close it again

But I guess there’s lots more, renaming doesn’t work etc.

tcitworld:
It wasn't claimed to be ready at all. ;)

raimund-schluessler:
Do you plan to support IE11 with this rewrite? The Tasks rewrite currently does not do it, and I don't have plans to do so, except it is very easy 😉

georgehrke:
> Do you plan to support IE11 with this rewrite? The Tasks rewrite currently does not do it, and I don't have plans to do so, except it is very easy 😉

Yes, I do. Should be fairly easy with Babel Polyfill I guess.

skjnldsv:
> Yes, I do. Should be fairly easy with Babel Polyfill I guess.

The secret is in the *should* word :see_no_evil:

bazzspeed:
Can this also be fixed #879 here ?

tcitworld:
FYI, https://github.com/webpack/webpack/issues/8656

georgehrke:
We should probably consider using https://fullcalendar.io/docs/vue

jakobroehrl:
Can anybody help here: 

grafik

Thanks!

tcitworld:
>  alarm-list-item (got some mockups, will upload later)

Never got a look at them. :(

src/models/calendar.js
```diff
@@ -78,10 +78,16 @@ export const getDefaultCalendarObject = (props = {}) => Object.assign({}, {
  * @returns {Object}
  */
 export function mapDavCollectionToCalendar(calendar) {
+	let color = calendar.color
```
tcitworld: Apparently `color` is not fetched by cdav for some reason and leads to undefined here.

tcitworld:
Let's make `illustrationProviderService` depend on `categories` as fallback if the title hasn't matched anything.

georgehrke:
> Let's make `illustrationProviderService` depend on `categories` as fallback if the title hasn't matched anything.

Good idea, but let me open a dedicated issue for that.

georgehrke:
@tcitworld I'm just finishing up the last embedding and read-only related things.
Would you be fine with it if i just go ahead and merge it tomorrow afternoon?

tcitworld:
Absolutely !

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

tcitworld:
The PR was first opened exactly one year ago. 😅

georgehrke:
I branched off master and turned it into a protected branch: https://github.com/nextcloud/calendar/tree/stable1.7

UtechtDustin:
Is there any roadmap for the release of this version ?

georgehrke:
Please see https://github.com/nextcloud/calendar/milestones

UtechtDustin:
That was quick! Thanks.

Ali-romdhani:
Hi, great app and needful updates!
Can you tell me how to change de default view on a an iframe ?
This is the iframe code proposed on the app, what should be added ?
`<iframe src="https://niceurl/apps/calendar/embed/xxxxxxxxxxxxxxxxx" width="400" height="800"></iframe>`
Many thanks

Issue #157, opened 2016-10-25
[Feature Request] Keyboard support


eppfel:
We try to make nextcloud more user friendly with a consistent way of keyboard control: nextcloud/server#1444

Are or were there keyboard shortcuts besides the browser defaults and I missed them?

Any proposals?
- [x] tabindex imrprovement #25 
- [ ] new event
- [ ] ...

georgehrke:
I always wanted `h, j, k, l` to navigate thru the calendar grid 😁

georgehrke:
@eppfel What would you suggest for new Event? 

simply Ctrl / Cmd + N ?

mammuth:
> simply Ctrl / Cmd + N ?

As stated in the linked issue at the server-repo, `Ctrl/Cmd` combinations should be avoided I think. 
`Ctrl/Cmd + N` is used by most browsers to open a new window, and should not be catched by the calendar app in my opinion.

calendar.google.com uses `C` (create/compose) for example, which I really like.

bpampel:
In general I would suggest adapting the keyboard shortcuts of google calendar, as it would also make migrating from there much easier. 
Not all of them are really that important but I definitely miss pressing "t" to get to the present date, as well as shortcuts for creating and editing events.

georgehrke:
@bpampel Do you know if there is a complete list of Google Calendar shortcuts? :)

bpampel:
I am not sure if it contains all shortcuts, but the help has a page on that: https://support.google.com/calendar/answer/37034?hl=en

holocronweaver:
Most important keys for me: `Delete` to delete an event. `Enter` to confirm event. `d` for day view, `w` for week, and so on. `c` to create an event.

Except for select-drag to create an event or click to select an event, I shouldn't have to use the mouse at all.

In the future, type to select event and type to create event would allow Calendar to become completely keyboard driven. That would beat Google hands down. But for now I'd be happy with basic shortcuts.

cls-nebadje:
I really dig the google shortcuts - but even other shortcuts are really welcome!

jaller94:
I am strongly in support for adding shortcuts and using the same as in Google Calendar.

This is a screenshot of their Keyboard Shortcut popup (accessible by pressing `Tab Tab Enter`):
![](https://user-images.githubusercontent.com/10872136/47250249-40970780-d3d3-11e8-9751-bfb24d7070f9.png)

georgehrke:
[...]

PR #929, opened 2018-10-20
Add accesskeys to the datepicker container buttons

jaller94:
As it had been mentioned in #157, I would prefer keyboard shortcuts which do not require combinations. There are also accessibility issues known when using `accesskey` described on MDN:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey

Still, I have the opinion that this would be a great quick fix to make some sort of keyboard shortcuts available until there is a JavaScript handling available.

- `P` for the **p**revious period
- `N` for the **n**ext period
- `D` for the **d**ay view
- `W` for the **w**eek view
- `M` for the **m**onth view
- `T` for the current day (**t**oday)

### DIFF
--- a/templates/part.buttonarea.php
+++ b/templates/part.buttonarea.php
@@ -24,10 +24,10 @@
 ?>
 
 <div class="togglebuttons">
-	<button class="button first" ng-value="agendaDay" ng-model="selectedView" uib-btn-radio="'agendaDay'"><?php p($l->t('Day')); ?></button>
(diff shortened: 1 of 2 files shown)


jaller94:
I noticed that Nextcloud Mail has a page for keyboard shortcuts. It is hidden in the settings.
https://github.com/nextcloud/mail/blob/master/js/templates/keyboard-shortcuts.html

And from the main app I found a file that handles most of the single key shortcuts:
https://github.com/nextcloud/server/blob/master/apps/files/js/keyboardshortcuts.js

georgehrke:
Closing because of the restrictions with accesskey.
This needs to be reimplemented in the Vue branch and we should use that opportunity to use some proper Vue Plugin for that.

PR #936, opened 2018-11-02
upgrad packages, fixing build

blizzz:
Failed with "ReferenceError: internalBinding is not defined" previously. Now, after a `yarn upgrade` the build works, the calendar can be enabled, and used (brief smoke testing).

Apart from that, I get those warnings in the build process:

```
Invalid Option: Unexpected option value "false" for rule "no-descending-specificity"
Invalid Option: Unexpected option value "false" for rule "no-duplicate-selectors"
```

The values are set in `js/.stylelintrc` and per documentation, true would be the only valid value. However, removing those lines would break the build process… so I left them in.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -56,9 +56,12 @@
     "angular": "^1.7.2",
     "angular-ui-bootstrap": "^2.5.0",
     "fullcalendar": "^3.9.0",
+    "hsl-to-rgb": "kayellpeee/hsl_rgb_converter",
     "hsl_rgb_converter": "kayellpeee/hsl_rgb_converter",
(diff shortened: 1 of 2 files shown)


blizzz:
> I didn't have any issues building recently (aside from those two warnings). Did you just executed make or did things manually ?

I pulled this morning and ran `make` afterwards, nothing else.

> Maybe I don't have the same yarn version, but I fail to see why the packages names are changed.

I got yarn 1.12.1

georgehrke:
@tcitworld This seems to be related to the underlying node version. Can you verify that this doesn't break the build on your node-version?

Issue #938, opened 2018-11-03
Styling broken with Nextcloud 14 using Dark theme


mathroc:
The stylesheet is not compatable with the Dark Theme introduced in Nextcloud 14

gerroon:
Yes please take a look at this, thge calendar ap has white day boxes all over :(

PR #941, opened 2018-11-11
Fix dark theme

tcitworld:
Mostly use CSS4 variables instead of SASS ones.

Closes https://github.com/nextcloud/calendar/issues/938


skjnldsv:
> Will the CSS4 variables be automatically replaced by our SASS compiler/transpiler or will this only be supported by modern brothers? (Since we still aim to support IE11)

No, ie11 supports css4 vars with [our ponyfill](https://github.com/nextcloud/server/blob/9eb5ab9e7072c75687c361791e332b4b2e348b6e/core/js/core.json#L19)

PR #949, opened 2018-11-23
Fix position of New Calendar button in IE11

georgehrke:


georgehrke:
Failing tests unrelated

PR #950, opened 2018-11-23
prevent delivering broken ics data via proxycontroller

georgehrke:


georgehrke:
failing tests unrelated

PR #1045, opened 2019-01-04
Allow thirdparty apps to hook the calendar app

newhinton:


newhinton:
That's great news! I found this usefull to write small extensions for apps. I hoped that this could be part of an update before the vue-release, so that i can easily extend this app until vue is released (and maybe even then, who nows? ;) )

newhinton:
There was a bug, now it is fixed :D

georgehrke:
Thx for your contribution!

PR #1085, opened 2019-01-28
Fix datepicker alignment

tcitworld:
Follow up to #1043 

## Before
(latest release, without #1043 and round buttons)
screenshot_2019-01-28 agenda - mon petit nuage

## After
Capture d’écran du 2019-03-17 11-38-11


georgehrke:
@skjnldsv nope

tcitworld:
Travis going wild but it shouldn't matter.

tcitworld:
@jancborchardt Addressed.
Capture d’écran du 2019-03-17 11-38-11

PR #1216, opened 2019-04-03
backport webcal access rules from DAV app implementation

georgehrke:


tcitworld:
Since we didn't document `webcalAllowLocalAccess` and it's not a toggle setting in the admin interface, is anyone using this at all ?

georgehrke:
> Since we didn't document webcalAllowLocalAccess

We should probably have a toggle in the admin settings of the dav app. Will send a pull-request for that.

Issue #678, opened 2017-12-12
update calendar 'today'


luzik:
### Steps to reproduce
1. Login into nexcloud webpage
2. Goto calendar month view
3. Wait till nextday

### Expected behaviour
Current day 'yellow background' show today

### Actual behaviour
Current day 'yellow background' show yesterday

georgehrke:
This is actually something we should  do upstream. Will report a bug soon.

ideadapt:
The following existing issue in upstream addresses this issue https://github.com/fullcalendar/fullcalendar/issues/3783. 
They treat that as a minor issues, whereas for me, this annoys me almost every day, since my nextcloud calendar is a pinned tab.

PR #1233, opened 2019-04-14
Update UI after gotoDate call of today button action

ideadapt:
Addresses issue #678.
Does not contain a solution to automatically update current date & ui at midnights.

This fix can easily be applied to the current AngularJS calendar implementation, just let me know if desired :)

### DIFF
--- a/src/components/FullCalendar.vue
+++ b/src/components/FullCalendar.vue
@@ -150,6 +150,8 @@ export default {
 
 			if (params.firstday !== this.currentDate) {
 				this.calendar.gotoDate(params.firstday)
+				this.calendar.render()
+				this.calendar.updateSize()
 				this.currentDate = params.firstday
 			}
 		}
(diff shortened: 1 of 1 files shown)


tcitworld:
> This fix can easily be applied to the current AngularJS calendar implementation, just let me know if desired :)

Please do!

ideadapt:
@tcitworld I was too optimistic. I originally have had applied the fix to the latest fullcalendar jquery demo. But the fullcalendar dependency version in nextclouds calendar app is not the latest one, maybe this is the reason the same approach does not work there. I tried to upgrade the dependency, but this caused too many other incompatibilities.
Any way I think this PR should be ready to be reviewed / merged?

georgehrke:
Closing, because it doesn't completely fix #678.

This needs to be handled with a `window.setTImeout` that keeps checking the date and rerenders if necessary. Let me take care of that in the vue branch.

PR #1371, opened 2019-07-25
Feautre/879 Show RSVP status of participants

rgergo67:
As requested in #879 users are able to view the RSVP status of participants in the calendar app in the detail page of an event and in the overview modal. 

It will look like this:
image
image

When hovering over a color, it shows the translated status text, like Accepted, Declined, Tentative or Needs action. If no participants are added to the event, the Participant responses row will be hidden.


georgehrke:
Hi @rgergo67.

Thx for your contribution.
I'm super sorry to say this now, but it seems you didn't see our note in the readme file:
> __We are currently moving the Calendar app from Angular 1 to Vue. (see [this PR](https://github.com/nextcloud/calendar/pull/926)). Please develop your bugfixes / features against the `vue` branch for the time being__ 

I'd be happy to port this change into our vue branch once ready, but i don't think it's something we can merge at the moment. cc @tcitworld 

Some remarks on your pull-request:
You are only using colours, among them red and green, which will cause trouble for people with color blindness. On top of using colours, we should use of different shapes.

Like a green circle for accepted.
A yellow triangle for tentative
a red square for declined
and some other grey shape for no response so far.

rgergo67:
Hi @georgehrke 

no problem, my mistake, I have just started to learn this open source contribution thing, don't really understand everything. We needed this function in our own calendar anyway until the next version comes out.

You are totally right about the color blind thing, I haven't thought about that. At first I made them with + - * and ?, but thought that colors would be better. I will try to combine them, and see if it looks good.

rgergo67:
Added some icons, now it looks like this:

image

image

tcitworld:
> I'd be happy to port this change into our vue branch once ready, but i don't think it's something we can merge at the moment. cc @tcitworld

Well, it's you who handles the release process so I would say it's mostly up to you to decide.
However I would think it would be acceptable to have another release with this feature until the 2.0 release.

In any case, this does not prevent us from working on this PR from an UX PoV.

The icons are not really appropriate (➕ and ➖) respectively meaning adding and removing elements. ✳ is not really defined.
What I suggest: `✅`, `❌`, `(✅)` (this one in yellow) and `❓`.
We may also style the background depending on the status.
Perhaps participation summary could be located on top.

Asking for @jancborchardt's help here. :)

jancborchardt:
Thanks for the contribution @rgergo67! :)

Fully agree with @georgehrke on that ideally this should be on the Vue branch if possible (I can recommend this course for Vue: https://scrimba.com/g/glearnvue) and also agree with @tcitworld specifically on the suggested icons. :+1:

rgergo67:
What do you think about these:
image
Participation summary located on top, icons replaced
image

If I change the background of the names too, they look like this:
image

If I change the color of the whole row, it hurts my eyes
image

raimund-schluessler:
> Participation summary located on top, icons replaced
> image

I like this best. But for the checkmark and the cross, I think there are Nextcloud default symbols (I might be wrong).

rgergo67:
With the latest commit it looks like this:
image
image

georgehrke:
961130EE-0DAB-48B3-BCAF-CA6009133EFB

We implemented this along with avatars in the vue branch.
Closing since this is no longer needed.
Thanks for your contribution anyway and for the discussion.

PR #1400, opened 2019-08-14
Fix travis database tests

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ dist: trusty
 sudo: required
 language: php
 php:
-  - 7.0
   - 7.1
   - 7.2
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,16 +5,11 @@ php:
   - 7.0
```
tcitworld: Also remove PHP 7.0 support.

.travis.yml
```diff
@@ -5,16 +5,11 @@ php:
...
   - 7.1
   - 7.2
```
tcitworld: Add PHP 7.3 support.

PR #1425, opened 2019-08-28
show sidebar values

jakobroehrl:
todo

### DIFF
--- a/src/components/Editor/Timezonepicker.vue
+++ b/src/components/Editor/Timezonepicker.vue
@@ -1,0 +1,53 @@
+<template>
+	<div>
+		<div v-if="show">
+			<Multiselect :options="timezones" />
+			<Multiselect :options="timezones" />
+		</div>
+		<button v-tooltip="t('calendar', 'set timezone')" class="icon-timezone" @click="show = !show" />
+	</div>
+</template>
+
+<script>
+import { Multiselect } from 'nextcloud-vue'
+export default {
+	name: 'Timezonepicker',
+	components: {
+		Multiselect
+	},
+	props: {
+		eventComponent: {
+			type: Object,
+			default: false
+		}
+	},
+	data() {
+		return {
+			show: false,
+			timezones: ['de', 'en']
+		}
+	},
+	computed: {
+		timeFormat() {
+			// if (this.eventComponent.isAllDay) {
+			if (true) {
+				return 'YYYY-MM-DD'
+			}
+
+			return 'YYYY-MM-DD HH:mm'
+		},
+		timeType() {
+			// if (this.eventComponent.isAllDay) {
+			if (true) {
+				return 'date'
+			}
+
+			return 'datetime'
+		}
+	}
+}
+</script>
+
+<style scoped>
+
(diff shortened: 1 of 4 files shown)


src/views/EditSidebar.vue
```diff
@@ -14,7 +14,18 @@
...
+				<input :value="location" type="text" placeholder="t('calendar', 'location')">
+			</div>
+			<div class="section-wrapper">
+				<textarea :value="description" placeholder="t('calendar', 'description')" />
+			</div>
+			<div class="section-wrapper">
+				<input :value="status" type="text" placeholder="t('calendar', 'status')" />
+			</div>
+			<div class="section-wrapper">
+				<input :value="accessClass" type="text" placeholder="t('calendar', 'accessClass')" />
```
georgehrke: You have to use v-model instead

src/views/EditSidebar.vue
```diff
@@ -80,6 +91,26 @@ export default {
 		subtitle() {
 			return this.isLoading ? 'LOADING' : 'LOADED'
 		},
+		location() {
```
georgehrke: You have two options. Either you use a computed property with get(){} and set(value){} and set the new value inside set.
Or you add `@change="setNewValue"` on the input element and create a new setNewValue method,

src/components/Editor/Timezonepicker.vue
```diff
@@ -0,0 +1,53 @@
...
+			<Multiselect :options="timezones" />
+		</div>
+		<button v-tooltip="t('calendar', 'set timezone')" class="icon-timezone" @click="show = !show" />
+	</div>
+</template>
+
+<script>
+import { Multiselect } from 'nextcloud-vue'
+export default {
+	name: 'Timezonepicker',
```
georgehrke: We already have a Timezone selection component: https://github.com/nextcloud/calendar/blob/vue/src/components/Shared/TimezoneSelect.vue

src/views/EditSidebar.vue
```diff
@@ -14,7 +13,23 @@
 		</template>
 
 		<AppSidebarTab name="Details" icon="icon-details" :order="0">
-			This is the details tab
+			<title-timepicker :event-component="eventComponent" />
+			<timezonepicker :event-component="eventComponent" />
+			<div class="section-wrapper">
+				<input v-model="location" type="text" placeholder="t('calendar', 'location')">
```
georgehrke: You should use a computed property for the translation.
Additionally, if you want the value to be evaluated as javascript, you need to use :placeholder.

src/views/EditSidebar.vue
```diff
@@ -14,7 +13,23 @@
 		</template>
 
 		<AppSidebarTab name="Details" icon="icon-details" :order="0">
-			This is the details tab
+			<title-timepicker :event-component="eventComponent" />
```
georgehrke: Title and time-picker should not be in the details tab, but inside primary actions, so they are always visible.

georgehrke:
Rebased, let's merge this.

PR #1447, opened 2019-09-02
Don't ship special build with babel for IE anymore

### LINKED ISSUE #1286: Stop shipping babel-polyfill, it's part of the server since 16
### Steps to reproduce
1. Open /apps/calendar/ in Internet Explorer 11 with Calendar 1.7 and Nextcloud 16

### Expected behaviour
Calendar should appear

### Actual behaviour
Page does not load at all (white screen). Following error is shown in developer tools.
```
Error: only one instance of babel-polyfill is allowed
```

As a workaround I could get it to work by using the default vendor js file instead:
```
cd apps/calendar/js/public
mv vendor.ie.min.js vendor.ie.min.js.old
mv vendor.js vendor.ie.min.js
```

Seems this is a known issue for NC16, already found by some other apps as well. See https://github.com/nextcloud/server/issues/14111

georgehrke:
fixes #1286 

### DIFF
--- a/js/gulpfile.js
+++ b/js/gulpfile.js
@@ -40,8 +40,6 @@ const cssBuildTarget = 'app.scss';
 const cssBuildTargetMin = 'app.min.scss';
 const vendorTarget = 'vendor.js';
 const vendorTargetMin = 'vendor.min.js';
-const vendorIETarget = 'vendor.ie.js';
-const vendorIETargetMin = 'vendor.ie.min.js';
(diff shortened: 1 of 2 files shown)


tcitworld:
Remove them as well from Gulp?

georgehrke:
Indeed. Done

pm-pm:
Now the Calender works again with IE11.
Great. :-)

PR #1472, opened 2019-09-11
circle sharing

jakobroehrl:
Signed-off-by: Jakob <jakob.roehrl@web.de>

### DIFF
--- a/src/components/AppNavigation/CalendarListItemSharing.vue
+++ b/src/components/AppNavigation/CalendarListItemSharing.vue
@@ -29,6 +29,7 @@
 <script>
 import client from '../../services/cdav'
 import debounce from 'debounce'
+import HttpClient from 'nextcloud-axios'
 
 import CalendarListItemSharingPublishItem from './CalendarListItemSharingPublishItem.vue'
 import CalendarListItemSharingItem from './CalendarListItemSharingItem.vue'
@@ -89,6 +90,7 @@ export default {
 		 * @param {String} query
 		 */
 		findSharee: debounce(async function(query) {
+			let circles = this.loadCircles()
 			this.isLoading = true
 			this.usersOrGroups = []
 			if (query.length > 0) {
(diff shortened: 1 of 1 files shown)


src/components/AppNavigation/CalendarListItemSharing.vue
```diff
@@ -89,6 +90,7 @@ export default {
 		 * @param {String} query
 		 */
 		findSharee: debounce(async function(query) {
+			let circles = this.loadCircles()
```
georgehrke: this should go inside the `if (query.length > 0) {` block

jakobroehrl:
@georgehrke 
Thanks for your hints, implemented it. Now I get the circles. 
The last problem is, how to get the results all in one list and return it?

src/components/AppNavigation/CalendarListItemSharing.vue
```diff
@@ -92,8 +93,15 @@ export default {
 			this.isLoading = true
 			this.usersOrGroups = []
 			if (query.length > 0) {
-				const results = await client.principalPropertySearchByDisplayname(query)
-				this.usersOrGroups = results.reduce((list, result) => {
+				
+				const dav = client.principalPropertySearchByDisplayname(query)
+				const circles = this.loadCircles()
+
+				Promise.all([dav, circles]).then(function(results) {
```
georgehrke: The function, that you hand over to `Promise.all([...]).then()` gets exactly the same amount as parameters as you hand over in the array to `Promise.all()`.

So in this case you get two parameters: `Promise.all([dav, circles]).then(function(davResults, circlesResults) {` :)

src/components/AppNavigation/CalendarListItemSharing.vue
```diff
@@ -105,14 +113,40 @@ export default {
 						})
 					}
 					return list
-				}, [])
+				}, []) 
+
+				let circles = resultCircles.data.ocs.data.circles
```
georgehrke: All this circles stuff needs to move inside the `.then()` block. 
And you will have to push stuff into the list array.

jakobroehrl:
Hi Georg, 
no problem. Please ping me, if I can support you further...

PR #1503, opened 2019-09-30
Add user agent to guzzle requests

PaulKiddle:
I'm facing the same issue also raised [here](https://help.nextcloud.com/t/adding-a-unique-user-agent-string-for-guzzlehttp-requests-in-calendar-app-feat-songkick-case-study/40001) of Songkick rejecting cal requests with Guzzle's default UA string. This is the proposed fix.

### DIFF
--- a/controller/proxycontroller.php
+++ b/controller/proxycontroller.php
@@ -140,6 +140,9 @@ class ProxyController extends Controller {
 			do {
 				$clientResponse = $client->get($queryUrl, [
 					'allow_redirects' => $allow_redirects,
+					'headers' => [
+						'User-Agent' => 'Nextcloud',
+					],
 				]);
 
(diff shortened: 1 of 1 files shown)


georgehrke:
The proxyController is deprecated and will be removed very soon. Please send your pull-request against this file: https://github.com/nextcloud/server/blob/master/apps/dav/lib/BackgroundJob/RefreshWebcalJob.php#L266

Thx!

tcitworld:
@georgehrke Isn't it already handled here anyway? https://github.com/nextcloud/server/blob/master/apps/dav/lib/BackgroundJob/RefreshWebcalJob.php#L207

georgehrke:
Ah. Indeed. So no need for a pull-request.

@PaulKiddle Thx for your contribution anyway! :)

Issue #748, opened 2018-03-01
Use webcals URI scheme when accessing via HTTPS (PFS directive)


Somebodyisnobody:
### Steps to reproduce

1. Create and share a calendar
2. Open shared calendar
3. Click right on "subscribe" to copy the link of the button and paste it somewhere to take a look on it.

### Expected behaviour
Nextcloud applies PFS to the webcal-link and forms it into an webcals-link

### Actual behaviour
If i access via TLS the subscribtion link is still webcal://

I cannot find an offical URI scheme but all of my clients does support the webcals://-sheme
Some examples:
- CalDAV for Android
- WebcalSync
- Thunderbird (acc. to serverlog)
- Outlook (acc. to serverlog)

@georgehrke said in https://github.com/nextcloud/server/issues/8039#issuecomment-361098617

> What the calendar app does for webcal is the following: try https, if that fails, fall back to http. And that's pretty much what I expect from any other webcal client.

I tested that with Outlook:
My server is running on TLS at port 444 (DNAT because of internal routes), so my scheme is webcals://my.domain:444/remote.php/dav/.....
- If i enter webcal://my.domain:444/remote.php/dav/..... Outlook just does nothing, webserver logs "400 (Bad request)"
- If i enter the webcals:// version Outlook wizard continues.

So in fact what Georg expects from other webcal clients isn't what other developers think... 😯 😞 
Considering the unofficial support of "webcals://" and the Outlook-test it would be appropriate to apply PFS accordingly and use webcals:// when accessing via HTTPS.

### Server configuration

**Web server:** nginx

**Database:** mysql

**PHP version:** 7.2

**Server version:** 13

**Calendar version:** 1.6.0

**Updated from an older installed version or fresh install:** updated

**Are you using external storage, if yes which one:** local

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** no

### Client configuration
**Browser:** newest FF

**CalDAV-clients:** Thunderbird, WebcalSync, CalDAV and infrequently Outlook

georgehrke:
PFS = Perfect forward secrecy?

georgehrke:
- [ ] but that also means that we have to support `webcals` ourselves

Somebodyisnobody:
> https://github.com/nextcloud/calendar/issues/748#issuecomment-369743506

Yeah i meant Perfect forward secrecy.

arunikayadav42:
Can I please get started with working on this issue?

[...]

PR #1535, opened 2019-10-19
Calendar 2.0.0 beta1

georgehrke:


tcitworld:
It would be nice if we also added an introduction paragraph explaining the migration to Vue.

georgehrke:
Missing in Changelog: 
- #748

georgehrke:
> It would be nice if we also added an introduction paragraph explaining the migration to Vue.

Definitely! Do you want to look into that? :)

tcitworld:
> Version 2.0 of the calendar has been fully rewritten with a different technology, switching from the legacy AngularJS framework to [Vue.js](https://vuejs.org/). Nextcloud is using more and more Vue.js throughout apps and server, which makes it easy to use [common components](https://nextcloud-vue-components.netlify.com/) everywhere. This allows faster development and a more coherent experience in all of Nextcloud.

> Even though all features present on the 1.x calendar app versions have been reimplemented, new bugs might have been introduced. Please report them if you find some.

Issue #968, opened 2018-11-28
Use illustrations for events


jancborchardt:
Like Google Calendar is using the illustrations for events like "Go to barber" or "Dentist", we can use proper illustrations in the header of the sidebar.

https://undraw.co is a collection of open source illustrations we can use. :)

You can even change the primary color of the SVG, so we could adjust it to the theming color:
![undraw](https://user-images.githubusercontent.com/925062/49144296-54e8f100-f2fd-11e8-864c-90c637bc7537.png)

cc @georgehrke @skjnldsv

georgehrke:
Example from Google calendar:

![98b27736-8ccc-4d45-a192-238e21aebcc3](https://user-images.githubusercontent.com/1250540/49144575-04be5e80-f2fe-11e8-9d7d-3c5108298008.jpg)

der-Daniel:
Microsoft also has something similar in their web-view of the calendar. 

![image](https://user-images.githubusercontent.com/7686866/51915282-029a7e00-23db-11e9-9a90-356f5fead20f.png)

For special event they have icons rendered next to the event name.

georgehrke:
Relaxation: https://42f2671d685f51e10fc6-b9fcecea3e50b3b59bdc28dead054ebc.ssl.cf5.rackcdn.com/illustrations/relaxation_1_wbr7.svg

Presentation: https://42f2671d685f51e10fc6-b9fcecea3e50b3b59bdc28dead054ebc.ssl.cf5.rackcdn.com/illustrations/Presentation_62e1.svg
https://42f2671d685f51e10fc6-b9fcecea3e50b3b59bdc28dead054ebc.ssl.cf5.rackcdn.com/illustrations/business_plan_5i9d.svg

Camping: https://42f2671d685f51e10fc6-b9fcecea3e50b3b59bdc28dead054ebc.ssl.cf5.rackcdn.com/illustrations/Camping_y4an.svg
https://42f2671d685f51e10fc6-b9fcecea3e50b3b59bdc28dead054ebc.ssl.cf5.rackcdn.com/illustrations/into_the_night_vumi.svg

Movie: Cinema: https://42f2671d685f51e10fc6-b9fcecea3e50b3b59bdc28dead054ebc.ssl.cf5.rackcdn.com/illustrations/movie_night_93wl.svg

[...]

PR #1537, opened 2019-10-19
Add illustration for (romantic) date as per Jan's suggestion

georgehrke:
https://github.com/nextcloud/calendar/issues/968#issuecomment-531820371


tcitworld:
I feel it's a little risky to have `date` in English but other than that it's fine.

georgehrke:
I totally agree, but I wasn't sure what other word to use.

tcitworld:
Thinking about it, I've realized there's no real good French translation of (romantic) *date*, as *rendez-vous* can be a business meeting or appointment and *rencard* is kinda slang. It matters pretty much only on the context of your sentence, and I feel it might be the same for other languages.

Is it worth it to have this if people might have the wrong illustration? Maybe we should restrict to proper and common nouns that have only one meaning?

(To be fair I've never entered any of my romantic dates into my calendar)

georgehrke:
I'm also fine with not adding it. :)

Let's close this pull-request for now.

PR #1538, opened 2019-10-19
Add different issue templates

georgehrke:
for bugs, feature requests and redirect people with questions to forum


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,37 @@
+---
+name: "\U0001F680 Feature Request"
```
tcitworld: Are the quotes necessary for the emoji?

georgehrke: As far as i can tell yes.

PR #1540, opened 2019-10-19
Add support for categories

### LINKED ISSUE #107: Request: Support for CATEGORIES field
(Spinoff from #71)

The ical standard defines a [CATEGORIES](https://tools.ietf.org/html/rfc5545#section-3.8.1.2) field. I would like this to be exposed in the calendar app as well.

A simple text field would do in the short term.

A nicer solution for the future would be a combination of free form field and auto-complete.

Ideally, categories could somehow modify the appearance of a `VEVENT`. Thunderbird does this with a tiny colour-flag, but sky's the limit.

Here's an example of categorized entries in Thunderbird's calendar. (the "Urlaub" entries have a category)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39954040-request-support-for-categories-field?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
fixes #107 

ToDo:
- [x] read only view is missing

6CF201B5-1891-4898-8984-55D3786F2B33

B98F2439-E10D-459F-B4AC-D0A387885A9A


georgehrke:
Requires: https://github.com/georgehrke/calendar-js/pull/38

georgehrke:
I also added an indicator in the calendar-grid view, but I'm not 100% sure about that.

3BDCD0F7-7D77-4EB8-A9D1-68E29FDAE97B

georgehrke:
@tcitworld What do you think about the indicator in the calendar-grid?

src/components/Editor/Properties/PropertySelectMultiple.vue
```diff
@@ -0,0 +1,117 @@
...
+	},
+	computed: {
+		display() {
+			return true
+		},
+		options() {
+			return this.propModel.options
+		},
+		foo() {
+			return uidToColor('abc')
```
georgehrke: Remove me

tcitworld:
Can we ask to move @skjnldsv's uidToColor.js into a common `@nextcloud` package?

georgehrke:
> Can we ask to move @skjnldsv's uidToColor.js into a common @nextcloud package?

Yes. Please. We were wondering earlier whether to put it into a dedicated package or just nextcloud/vue

georgehrke:
(Damn, npm naming, sorry for mentioning everyone)

tcitworld:
> What do you think about the indicator in the calendar-grid?

I've never seen any indicator apart from the thunderbird screenshot in the issue, so I'm not sure how much it would be useful.

Maybe we should get it consistent with https://github.com/nextcloud/server/issues/13330 and https://github.com/nextcloud/server/issues/6778?

skjnldsv:
> Yes. Please. We were wondering earlier whether to put it into a dedicated package or just nextcloud/vue

Currently it's already in @nextcloud/vue :)

jancborchardt:
With so many full-on colors, it probably makes sense to maybe look into using a similar way of coloring as macOS Calendar, with the events by default not being full-saturation (but only the active one):


More: https://support.apple.com/en-euro/guide/calendar/welcome/mac

georgehrke:
@jancborchardt How is that related to categories? :)

jancborchardt:
Sorry, that was in reference to your comment above at https://github.com/nextcloud/calendar/pull/1540#issuecomment-544563314

So basically I meant to say: Let’s not put any indicators in the view for now. We have more stuff to clean up before we put stuff in again. :)

georgehrke:
I removed the indicator in the calendar-grid for now and added a read-only version:
BC208BD7-FBE3-48BF-990F-A45859A261B8

PR #1541, opened 2019-10-20
Repeat summary

georgehrke:
This implements a repeat summary and based on that a read-only view for the repeat tab.

60CE4628-359A-4B68-AA1F-CB4E833367C2

### DIFF
--- a/src/components/Editor/Repeat/RepeatSummary.vue
+++ b/src/components/Editor/Repeat/RepeatSummary.vue
@@ -21,14 +21,270 @@
   -->
 
 <template>
-	<div>
-		<h2>{{ $t('calendar', 'Summary') }}</h2>
-		<span>{{ $t('calendar', 'This event does not repeat.') }}</span>
+	<div class="repeat-option-set repeat-option-set--summary">
+		<span class="repeat-option-set-summary__label">{{ $t('calendar', 'Summary') }}:</span>
+		<span class="repeat-option-set-summary__summary">{{ summary }}</span>
 	</div>
 </template>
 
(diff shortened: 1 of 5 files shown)


src/components/Editor/Repeat/RepeatSummary.vue
```diff
@@ -21,14 +21,270 @@
...
+				byDayNames.push(allByDayNames[5])
+			}
+			if (byDayList.includes('SA')) {
+				byDayNames.push(allByDayNames[6])
+			}
+			if (byDayList.includes('SU')) {
+				byDayNames.push(allByDayNames[0])
+			}
+
+			return byDayNames.join(', ')
```
tcitworld: Support is pretty bad right now but we could use [`Intl.ListFormat.format()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ListFormat) here.

georgehrke: https://caniuse.com/#feat=internationalization

Support seems not that bad anymore

tcitworld: I feel like this caniuse page is generic and concerns things like `Intl.DateTimeFormat` (for which support is good).

georgehrke: Ah, yes.
MDN shows way more red: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ListFormat#Browser_compatibility

Issue #1020, opened 2018-12-20
Notice of 2 invalid options during make


jancborchardt:
During `make` of latest master, the log contains this:
```
Invalid Option: Unexpected option value "false" for rule "no-descending-specificity"
Invalid Option: Unexpected option value "false" for rule "no-duplicate-selectors"
```

We should make sure that make runs through cleanly without any warnings or errors, even if they don’t break anything. :)

georgehrke:
Since #926 was merged, make works as expected again.

jancborchardt:
I only got these 2 test errors, unrelated to above but still issue during make:

```
 FAIL  js/vendor/jstzdetect/utilities/test.js
  ● Test suite failed to run

    Your test suite must contain at least one test.

      at node_modules/@jest/core/build/TestScheduler.js:242:24
      at asyncGeneratorStep (node_modules/@jest/core/build/TestScheduler.js:131:24)
      at _next (node_modules/@jest/core/build/TestScheduler.js:151:9)
      at node_modules/@jest/core/build/TestScheduler.js:156:7
      at node_modules/@jest/core/build/TestScheduler.js:148:12
      at onResult (node_modules/@jest/core/build/TestScheduler.js:271:25)

Summary of all failing tests
 FAIL  js/vendor/jstzdetect/utilities/test.js
  ● Test suite failed to run

    Your test suite must contain at least one test.

      at node_modules/@jest/core/build/TestScheduler.js:242:24
      at asyncGeneratorStep (node_modules/@jest/core/build/TestScheduler.js:131:24)
      at _next (node_modules/@jest/core/build/TestScheduler.js:151:9)
      at node_modules/@jest/core/build/TestScheduler.js:156:7
      at node_modules/@jest/core/build/TestScheduler.js:148:12
      at onResult (node_modules/@jest/core/build/TestScheduler.js:271:25)


Test Suites: 1 failed, 38 passed, 39 total
```

georgehrke:
Ah, `js/vendor` is an old folder from version 1.
Let me just delete the entire `js/` folder on make

PR #1549, opened 2019-10-21
Delete the entire js folder on make clean to get rid of old v1 code

georgehrke:
https://github.com/nextcloud/calendar/issues/1020#issuecomment-544427084


georgehrke:
yes :)

jancborchardt:
Thanks a lot, works now! :)

PR #1551, opened 2019-10-21
Merge calendar and subscription list

georgehrke:
At Nextcloud conference, @tcitworld and me also discussed that we should consider merging calendar and subscription list, because many users don't understand why some shared calendars appear in calendars and others appear in subscriptions.

I was experimenting a little and came up with this for a new list design:

81C681FD-5747-4C1F-A82D-011BBFB168CB

716E506D-D56B-4AA5-8A11-9117EF39CD38

This would also satisfy users who preferred the old behaviour of calendars with VEvent and VTodo.

@tcitworld @jancborchardt What do you think about this proposal?


georgehrke:
78C07BA9-FCBF-46CA-8ACE-17BC83D56559

Or alternatively without bold font, but with icon.

tcitworld:
What I can think of right away:
* Maybe an icon like [this one](https://materialdesignicons.com/icon/calendar-check) for the calendars with tasks?
* How do you distinguish shared calendars which are read-only and write access?
* Find a way to distinguish subscriptions? We have already the « shared by public link » information that is associated to the link icon, so it's not great to have a link icon to express this.

Maybe we can only differentiate write/read-only status ?

georgehrke:
> Maybe an icon like this one for the calendars with tasks?

👍 

> How do you distinguish shared calendars which are read-only and write access?

Does it matter to be able to tell them apart in the calendar-list?
If you click an event that's in one of those calendars or you try to create a new one, you will only see the writable ones.

> Maybe we can only differentiate write/read-only status ?

Not sure i completely get your point, but that's what calendar list and subscription list did all those years?

tcitworld:
>> Maybe we can only differentiate write/read-only status ?

> Not sure i completely get your point, but that's what calendar list and subscription list did all those years?

Yeah, but I felt it made still sense to differentiate them somehow, even in a merged list, in order to explain why they wouldn't be available in the calendar picker.
The use case I'm thinking of is when someone shares you a calendar but you can't understand why you can't create events in it, so you might think of a bug but it's just that the owner of the calendar didn't check "write access".

Therefore I feel it would be nice to show something like a tooltip / title when hovering over the calendar to tell this, or some kind of other visual tip.

jancborchardt:
Sounds good. But the heading "My calendars" is a bit unneeded here (apart from us never using the "my" form in the interface).

We could cut the header and put the "+ Add calendar" (which opens the popover) as a half-grey entry at the bottom of the list. (Just like we have that proposal for Contacts for "+ Add group" as last item in the group list.)

georgehrke:
@tcitworld I checked and no other app does that. Like in files, you can see the Avatar to indicate it is shared with you, but from this screenshot you can't that one is read-only and the other is read-write: 
0A2FC2A2-DA8C-4096-ADC6-9CFF41B852C0

Same for contacts.

@jancborchardt What's your call on this? Is it necessary to be able to tell read-only and read-write shared calendars apart in the calendar-list?

tcitworld:
> I checked and no other app does that. Like in files, you can see the Avatar to indicate it is shared with you, but from this screenshot you can't tell that one is read-only and the other is read-write.

Thinking back, the issue is not at the calendar list level but at the calendar picker level (since you're not using the calendar list to create a new event). So...disregard everything I said, it doesn't apply here. I'm still interested in discussing how we could do something at the calendar picker level.

georgehrke:
> Thinking back, the issue is not at the calendar list level but at the calendar picker level (since you're not using the calendar list to create a new event). So...disregard everything I said about this, it doesn't apply here. I'm still interested in discussing how we could do something at the calendar picker level.

I already added an avatars to the calendar-picker, in case the calendar is shared with you. 
But feel free to open another issue to discuss further changes to it :)

jancborchardt:
Yeah, I’d also say the discussion on adding an indicator or how to differentiate is separate from this here. :)

georgehrke:
5E57249B-9B15-4988-AE34-E79CF12A0165

9A1CA80A-6F46-4B8C-B4CB-4B0ADDDBEF9B

georgehrke:
I still have to reduce the default opacity of the New Calendar entry.

jancborchardt:
Nice stuff! More design feedback:
- [ ] "+ Add calendar" is better wording as all the items in the menu are "Add"
- [ ] The whole "+ Add calendar" list entry should be clickable, not just the icon. As such, the menu should open centered below the list item, with the triangle pointing to the middle, not to the icon. :) Then the popdown can also be wider to fit the whole text of the entries, not needing to break the text.
- [ ] Minor wording: "Add calendar with task list" (not "tasks list", and lowercase "calendar" as it doesn’t refer to the app, same for the "Add calendar" entry above)
- [ ] The entry should be `var(--color-text-maxcontrast)` just like the "Show all folders"/"Collapse" list entries in Mail

Kind of related since we are working on the list:
- [ ] The "Shared" icon with text is closer to the 3-dot icon than the share icon, why that?
- [ ] The "Contact birthdays" calendar, being a default calendar, needs a color that is better visible and fits with the others? The one you use for "Default calendar" in the screenshot would fit.

georgehrke:
> The whole "+ Add calendar" list entry should be clickable, not just the icon. 

That's already the case. https://github.com/nextcloud/calendar/pull/1551/files#diff-d389c50efe9671db3af56f7d289e3f62R26 :) 

>  As such, the menu should open centered below the list item, with the triangle pointing to the middle, not to the icon. :) Then the popdown can also be wider to fit the whole text of the entries, not needing to break the text.

I can make it wider, but i would like to stick to using the vue `Actions` component, because it has all the great accessibility keyboard stuff, which is a lot of work to rebuild. The only downside about Actions is that it currently requires an icon as central element.
Unless @skjnldsv has some good recommendation here :)

src/components/AppNavigation/CalendarList/CalendarListTitle.vue
```diff
@@ -0,0 +1,78 @@
...
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<AppNavigationItem
+		class="app-navigation-entry-new-calendar"
+		:title="$t('calendar', 'New calendar')"
+		@click="toggleMenu">
+		<template slot="icon">
```
skjnldsv: Please do not do that!
This is the reason I wanted to have strict rules ^^'
Actions needs to go at the end of the line, not ont he icon please.

skjnldsv:
> Unless @skjnldsv has some good recommendation here :)

See https://github.com/nextcloud/calendar/pull/1551/files#r338393703

src/components/AppNavigation/CalendarList/CalendarListTitle.vue
skjnldsv: I mean common, we're creating standards, let's not break them not even a week after they're out :sob:

src/components/AppNavigation/CalendarList/CalendarListTitle.vue
```diff
@@ -0,0 +1,78 @@
...
+		AppNavigationItem
+	},
+	data() {
+		return {
+			isOpen: false
+		}
+	},
+	methods: {
+		toggleMenu() {
+			setTimeout(() => {
```
skjnldsv: Why this?

georgehrke:
| Closed menu | open menu |
|:---------:|:------:|
|E2518B23-8734-4510-BC58-DD15DF444EEB|8980FA5A-BCA1-4E98-93C9-595193C2E699|

skjnldsv:
The little `+` in front feels a bit too much :thinking:

georgehrke:
We have the same in front of "+ New event". :)

skjnldsv:
> We have the same in front of "+ New event". :)

then same comment :p
Can you use the icon-add for the event button?
And no `+` for the calendar button?

georgehrke:
It is what @jancborchardt suggested and I like it as well

skjnldsv:
> It is what @jancborchardt suggested and I like it as well

I think he meant the + we use in every other apps, so the icon-add button :)
dev skjnldsv com_apps_contacts_All%20contacts_41e602ab-83c1-46fc-89c2-57dc392439x01_contacts

georgehrke:
@jancborchardt ^ ? :)

src/components/AppNavigation/CalendarList/CalendarListNew.vue
```diff
@@ -22,94 +22,250 @@
...
-		:value="name"
-		:disabled="isCreating"
-		@submit.prevent.stop="addCalendar">
-		{{ $t('calendar', 'Name of calendar') }}
-	</ActionInput>
+		class="app-navigation-entry-new-calendar"
+		:class="{'app-navigation-entry-new-calendar--open': isOpen}"
+		:title="$t('calendar', '+ New calendar')"
+		@click.prevent.stop="openDialog">
+		<template slot="counter">
```
skjnldsv: Please use the actions slot https://nextcloud-vue-components.netlify.com/#/Components/App%20containers?id=appnavigationitem
image

georgehrke: You can't replace the icon when you use actions.  And i would like to have the icon-add, not the three dots

skjnldsv: Then fix it, don't abuse the current components please

georgehrke: https://github.com/nextcloud/nextcloud-vue/pull/686

tcitworld:
Let me find time to play with it a bit.

tcitworld:
Also a click on `+ New calendar` (not on the 3 dots button) opens the menu but doesn't close it.

tcitworld:
And finally I'm also interested in @jancborchardt opinion on using an icon or a `+` in front of `New event` or `New calendar`.

georgehrke:
I fixed both issues you mentioned

PR #1556, opened 2019-10-22
use generateUrl instead of linkTo as router base, in order to accomod…

georgehrke:
…ate url rewrite to hide index.php

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -21,7 +21,7 @@
 
 import Vue from 'vue'
 import Router from 'vue-router'
-import { linkTo } from '@nextcloud/router'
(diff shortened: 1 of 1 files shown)


src/router.js
```diff
@@ -33,7 +33,7 @@ Vue.use(Router)
 
 const router = new Router({
 	mode: 'history',
-	base: linkTo('calendar', 'index.php'),
+	base: generateUrl('/apps/calendar', ''),
```
juliushaertl: The second parameter should be an object or undefined, so you can omit it here: https://github.com/nextcloud/nextcloud-router/blob/master/lib/index.ts#L54

PR #1557, opened 2019-10-22
Update README

georgehrke:
@tcitworld Should we also update our supported browsers?

- Firefox 45 is no longer supported by Mozilla. Latest ESR is 68, 60 ESR is still supported.
- Latest supported Chrome/Chromium is 77

Same for Edge. I think we should only support the Blink based version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,52 +1,58 @@
 # Nextcloud Calendar 
 
 [![Build Status](https://travis-ci.org/nextcloud/calendar.svg?branch=master)](https://travis-ci.org/nextcloud/calendar)
-[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/nextcloud/calendar/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/nextcloud/calendar/?branch=master)
-[![Code Coverage](https://scrutinizer-ci.com/g/nextcloud/calendar/badges/coverage.png?b=master)](https://scrutinizer-ci.com/g/nextcloud/calendar/?branch=master)
 [![Coverage Status](https://coveralls.io/repos/github/nextcloud/calendar/badge.svg?branch=master)](https://coveralls.io/github/nextcloud/calendar?branch=master)
-[![Dependency Status](https://www.versioneye.com/user/projects/57dc165a037c200040cdced9/badge.svg?style=flat-square)](https://www.versioneye.com/user/projects/57dc165a037c200040cdced9)
 
 **A calendar app for [Nextcloud](http://nextcloud.com). Easily sync events from various devices with your Nextcloud and edit them online.**  
 
 ![](https://raw.githubusercontent.com/nextcloud/screenshots/master/apps/Calendar/view_week.png)
 
-## Why is this so awesome?
+## :blue_heart: :tada: Why is this so awesome?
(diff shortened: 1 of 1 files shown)


tcitworld:
Just say the same as https://github.com/nextcloud/browserslist-config-nextcloud says ?
So basically https://browserl.ist/?q=last+2+versions%2C+ie+%3E%3D+11%2C+firefox+ESR%2Cnot+dead

README.md
```diff
@@ -41,12 +40,21 @@ Check out our [FAQ](https://github.com/nextcloud/calendar/wiki/FAQs). If you don
...
+We’re also in [#nextcloud-calendar on freenode IRC](https://webchat.freenode.net/?channels=nextcloud-calendar).
 
-Testing is a great way to contribute without having to write source code.
-Although it's straight forward, setting up the development environment requires some knowledge and extra tools on your device.
+We’d like to thank [BrowserStack](https://www.browserstack.com) for providing us with a free subscription.
+
+## Nightly builds / Pre-releases
 
-We provide [nightly builds](https://nightly.portknox.net/calendar/?C=M;O=D) to enable everyone help testing without setting up the development environment.  
+Instead of setting everything up manually, you can just [download the nightly builds](https://nightly.portknox.net/calendar/?C=M;O=D) or [download a pre-release](https://github.com/nextcloud/calendar/releases) instead.
```
tcitworld: Nightly build link seems to be dead.

georgehrke: Yes, i know. Let me send an email to Steffen

georgehrke:
> Just say the same as https://github.com/nextcloud/browserslist-config-nextcloud says ?
So basically https://browserl.ist/?q=last+2+versions%2C+ie+%3E%3D+11%2C+firefox+ESR%2Cnot+dead

I'm not a big fan of the current config, actually. It also includes Opera Mini and other weird esoteric browsers with next to no market share.

tcitworld:
> Just say the same as https://github.com/nextcloud/browserslist-config-nextcloud says ?

Except maybe IE. 😬

tcitworld:
> I'm not a big fan of the current config, actually. It also includes Opera Mini and other weird esoteric browsers with next to no market share.

So maybe closer to https://jamie.build/last-2-versions ?

georgehrke:
But besides that, pretty much that list yes. Although I would still limit it to Firefox, Chrome, Edge, Safari and maybe opera. 

While our browser list config theoretically supports things like KaiOS browser, we realistically can't test / debug it.

georgehrke:
I updated the browser list accordingly. Should we get this in? :)

README.md
```diff
@@ -1,52 +1,60 @@
...
-* Edge 14+
-* Firefox 45+
+* Chrome/Chromium 76+
+* Edge 40+
+* Firefox 60+
 * Internet Explorer 11
-* Safari 10+
+* Safari 12.1+
 
 We don't support Internet Explorer 10 and below. Patches for IE9+ are accepted though.
```
tcitworld: Maybe just remove the second sentence.

georgehrke: Then I can also remove both sentences, since it should be obvious that we don't support IE 10 if the list says IE11

PR #1561, opened 2019-10-22
Allow to edit location and description in popover editor, if event al…

### LINKED ISSUE #680: When location and/or description is set for an event, already show that info in the small popover
Currently from the small popover it is not visible at all what info is set. Actually that view should be a summary of all info which is set. At the bare minimum of course only showing the basics, but _when it is set_, also showing things like location and/or description (and maybe more).

cc @nextcloud/calendar 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/52640794-when-location-and-or-description-is-set-for-an-event-already-show-that-info-in-the-small-popover?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
…ready has some set

fixes #680 

7878D11E-75C2-447B-B915-31923F302A79

### DIFF
--- a/src/views/EditSimple.vue
+++ b/src/views/EditSimple.vue
@@ -146,6 +146,17 @@ export default {
 			this.$refs.popover
 				.$children[0]
 				.$_restartPopper()
+		},
+		calendarObjectInstance() {
+			this.hasLocation = false
+			this.hasDescription = false
+
+			if (typeof this.calendarObjectInstance.location === 'string' && this.calendarObjectInstance.location.trim() !== '') {
+				this.hasLocation = true
+			}
+			if (typeof this.calendarObjectInstance.description === 'string' && this.calendarObjectInstance.description.trim() !== '') {
+				this.hasDescription = true
+			}
 		}
 	},
 	mounted() {


src/views/EditSimple.vue
```diff
@@ -146,6 +146,17 @@ export default {
 			this.$refs.popover
 				.$children[0]
 				.$_restartPopper()
+		},
+		calendarObjectInstance() {
+			this.hasLocation = false
+			this.hasDescription = false
+
+			if (this.calendarObjectInstance.location !== null) {
```
tcitworld: May those be empty strings?

georgehrke: only if the ICS contains
```
LOCATION:
```

tcitworld: So we might want to exclude these.

georgehrke: sure, let me change that

jancborchardt:
Detail: The location field could use vue-autosize and have `resize: none;` set. :) (The mail app uses that for the signature setting, if you need a reference: https://github.com/nextcloud/mail/search?q=autosize&unscoped_q=autosize )

georgehrke:
It already uses vue-autosize.
Let me add resize: none in a different pull-request. :)

PR #1562, opened 2019-10-22
Update Today in calendar-grid

### LINKED ISSUE #678: update calendar 'today'

georgehrke:
fixes #678

### DIFF
--- a/src/views/Calendar.vue
+++ b/src/views/Calendar.vue
@@ -132,6 +132,8 @@ export default {
 		return {
 			loadingCalendars: true,
 			timeFrameCacheExpiryJob: null,
+			updateTodayJob: null,
+			updateTodayJobPreviousDate: null,
 			showEmptyCalendarScreen: false
 		}
 	},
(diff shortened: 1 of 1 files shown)


src/views/Calendar.vue
```diff
@@ -254,6 +256,22 @@ export default {
...
+				return
+			}
+
+			if (this.updateTodayJobPreviousDate !== newDate) {
+				this.updateTodayJobPreviousDate = newDate
+
+				const calendarApi = this.$refs.fullCalendar.getApi()
+				calendarApi.render()
+			}
+		}, 1000)
```
raimund-schluessler: No idea how much computation this needs, but wouldn't it be enough to do it every minute or so? I guess checking every second is not really necessary, since 86399 times out of 86400 nothing has to be done.

georgehrke: I clicked around and couldn't tell any performance difference.

I would personally prefer the more accurate version of updating it immediately, because that's what people are used to from other calendaring software.

georgehrke: I also tried to reduce write access by only updating `this.updateTodayJobPreviousDate` when it actually changed, not every second. So we don't unnecessarily cause any Vue watchers to run.

raimund-schluessler: Fine with me.

Issue #1021, opened 2018-12-20
Simpler design for top part of navigation


jancborchardt:
The current top part of navigation has a bunch of actions which are rarely needed:

![calendar nav before](https://user-images.githubusercontent.com/925062/50314089-cefc3800-04ad-11e9-9f07-4a3d2cae3e40.png)

This is a new proposal:
- A prominent (primary) "+ New event" button, which opens the event popover, by default a full-day event on _today_. Only this button has bold text.
- The "Today" button is put on the same line as it’s not soo important
- The day/week/month switcher is now just an icon indicating the current state. On click it opens a popover menu with a list of Day, Week, Month with the current one highlighted.

![calendar navigation top new](https://user-images.githubusercontent.com/925062/50314211-45993580-04ae-11e9-9203-bf3ff163ee7e.png)

cc @nextcloud/calendar

tcitworld:
Interesting !

What about putting the "+ New event" on the right of the second line instead, to make this button more accessible, since the cursor will probably on this area of the screen initially?

jancborchardt:
You mean switching the order from: "New event, Today, display toggle" to "Today, display toggle, New event"?

I'd rather first go with the initial proposal, as otherwise the primary action is strangely floating in space.

Another option is to put the "Today" button (probably icon only) and the display toggle on the first row with the date switcher as well. But this seems less ideal, especially as then 2 actions would have to be icons.

georgehrke:
> The day/week/month switcher is now just an icon indicating the current state. On click it opens a popover menu with a list of Day, Week, Month with the current one highlighted.

I'm not sure if putting the views in a popover is the best decision. It's a very common, frequently used action. Many people use the month view to get an overview and then week view to look into some days in detail / to schedule events.

But maybe that can be connected with https://github.com/nextcloud/calendar/issues/796

jancborchardt:
> I'm not sure if putting the views in a popover is the best decision. It's a very common, frequently used action. Many people use the month view to get an overview and then week view to look into some days in detail / to schedule events.

Just for reference, the new Google Calendar also does this.


And as a first step maybe we could simply put "Today" and the "Day / Week / Month" switcher on one row, not adding the "New event" button yet.

georgehrke:
> And as a first step maybe we could simply put "Today" and the "Day / Week / Month" switcher on one row,

We had this before, but it broke if the translated strings are too long: https://github.com/nextcloud/calendar/issues/312

[...]

PR #1563, opened 2019-10-22
Default time to 10am to 11am when changing from allDay to timed

### LINKED ISSUE #532: Feature Request: Tweak Default Start Time
### Steps to reproduce
1. Create an event (it defaults to "All Day")
2. Uncheck the All day Event checkbox
3. The start time for the event is set at 12 AM

It's a minor thing, but I don't have too many events that start at midnight. :) Google Calendar currently defaults to 10 AM which is a nice feature in my opinion. As it is now, if a user unchecks the "All day Event" checkbox and forgets to update the start time, they can get reminders in the middle of the night.

Again, it's a minor issue, but I thought I would bring it up. Keep up the great work.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46968720-feature-request-tweak-default-start-time?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
fixes #532

### DIFF
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -210,6 +210,25 @@ const mutations = {
 	},
 
 	/**
+	 * Changes the time of a timed event to the default values
+	 *
+	 * @param {Object} state The Vuex state
+	 * @param {Object} data The destructuring object
+	 * @param {Object} data.calendarObjectInstance The calendarObjectInstance object
+	 */
+	changeTimeToDefaultForTimedEvents(state, { calendarObjectInstance }) {
+		const startDate = calendarObjectInstance.eventComponent.startDate
+		const endDate = calendarObjectInstance.eventComponent.endDate
+		if (startDate.hour === 0 && startDate.minute === 0 && endDate.hour === 0 && endDate.minute === 0) {
+			startDate.hour = 10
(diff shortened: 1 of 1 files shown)


tcitworld:
Shouldn't we consider current time when the day is today?

georgehrke:
I'm not sure if the behaviour should be different for one particular day.

But my idea for https://github.com/nextcloud/calendar/issues/1021 was that when you click "New event" it will automatically open the new event on today and round up to the next hour.
Does that make sense?

tcitworld:
> I'm not sure if the behaviour should be different for one particular day.

My use case was when you have an all day event on today's date and uncheck the all day checkbox, it's odd to default to an already passed time. It's not that important though, so I'm fine with ignoring it for now.

> But my idea for #1021 was that when you click "New event" it will automatically open the new event on today and round up to the next hour.
Does that make sense?

Yes.

tcitworld:
See #1567

PR #1575, opened 2019-10-22
Put text inside ActionLink component to prevent weird formatting

georgehrke:
Before:
65FE305C-ED31-4907-9B63-147B4406388A

After:
1EBD421C-A88E-4BA7-8363-B5191E621DC9


georgehrke:
nextcloud/nextcloud-vue#667 should obviously still be fixed independently from this, but this mitigates the issue for now :)

Issue #1474, opened 2019-09-11
Optimize bundle size


georgehrke:
- [x] dynamic loading of locales in bundle 
- [ ] #1473 
- [x] replace Babel polyfill with core-js
- [x] Do not import Nextcloud/vue, but individual components 
- [ ] ...

raimund-schluessler:
For Tasks the effect was not huge, but removing comments saved 35 kB https://github.com/nextcloud/tasks/pull/571

georgehrke:
Only #1473 is left, closing this ticket then.

PR #1578, opened 2019-10-23
load FullCalendar locale dynamically

### LINKED ISSUE #1569: Locale is not respected anymore
Since the Vue rewrite merge, it seems the locale of the personal settings is not respected anymore?

I have it set to German, but Sunday is shown as first day of the week. What am I missing? @georgehrke @tcitworld? :)

georgehrke:
fixes #1569 

Summoning our Webpack export @skjnldsv 😉 

This doesn't look right to me:
Master:
```
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  calendar.js (1.99 MiB)
```

This branch:
```
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  calendar.js (2.18 MiB)
```

I would have expect it to get much smaller when loading the locales dynamically 🤔

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -53,7 +53,7 @@ module.exports = {
 	plugins: [
 		new VueLoaderPlugin(),
 		new StyleLintPlugin(),
-		new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
+		new webpack.IgnorePlugin(/^\.\/locale(s)?$/, /(@fullcalendar\/core)$|(moment)$/),
 		new webpack.DefinePlugin({
 			appVersion: JSON.stringify(require('./package.json').version)
 		})
--- a/src/views/Calendar.vue
+++ b/src/views/Calendar.vue
@@ -53,6 +53,7 @@
 				:default-date="defaultDate"
 				:locales="locales"
 				:locale="locale"
+				:first-day="firstDay"
 				:selectable="isSelectable"
 				:select-mirror="true"
 				:lazy-fetching="false"
@@ -82,7 +83,6 @@ import listPlugin from '@fullcalendar/list'
 import '@fullcalendar/list/main.css'
 import timeGridPlugin from '@fullcalendar/timegrid'
 import '@fullcalendar/timegrid/main.css'
-import allLocales from '@fullcalendar/core/locales-all'
 import {
 	AppNavigation,
 	AppContent,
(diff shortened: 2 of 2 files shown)


georgehrke:
(Especially since we just loaded **all** locales before)

skjnldsv:
> Summoning our Webpack export @skjnldsv

ahahaha! That lapsus!
```js
import 'skjnldsv'
```

webpack.common.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
 	plugins: [
 		new VueLoaderPlugin(),
 		new StyleLintPlugin(),
-		new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
+		new webpack.IgnorePlugin(/^\.\/locale$/, /^\.\/locales$/, /moment$/),
```
skjnldsv: I'm guessing this is not a proper regex.
Just this should be enough to reduce the total size :)

Also, please use a babel.config.js file instead of having the babel confs in the webpack file :)

georgehrke: I tried changing it to:
```
		new webpack.IgnorePlugin(/^\.\/locale(s)?$/, /(@fullcalendar)$|(moment)$/),
```

but that only brings it down to 1.98, not more :/

skjnldsv: The webpack output should display the other chunks if it is working :)
https://travis-ci.com/nextcloud/contacts/jobs/248200403#L2009

georgehrke:
Ok, so checking with the [webpack-bundle-analyzer](www.npmjs.com/package/webpack-bundle-analyzer) (awesome recommendation by @skjnldsv) I can tell that fullcalendar locales are not inside the main bundle, but only there as chunks. They also appear in the network log as additional request.

So this is ready to be reviewed :)

skjnldsv:
> So this is ready to be reviewed :)

what is the final size?

georgehrke:
nextcloud/vue :P 

0AE8A57D-D688-42C7-9392-10399DBDE098

georgehrke:
Okay, let me try that. But that's something for different PR :)

skjnldsv:
ALso, babel-polyfill is deprecated https://babeljs.io/docs/en/babel-preset-env#usebuiltins ;)

georgehrke:
https://github.com/nextcloud/calendar/issues/1474 :)

PR #1579, opened 2019-10-23
Implement new Design for upper AppNavigation part

### LINKED ISSUE #1021: Simpler design for top part of navigation

georgehrke:
fixes #1021 

When you click "+ New event", it automatically opens the editor with a timed event set for today at the next full hour with a default length of one hour.

029982EB-2022-4B2B-A577-B09CBCF77991

DA438D65-304B-4DA3-8B44-5D2F597A55CD

92950D63-BAF6-4C33-9FC0-80D3906D9BB6

60FEE4AC-308B-4075-A8DB-FD089001748E


jancborchardt:
Looks so much better, good stuff @georgehrke! :) Some notes:
- @tcitworld regarding your note on "Today" yeah, that’s a tough one. We could use the "Calendar icon which has the current date in it" but is that so understandable? Maybe worth to try it out.
- Since we have icons for all of "Day / Week / Month", maybe it makes more sense to use the currently active view as indicator of the switcher instead of the "Grid view" toggle icon. @skjnldsv noted the confusion of using the same icon in the past, and this could fix it?

jancborchardt:
Another option is of course to put the "Today" entry out of the main view and inside the "Day / Week / Month" toggle as first entry. Then we can have the full text, but with the drawback that it’s not soo discoverable.

georgehrke:
Let me replace Today with an icon and post a screenshot.
That will make it easier to discuss it :)

georgehrke:
With button styles:
4F735354-336D-4129-BD5D-BFDFE0B32E81

Without:
B76BBCA4-5256-4D81-9E40-20AB1F0E7705

skjnldsv:
> * @skjnldsv noted the confusion of using the same icon in the past, and this could fix it?

I'm afraid it looks even more confusing on the second screenshot :)
But I agree with the today icon :thinking: 

How does others calendar manage the view toggle icon?

skjnldsv:
So, what about a dropdown with the current selected mode icon? :)
(and a triangle of course)

jancborchardt:
I’d say for now what looks best of all of these is the current state of the pull request and @georgehrke’s screenshots above. It’s a big improvement on the current state.

So how about we merge it like this and look at the individual issues separately? :)

jancborchardt:
> So, what about a dropdown with the current selected mode icon? :)

This is exactly what I suggested above ;)

georgehrke:
D2E0E305-9039-463A-AC3F-826AB04817F5
0E53285D-53BA-49C2-9C60-9510A93FC1A2
1C55FE34-DE79-4158-BCE2-1D37AD3AD969

georgehrke:
And since we use Flexbox now, the button automatically gets bigger:
A1BB7AC6-92B1-4B0B-A911-F02AC1BFC6C5

georgehrke:
Not sure if that's the correct french translation, but then the button will have two lines:
F86DB4F5-56D7-41D5-873D-C0B1E5959126

skjnldsv:
> Not sure if that's the correct french translation, but then the button will have two lines:

I'm guessing ellipsis would be good here.
It will take care of abbreviating long text. `Nouvel Evenem.`

georgehrke:
C7DD62A4-9E8E-4A89-ABF3-1CB47495D23A

Let me push that

skjnldsv:
Though it would be glad if we could ellipsis with one dot instead of 3! :p

georgehrke:
Can you do that with CSS? I'm only aware of `text-overflow: ellipsis`, which does ...

skjnldsv:
> Can you do that with CSS? I'm only aware of `text-overflow: ellipsis`, which does ...

I don't think so, not without hacks :(
https://stackoverflow.com/a/37594113

georgehrke:
> https://stackoverflow.com/a/37594113

But this one shows ... even when nothing was truncated

skjnldsv:
> But this one shows ... even when nothing was truncated

Yep, not really great :)
Let's move on! This is already an awesome improvement!

tcitworld:
What about
```css
text-overflow: '.';
```

Also since the button's text translation is `+ New event` translators may choose to simple translate it into `+ Event` (and there's a smaller risk of using such a translation elsewhere).

skjnldsv:
> ```css
> text-overflow: '.';
> ```

https://caniuse.com/#feat=mdn-css_properties_text-overflow_string

tcitworld:
> https://caniuse.com/#feat=mdn-css_properties_text-overflow_string

Ouch, I thought I used this previously.

georgehrke:
It's all so easy and straight forward with nextcloud/vue.  😍
The popover alone would have been a pain to implement in the old calendar.

skjnldsv:
Glad you're enjoying it as well ! :rocket:

jancborchardt:
> > ```css
> > text-overflow: '.';
> > ```
> 
> https://caniuse.com/#feat=mdn-css_properties_text-overflow_string

Let’s maybe not go with custom ellipsis though, no? One dot is not really obvious, and a bit confusing if it cuts it off mid-sentence with just a dot at the end of a sente.

PR #1600, opened 2019-10-26
Remove deprecated color utils, replace with uidToColor, add tests

georgehrke:
Replaced `getRandomColor` and `getDefaultColor` with `uidToColor`.

Wrote unit tests for most modules inside `utils/`

### DIFF
--- a/src/views/Calendar.vue
+++ b/src/views/Calendar.vue
@@ -90,7 +90,7 @@ import {
 	Content
 } from '@nextcloud/vue'
 import debounce from 'debounce'
-import { getRandomColor } from '../utils/color.js'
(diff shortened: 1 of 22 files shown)


src/views/Calendar.vue
```diff
@@ -321,7 +321,7 @@ export default {
 						this.loadingCalendars = true
 						this.$store.dispatch('appendCalendar', {
 							displayName: this.$t('calendars', 'Personal'),
-							color: getRandomColor(), // TODO - use uid2color
+							color: uidToHexColor(this.$t('calendars', 'Personal')),
```
tcitworld: I'm not sure it's worth it.

PR #1614, opened 2019-10-29
Allow to set custom color per event

### LINKED ISSUE #71: Custom color per event [$75]
Hi,
is it possible to add an Option, to change a meetings color in the calendar? 
(it is only defined by calendar-source colour as at the moment) ...sometimes there may be an important meeting and i want to make it stand out by red color for example...(without changing calendar). i think there should some color/Highlight function? any suggestions? maybe tags+a searchbox (filter) could also be great... #ux

<bountysource-plugin>

---
There is a **[$60 open bounty](https://www.bountysource.com/issues/39954046-custom-color-per-event?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
fixes #71 

| 1. | 2. |
|:---------:|:------:|
|6ED5BBDD-4868-4BAA-80FC-F2F9457F994C|15AD19CF-D65D-48C3-87B0-50025FD35A62|

| 3. | 4. |
|:---------:|:------:|
|EADC6DE0-E763-4EBC-ADF6-CE26578502C5|9EFD2ADC-335E-44C5-903F-5E1D268B7B4D|

1. By default, it's just showing the calendar color (and no x button)
2. Color picker
3. Once the user selected a custom color, they can press the x button to go back to the calendar color
4. Read-only view


simonspa:
Dear @georgehrke, do you think this will make it into the next release?

georgehrke:
@tcitworld Should we just hide the color in read-only mode?

tcitworld:
Not sure the themed illustration pic would be enough, so I think we should keep it for now.

Btw, the "x button" to reset the color is hard to find at the end of the line (and low contrast).

georgehrke:
If i remove the double opacity from `__info` and just leave the opacity from the `ActionButton` itself, it looks like this: 
CAC6AFF9-E88F-476F-BB4F-DBEC3B67988B

Putting the X right next to the color looks a bit odd imho:
3383E286-7949-4A57-BFE7-B8FDF7A85DAF

tcitworld:
I think it makes more sense if it's [a "reset" icon](https://thenounproject.com/term/reset/3039083) next to it.

georgehrke:
We could use `icon-history` for that:
3CDA8690-A2A5-4E0B-8039-B2717E959F01

raimund-schluessler:
How would it look with the color input being the full width? Then everything would be aligned.

georgehrke:
> How would it look with the color input being the full width? Then everything would be aligned.

I tried that, but that looks very weird.
It's way to big and dominant for a simple color-picker:
0A62279E-FBAA-466C-9346-6610CEAEC4B0

tcitworld:
We could display the CSS color name aside the input since we have it (joking, don't do that).

What we have in https://github.com/nextcloud/calendar/pull/1614#issuecomment-593332901 seems fine to me.

georgehrke:
I changed it to https://github.com/nextcloud/calendar/pull/1614#issuecomment-593332901

georgehrke:
@tcitworld Good to go? :)

MartinConsultingServicesInc:
When I change the color, it looks good on the day and week calendar views, but appears merely as a small colored dot on the month view.  Being slightly color blind, the small dot is easy to miss. Can we change the background color on the month view, like we do on the day and week views?

ChristophWurst:
@MartinConsultingServicesInc use the forums please https://help.nextcloud.com/c/apps/calendar/42

Issue #144, opened 2016-10-20
Calendar-inbox: Accept / Reject and show responses in calendar


steckerhalter:
So currently when someone creates an event and adds attendees it sends an email to those people and one has to use a client applications to be able to accept said event.

Shouldn't it also be possible to accept events via the Nextcloud web interface as well? Otherwise the calendar system is not really usable without an external client.

Thinking a bit further, if it would be possible to accept events in the web interface, the next step would be some kind of free/busy implementation.

Is there already some plan regarding these things?

steckerhalter:
I reported this here first because I thought it would require changes to the server: https://github.com/nextcloud/server/issues/1811

georgehrke:
> Is there already some plan regarding these things?

Yes, there is.
The search function is your friend ;)
https://github.com/nextcloud/calendar/issues?utf8=✓&q=is%3Aissue%20is%3Aopen%20free%20busy

https://github.com/nextcloud/calendar/issues/39

georgehrke:
> Shouldn't it also be possible to accept events via the Nextcloud web interface as well? Otherwise the calendar system is not really usable without an external client.

@Gomez Is that already possible via the mail app?

tcitworld:
I can do this last feature once https://github.com/nextcloud/calendar/pull/97 is done ;-)

Gomez:
You mean ics import? Yes possible:
https://github.com/owncloudarchive/mail/pull/1473

georgehrke:
@tcitworld I'm not sure how this is related to #97 
AFAIK we need to use the special INBOX calendar.

mikeziri:
When I create an event and add attendees, does nextcloud send out email to them if email is configured?

georgehrke:
@mikeziri Yes

mikeziri:
@georgehrke awesome. tnx. I couldn't configure my email smtp yesterday to confirm it. it sure does. the only thing I still wanted to know, already mentioned above, is if the attendees that receive the email, get a way to rsvp via api link (yes, no, maybe).

keep up the good work

georgehrke:
@mikeziri Yes, that's planned. See https://github.com/nextcloud/server/issues/2338

tcitworld:
Related to https://github.com/nextcloud/server/issues/19568

Mer0me:
Any update on this issue ? When will it be possible to accept/decline an invitation from de Web UI ? Can we help in some way ?

Mer0me:
[...]

PR #1615, opened 2019-10-29
Add scheduling inbox item to app navigation

georgehrke:


ChristophWurst:
I'm trying to figure out what this was about. @tcitworld do you know? I suppose this has to do with https://github.com/nextcloud/calendar/issues/39 but for what purpose?

tcitworld:
From the branch name it seems to be https://github.com/nextcloud/calendar/issues/144, but I don't see how this would work without server changes.

Issue #1576, opened 2019-10-22
Design polishing of right sidebar


jancborchardt:
Some first review after the Vue merge. :) As usual it’s very nice, these are mostly polishing details!
(This is a review similar to Design polishing of new event popover #1570)

## 📅 Top part

- [x] Event title input placeholder should be "Event title"
- [x] ~~Event title seems rather small. In the popover it’s 20 px, here only 16.~~ (Nextcloud-vue issue)
- [ ] Event title input doesn’t go full width EDIT: Apparently this is to leave space for the menu, which only appears later. This is confusing. Either we show the menu with only "Delete" in it right away, or we show the event title input full width for new events. :)
- [x] ~~The menu button by default has a grey circle, which is too much emphasis. This should only happen on hover or focus as per standard.~~ (Nextcloud-vue issue)
- [x] ~~The "x days ago" subline is _not_ a heading so should not be h4 but simply a `p`~~(Nextcloud-vue issue)
- [x] For a non-all-day event, there seems to be no way to edit the time without first picking the date from the datepicker again. This is very off, and when you click/tap on the time in the input it should for sure open the timepicker instead. The input is wide enough, and it totally worked great before.
- [x] The timepicker itself is also quite strange. Difficult question how to fix that one, but at least it should start at 12:00 instead of 00:00, because 00:00 is one of the _least_ likely times for an event to start. :)
- [x] Same here as in the popover, unchecking the "All day" box is not instant, but there is a tiny delay which feels off.

## ℹ️ Details tab
- [x] Same as for the quick popover: If you only have 1 calendar, it does not need to be displayed.
- [x] Since the location and description fields use autosize, the resize handle can be reomved with `resize: none;`
- [x] The description field should by default be a bit higher which makes it more obvious as description field. 2 lines at least
- [x] ~~Free / Busy only has 2 choices so it doesn’t need to be a dropdown. It could simply be 2 radio choices next to each other.~~
- [x] ~~Similar with "Confirmed/Tentative/Cancelled" which could be a radio choice to have all visible at the same time. Isn’t this also what Google Calendar does simply with "Attending: Yes / No / Maybe"?~~
- [x] ~~And same thing for the "When shared: Show full event / Show only busy / Hide event".~~

## 👥 Attendees tab
- [x] Wording of the emptycontent is too long: "To send out invitations and handle responses, an email address is necessary."
- [x] There should be a button in this emptycontent view to directly "Add email address" which goes straight to the personal settings.
[...]

PR #1632, opened 2019-11-05
Do not show recurrence summary when repeat is set to NONE

georgehrke:
Related to #1576

### DIFF
--- a/src/components/Editor/Repeat/RepeatSummary.vue
+++ b/src/components/Editor/Repeat/RepeatSummary.vue
@@ -21,7 +21,7 @@
   -->
 
 <template>
-	<div class="repeat-option-set repeat-option-set--summary">
+	<div v-if="display" class="repeat-option-set repeat-option-set--summary">
 		<span class="repeat-option-set-summary__label">{{ $t('calendar', 'Summary') }}:</span>
 		<span class="repeat-option-set-summary__summary">{{ summary }}</span>
 	</div>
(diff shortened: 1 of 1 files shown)


src/components/Editor/Repeat/RepeatSummary.vue
```diff
@@ -92,6 +92,15 @@ export default {
...
+		 * We do not want to show it if it doesn't repeat
+		 *
+		 * @returns {boolean}
+		 */
+		display() {
+			return this.frequency !== 'NONE'
+		},
 		summary() {
 			if (this.frequency === 'NONE') {
 				return this.$t('calendar', 'Does not repeat')
```
tcitworld: So…this isn't shown anymore then? Might as well remove it.

georgehrke: Indeed. Let me remove that

georgehrke: Actually no, we need this.

The RepeatSummaryReadOnly component extends RepeatSummary: https://github.com/nextcloud/calendar/blob/master/src/components/Editor/Repeat/RepeatSummaryReadOnly.vue#L37

And this is also shown when the repeat is non-recurring.

tcitworld: Please add a comment so this isn't forgotten, then. :+1:

PR #1635, opened 2019-11-05
Update timezone-database to 2019c

georgehrke:
Related to https://github.com/nextcloud/calendar/issues/1473,
Not automised yet, we should still update it.


georgehrke:
Damn, this actually broke stuff, because there no longer is a timezoneId in the VTIMEZONE block.
PR incoming ...

PR #1639, opened 2019-11-06
fix all translation reporting

georgehrke:


src/components/Shared/DatePicker.vue
```diff
@@ -113,7 +113,7 @@ export default {
 				days: getDayNamesShort(),
 				months: getMonthNamesShort(),
 				placeholder: {
-					date: t('calendar', 'Select Date'),
+					date: this.$t('calendar', 'Select Date'),
```
tcitworld: Why isn't `t` / `translate` imported here?

georgehrke: because we are inside a Vue component and in main.js, we set `Vue.prototype.$t = translate`

PR #1657, opened 2019-11-13
Unit tests

georgehrke:
This pull-request is:
- adding unit tests for `defaults/`, `filters/`, `fullcalendar/`
- making sure that wherever we use moment to set the locale properly
- loads the locale once globally instead of all over again in each component
- hardening against errors in calendar-js / cdav-js while rendering events 


tcitworld:
I'm half-way through reviewing this, will finish tonight.

PR #1658, opened 2019-11-13
l10n: Triple dot ellipsis

rakekniven:
Reported at Transifex.

Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5052,7 +5052,7 @@
       "version": "github:nextcloud/cdav-library#1cefd176749fc7a75e47b98b50bd0140393f147a",
       "from": "github:nextcloud/cdav-library",
       "requires": {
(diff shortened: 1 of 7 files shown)


georgehrke:
@rakekniven I fixed some more :)

tcitworld:
Also make sure we have a [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space) before the ellipsis so that it doesn't go on another line in some cases (we shouldn't have those, but still).

tcitworld:
`eslint` is not happy about the non-breaking spaces.

georgehrke:
I opened an issue in our eslint-config: https://github.com/nextcloud/eslint-config-nextcloud/issues/9

PR #1697, opened 2019-12-02
Allow to send publicly shared link via email

### LINKED ISSUE #1640: Sharing published link via email doesn't work
*cough* *cough*
https://github.com/nextcloud/calendar/blob/master/src/components/AppNavigation/CalendarList/CalendarListItemSharingPublishItem.vue#L213L228

georgehrke:
fixes #1640

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListItemSharingPublishItem.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListItemSharingPublishItem.vue
@@ -61,7 +61,9 @@
 			<ActionInput
 				v-if="showEMailInput"
 				icon="icon-mail"
-				@submit.prevent.stop="sendLinkViaEMail" />
+				@submit.prevent.stop="sendLinkViaEMail">
+				{{ $t('calendar', 'Enter one address') }}
+			</ActionInput>
 			<ActionText
 				v-if="showEMailSending"
 				icon="icon-loading-small">
@@ -144,6 +146,7 @@ import {
 	generateRemoteUrl,
 	linkTo,
 } from '@nextcloud/router'
+import HttpClient from '@nextcloud/axios'
 
 export default {
 	name: 'CalendarListItemSharingPublishItem',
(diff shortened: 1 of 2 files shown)


src/components/AppNavigation/CalendarList/CalendarListItemSharingPublishItem.vue
```diff
@@ -61,7 +61,9 @@
 			<ActionInput
 				v-if="showEMailInput"
 				icon="icon-mail"
-				@submit.prevent.stop="sendLinkViaEMail" />
+				@submit.prevent.stop="sendLinkViaEMail">
+				{{ $t('calendar', 'Enter one address') }}
```
georgehrke: I tried something more descriptive first, but the place is very limited.

Later on it would be nice to allow users to enter a comma separated list of email addresses (like up to 25 at a time)

src/components/AppNavigation/CalendarList/CalendarListItemSharingPublishItem.vue
```diff
@@ -213,21 +216,30 @@ export default {
...
-			//
-			// // HttpClient.post
-			// event.stopPropagation()
-			//
-			// this.sendingEMailLink = true
-			// this.showEMailLinkInput = false
-			//
-			// const value = event.target.querySelector('input[type=text]').value
-			// console.debug(value)
+			const emailAddress = event.target.querySelector('input[type=text]').value
```
tcitworld: I don't see right away why we need to use this ?

src/components/AppNavigation/CalendarList/CalendarListItemSharingPublishItem.vue
tcitworld: In that case we may use some kind of popup so we can also add an eventual message?

src/components/AppNavigation/CalendarList/CalendarListItemSharingPublishItem.vue
georgehrke: Because ActionInput emits the actual event, not the value of the input: 
https://nextcloud-vue-components.netlify.com/#/Components/Actions?id=actioninput

tcitworld: Kinda dirty. :/

PR #1701, opened 2019-12-05
Replace New Reminder button with Multiselect to allow easier selection of alarm

georgehrke:
Related to #1576 

It was not really discoverable, that clicking the "New reminder" button multiple times, gives you different alarms.

And if you want an alarm 15 mins before, clicking 4 times and deleting three reminders (at time, 5 mins before, 10 mins before) is not really intuitive either.

B38077DF-C45B-4AF3-9E83-0B3BC94D039F

EB327709-79CD-44FB-9541-E9938AA5961D

### DIFF
--- a/src/components/Editor/Alarm/AlarmListNew.vue
+++ b/src/components/Editor/Alarm/AlarmListNew.vue
@@ -21,22 +21,85 @@
   -->
 
 <template>
-	<button
-		class="editor-reminders-list-new-button icon-add"
-		@click="addReminder">
-		{{ $t('calendar', 'Add reminder') }}
-	</button>
+	<Multiselect
+		class="property-alarm-new"
+		track-by="value"
+		label="label"
+		:placeholder="$t('calendar', '+ Add reminder')"
+		:options="options"
+		:searchable="false"
+		@select="addReminderFromSelect" />
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


src/components/Editor/Alarm/AlarmListNew.vue
```diff
@@ -21,22 +21,86 @@
...
+		 *
+		 * @param {Number} time Total amount of seconds for the trigger
+		 * @returns {Object} The alarm object
+		 */
+		getAlarmObjectFromTriggerTime(time) {
+			const timedData = getAmountAndUnitForTimedEvents(time)
+			const allDayData = getAmountHoursMinutesAndUnitForAllDayEvents(time)
+
+			return {
+				type: 'AUDIO',
```
tcitworld: Not `DISPLAY` ?

tcitworld: Not `DISPLAY`?

tcitworld: (even though it's the same thing for NC server)

georgehrke: This object is only needed, so we can pass it to the alarmFilter.
The actual creation will just use the total amount of seconds, not this object.

Let me actually remove the type, since alarmFilter does not need it.

PR #1704, opened 2019-12-05
Remove comments from final minified build

georgehrke:
Similar to https://github.com/nextcloud/calendar/compare/bugfix/noid/remove_comments_production?expand=1

Brings bundle size down by a couple of hundred kB


georgehrke:
I think we should keep that, because it's extracting the license headers by default.
And since the bundled js file does not only contain our own agpl-licensed stuff, but also code from other developers with other licenses, we should keep that copyright note.

It's not fetched by the browser anyway, it's only making the tar.gz. some kb bigger.

PR #1713, opened 2019-12-09
Increment version to 2.0.0 beta3

georgehrke:


georgehrke:
@tcitworld I would like to release beta3 and just move all remaining tickets to beta4.

PR #1714, opened 2019-12-09
Release/2.0.0 beta3 appstore

georgehrke:
Special build of 2.0.0 beta 3 to be released on the appstore.
Only difference is that min-version is 18.

(This allows people using Nextcloud 18 beta to easily test the calendar, without publishing it as stable for the 17 branch yet)


georgehrke:
Tests run fine, except for the ones running on Nc 17, which was to be expected.

georgehrke:
Released in https://github.com/nextcloud/calendar/releases/tag/v2.0.0-beta3-nc18build

PR #1732, opened 2019-12-17
Allow to create talk rooms from calendar

georgehrke:

### DIFF
--- a/lib/Controller/ViewController.php
+++ b/lib/Controller/ViewController.php
@@ -22,11 +22,13 @@ declare(strict_types=1);
  */
 namespace OCA\Calendar\Controller;
 
+use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\IConfig;
 use OCP\IRequest;
 use OCP\IURLGenerator;
+use OCP\Notification\IApp;
 
 /**
  * Class ViewController
@@ -46,18 +48,26 @@ class ViewController extends Controller {
 	private $userId;
 
 	/**
+	 * @var IAppManager
+	 */
+	private $appManager;
+
+	/**
 	 * @param string $appName
 	 * @param IRequest $request an instance of the request
+	 * @param IAppManager $appManager
 	 * @param IConfig $config
 	 * @param string $userId
 	 */
 	public function __construct(string $appName,
 								IRequest $request,
 								IConfig $config,
+								IAppManager $appManager,
 								string $userId) {
 		parent::__construct($appName, $request);
 		$this->config = $config;
 		$this->userId = $userId;
+		$this->appManager = $appManager;
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


tcitworld:
We should add the URI to the [`CONFERENCE`](https://tools.ietf.org/html/rfc7986#section-5.11) property as well.

georgehrke:
I tried that, there is a problem with the encoding of FEATURE param in ical.js

Have to fix / report that upstream. :)

src/components/Editor/Invitees/InviteesList.vue
```diff
@@ -161,6 +191,30 @@ export default {
...
 			})
 		},
+		async createTalkRoom() {
+			try {
+				this.creatingTalkRoom = true
+				const url = await createTalkRoom(this.calendarObjectInstance.title)
+
+				let newDescription
+				if (!this.calendarObjectInstance.description) {
+					newDescription = url
```
tcitworld: Add a new line at the end so it's easier to edit afterwards.

src/components/Editor/Invitees/InviteesList.vue
```diff
@@ -161,6 +191,30 @@ export default {
...
+				} else {
+					newDescription = this.calendarObjectInstance.description + '\r\n\r\n' + url
```
tcitworld: Probably here as well.

lib/Controller/ViewController.php
```diff
@@ -76,6 +86,7 @@ public function index():TemplateResponse {
 			'show_weekends' => $this->config->getUserValue($this->userId, $this->appName, 'showWeekends', 'yes') === 'yes',
 			'show_week_numbers' => $this->config->getUserValue($this->userId, $this->appName, 'showWeekNr', 'no') === 'yes',
 			'skip_popover' => $this->config->getUserValue($this->userId, $this->appName, 'skipPopover', 'no') === 'yes',
+			'talk_enabled' => $this->appManager->isEnabledForUser('spreed'),
```
rullzer: It really is time to itnroduce you to the initialstate api ;) @ChristophWurst we must spread the word of it!

georgehrke: Yes, it's on my never ending todo list ;)

src/components/Editor/Invitees/InviteesList.vue
georgehrke: Resolved

src/components/Editor/Invitees/InviteesList.vue
georgehrke: Resolved

georgehrke:
`CONFERENCE` property and initial state thing is something i would like to address later :)

src/services/talkService.js
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+import HTTPClient from '@nextcloud/axios'
+import { translate as t } from '@nextcloud/l10n'
+import { generateUrl, generateOcsUrl } from '@nextcloud/router'
+
+export function isTalkEnabled() {
```
tcitworld: Remove me

Issue #1197, opened 2019-03-26
Add a visual clue for events with/without alarms


MarcosBL:
### Steps to reproduce
1. Open calendar
2. No visual clue on which events have/doesn't have alarms set

### Expected behaviour
I would like to have a visual clue on which events already have an alarm set, along the lines of this (AgenDAV):

<img src="https://i.imgur.com/UVjx3IT.png"/>

### Actual behaviour
No visual clue

### Sample

For example, in **app.js**, at the end of 

````
eventRender: function eventRender(event, element) {
````

We could add something like:

````
if(event.getSimpleEvent().alarm.length) {
    element.css({ 'background-image': 'url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAA4AAAAOCAYAAAAfSC3RAAAABHNCSVQICAgIfAhkiAAAAAlwSFlzAAAAaAAAAGgBo+Z6PAAAABl0RVh0U29mdHdhcmUAd3d3Lmlua3NjYXBlLm9yZ5vuPBoAAAC6SURBVCiRrdKhTgNBEMbx35HUIUguIIrlDQ5HVQmkogjehcfgXbCowwCtQVNRVVJTdaL+MHNwIXtkSfiSSTbzzX9msrukVWCJRZyzVOAGbcQ8Bz6JKS22ES1ecfzbpBdscNHLT/CB56HJ19F9mvAuw7vqEgc9s0KDpwRYh3eeAkvsovNPteGVKfBP+hdwlFH/VdNd7wwP2ON+ALrDIW7x2IFvOMrcskE19JXOfL9njXVmU6dY4R3jVMEnDh8iN6jojq8AAAAASUVORK5CYII=")', 'background-repeat': 'no-repeat', 'background-position': 'right', 'padding-right': '16px' });
}
````

or even better, apply a proper class.

That would give us something like this in Nextcloud:

<img src="https://i.imgur.com/kWFOwgY.png"/>

MarcosBL:
May I try a PR with this simple approach ?

georgehrke:
#1755 was merged, so this will be included in the upcoming 2.0 version.

PR #1755, opened 2020-01-01
Add a reminder icon for events with notifications

justin-sleep:
Implements #1197:

As I'm new to the codebase (and Vue in general), I'm *very* open to criticism! My method for detecting light vs. dark calendar colors felt particularly hacked together; if a better, more general solution exists, I'd prefer to use it.

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -171,3 +171,24 @@
 .fc * {
 	box-sizing: content-box !important;
 }
+
+// Reminder icon on events with alarms set
+.fc-event-nc-alarms {
+	.icon-event-reminder {
+		background-color: inherit;
+		background-position: right;
+		position: absolute;
+		top: 0;
+		right: 0;
+		&--light {
+			@include icon-color('reminder', 'calendar', '#fffffe', 1);
+		}
+		&--dark {
+			@include icon-color('reminder', 'calendar', '#000001', 1);
+		}
+	}
+}
+
+.fc-content {
+	background-color: inherit;
+}
--- a/src/fullcalendar/eventRender.js
+++ b/src/fullcalendar/eventRender.js
@@ -29,6 +29,16 @@
  * @param {Node} el The HTML element
  */
 export default function({ event, el }) {
+	if (el.classList.contains('fc-event-nc-alarms')) {
+		const notificationIcon = document.createElement('span')
+		notificationIcon.classList.add('icon-event-reminder')
+		if (event.extendedProps.darkText) {
+			notificationIcon.classList.add('icon-event-reminder--dark')
+		} else {
+			notificationIcon.classList.add('icon-event-reminder--light')
+		}
+		el.firstChild.appendChild(notificationIcon)
+	}
 	el.dataset.objectId = event.extendedProps.objectId
 	el.dataset.recurrenceId = event.extendedProps.recurrenceId
 }
--- a/src/fullcalendar/eventSourceFunction.js
+++ b/src/fullcalendar/eventSourceFunction.js
@@ -20,6 +20,10 @@
  *
  */
 import { translate as t } from '@nextcloud/l10n'
+import {
+	hexToRGB,
+	isLight,
+} from '../utils/color.js'
 import logger from '../utils/logger.js'
 
 /**
@@ -52,6 +56,10 @@ export function eventSourceFunction(calendarObjects, calendar, start, end, timez
 				classNames.push('fc-event-nc-tentative')
 			}
 
+			if (object.hasComponent('VALARM')) {
+				classNames.push('fc-event-nc-alarms')
+			}
+
 			const fcEvent = {
 				id: [calendarObject.id, object.id].join('###'),
 				title: object.title || t('calendar', 'Untitled event'),
@@ -66,6 +74,7 @@ export function eventSourceFunction(calendarObjects, calendar, start, end, timez
 					calendarOrder: calendar.order,
 					calendarName: calendar.displayName,
 					calendarId: calendar.id,
+					darkText: isLight(hexToRGB(calendar.color)),
 				},
 			}
 
--- a/tests/javascript/unit/utils/color.test.js
+++ b/tests/javascript/unit/utils/color.test.js
@@ -20,8 +20,10 @@
  *
  */
 import {
-	generateTextColorForRGBString,
+	generateTextColorForHex,
 	generateTextColorForRGB,
+	hexToRGB,
+	isLight,
 	uidToHexColor,
 	detectColor
 } from '../../../../src/utils/color.js'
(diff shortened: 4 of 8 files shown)


georgehrke:
cc @nextcloud/designers

georgehrke:
@justin-sleep That looks really cool. Thx a lot for your pull-request. I will give this a review tomorrow.

tcitworld:
UI looks correct. Needs a check that ellipsis is still handled correctly and that it displays correctly in day view.

justin-sleep:
@tcitworld I've added a fix for long calendar names colliding with the icon:




As for the day view, this is what it looks like:

skjnldsv:
I think it looks great!

tcitworld:
It seems not very visible on day view, but I don't see any better way.

georgehrke:
What about giving `span.icon-event-reminder` an opacity of `.6` by default and `1` on hover? I'm not sure if we need that icon so prominently every time.

src/fullcalendar/eventSourceFunction.js
```diff
@@ -52,6 +52,10 @@ export function eventSourceFunction(calendarObjects, calendar, start, end, timez
 				classNames.push('fc-event-nc-tentative')
 			}
 
+			if (object.getAlarmList().length > 0) {
```
georgehrke:
```suggestion
			if (object.hasComponent('VALARM')) {
```

Does the same without actually loading all the alarms :)

tcitworld:
> What about giving span.icon-event-reminder an opacity of .6 by default and 1 on hover? I'm not sure if we need that icon so prominently every time.

I feel like it will have mixed results depending on event displayed color.

src/fullcalendar/eventRender.js
```diff
@@ -29,6 +29,16 @@
  * @param {Node} el The HTML element
  */
 export default function({ event, el }) {
+	if (el.classList.contains('fc-event-nc-alarms')) {
+		const notificationIcon = document.createElement('span')
+		notificationIcon.classList.add('icon-event-reminder')
+		if (el.style.color === 'rgb(0, 0, 0)') {
```
georgehrke: This will break, as soon we decide to change the [text colors](https://github.com/nextcloud/calendar/blob/master/src/utils/color.js#L47), which we probably will at some point since `#000000` is not the default text color in Nextcloud.

Furthermore I'm not sure that you can rely on the browser returning `'rgb(0, 0, 0)'` here. Others might return `'#000000'` or `'rgba(0, 0, 0, 0)'` or `'black'`.

Instead, you should add the information whether or not to use the dark icon in `event.extendedProps`

raimund-schluessler: Couldn't you use the svg icon api similar to https://github.com/nextcloud/tasks/blob/master/css/src/style.scss#L1174.

justin-sleep: I've modified the check to use a `darkText` property in `event.extendedProps`. I've also refactored the color helper functions a bit, since I needed a general function for detecting if a color was light or dark.

jancborchardt:
It looks great as it is in the screenshots! :+1: nice work @justin-sleep
I don’t think we need to adjust it further for the day view for now.

georgehrke:
@justin-sleep Can you please [squash the commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing) into a single one? :)

justin-sleep:
@georgehrke Sure, this is done. Looks like the build failed for unrelated reasons.

css/fullcalendar.scss
```diff
@@ -171,3 +171,24 @@
...
+// Reminder icon on events with alarms set
+.fc-event-nc-alarms {
+	.icon-event-reminder {
+		background-color: inherit;
+		background-position: right;
+		position: absolute;
+		top: 0;
+		right: 0;
+	}
+	.icon-event-reminder-light {
```
georgehrke: Nitpick: Can you please change this to [BEM style](https://css-tricks.com/bem-101/).

Then you can turn `.icon-event-reminder-light` to `.icon-event-reminder--light` and using SCSS, you can include it in the `.icon-event-reminder {` block as `&--light`.

css/fullcalendar.scss
```diff
@@ -171,3 +171,24 @@
...
+		background-image: var(--icon-calendar-reminder-fff);
+	}
+	.icon-event-reminder-dark {
```
georgehrke: Same here about [BEM style](https://css-tricks.com/bem-101/).

tests/javascript/unit/utils/color.test.js
```diff
@@ -42,6 +43,10 @@ describe('utils/color test suite', () => {
 		expect(generateTextColorForRGB({ red: 0, green: 255, blue: 0 })).toEqual('#000000')
 	})
 
+	it('should provide a RGB object for a hex string', () => {
+		expect(hexToRGB('#C98879')).toEqual({ red: 201, green: 136, blue: 121 })
+	})
+
```
georgehrke: Can you please add a test for `isLight` as well?

src/fullcalendar/eventRender.js
```diff
@@ -29,6 +29,16 @@
  * @param {Node} el The HTML element
  */
 export default function({ event, el }) {
+	if (el.classList.contains('fc-event-nc-alarms')) {
+		const notificationIcon = document.createElement('span')
+		notificationIcon.classList.add('icon-event-reminder')
+		if (event.extendedProps.darkText) {
+			notificationIcon.classList.add('icon-event-reminder-dark')
```
georgehrke: Needs to be adjusted when using BEM.

src/fullcalendar/eventRender.js
```diff
@@ -29,6 +29,16 @@
...
+		} else {
+			notificationIcon.classList.add('icon-event-reminder-light')
```
georgehrke: Needs to be adjusted when using BEM.

justin-sleep:
@georgehrke I forgot the dark theme existed! I've worked around it by accessing the SVG icons by URL rather than using the variables. I've also made the requested small changes.

css/fullcalendar.scss
```diff
@@ -171,3 +171,24 @@
...
+// Reminder icon on events with alarms set
+.fc-event-nc-alarms {
+	.icon-event-reminder {
+		background-color: inherit;
+		background-position: right;
+		position: absolute;
+		top: 0;
+		right: 0;
+		&--light {
+			background-image: url('/svg/calendar/reminder?color=ffffff')
```
georgehrke: This will break if there is index.php in the url or if Nextcloud is installed in a subdirectory.

georgehrke:
```suggestion
			@include icon-color('reminder', 'calendar', '#fffffe', 1);
```

css/fullcalendar.scss
```diff
@@ -171,3 +171,24 @@
...
+		}
+		&--dark {
+			background-image: url('/svg/calendar/reminder?color=000000')
```
georgehrke:
```suggestion
			@include icon-color('reminder', 'calendar', '#000001', 1);
```

Issue #496, opened 2017-06-03
Issues with jstz


daniellukaszgmailcom:
### Steps to reproduce
1. Open the calendar on firefox: .../index.php/apps/calendar/ or .../index.php/apps/calendar

### Expected behaviour
It should load the calendar app with my events on correct timezone

### Actual behaviour
Calendar is loading with wrong timezone and message say:
"You are in an unknown timezone (Etc/GMT-1), falling back to UTC"

### Server configuration
**Operating system**: raspbian 4.9.24-v7+
**Web server:** apache 2.4.10 (Raspbian)
**Database:** mysql 5.5.54-0+deb8u1
**PHP version:** 5.6.30-0+deb8u1
**Server version:** 12.0.0.29
**Calendar version:** 1.5.3
**Updated from an older installed version or fresh install:** fresh (fresh installation was performed because of this issue)
**Signing status (ownCloud/Nextcloud 9.0 and above):** No errors have been found.

**List of activated apps:** 
Enabled:
  - activity: 2.5.2
  - admin_audit: 1.2.0
  - audioplayer: 2.0.1
  - bookmarks: 0.10.0
  - bruteforcesettings: 1.0.2
  - calendar: 1.5.3
  - checksum: 0.3.4
  - comments: 1.2.0
  - contacts: 1.5.3
  - dav: 1.3.0
  - encryption: 1.6.0
  - external: 2.0.3
  - federatedfilesharing: 1.2.0
  - federation: 1.2.0
  - files: 1.7.2
  - files_accesscontrol: 1.2.4
  - files_external: 1.3.0
  - files_pdfviewer: 1.1.1
  - files_reader: 1.0.4
  - files_sharing: 1.4.0
  - files_snapshots: 0.1.1
  - files_texteditor: 2.4.1
  - files_trashbin: 1.2.0
  - files_versions: 1.5.0
  - files_videoplayer: 1.1.0
  - firstrunwizard: 2.1
  - gallery: 17.0.0
  - gpxedit: 0.0.6
  - groupfolders: 1.0.2
  - keeweb: 0.4.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.0.0
  - mail: 0.6.4
  - news: 11.0.5
  - nextcloud_announcements: 1.1
  - notes: 2.2.0
  - notifications: 2.0.0
  - oauth2: 1.0.5
  - onlyoffice: 1.0.4
  - ownbackup: 17.5.0
  - ownpad: 0.5.10
[...]

PR #1765, opened 2020-01-09
Receive Timezone from Intl.DateTimeFormat if available

georgehrke:

### DIFF
--- a/src/services/timezoneDetectionService.js
+++ b/src/services/timezoneDetectionService.js
@@ -27,6 +27,13 @@ import jstz from 'jstz'
  * @returns {String} Current timezone of user
  */
 export default () => {
+	if (window.Intl && typeof window.Intl === 'object') {
+		const { timeZone } = Intl.DateTimeFormat().resolvedOptions()
+		if (timeZone) {
+			return timeZone
+		}
+	}
+
 	const determinedTimezone = jstz.determine()
 	if (!determinedTimezone) {
 		return 'UTC'


src/services/timezoneDetectionService.js
```diff
@@ -27,6 +27,13 @@ import jstz from 'jstz'
  * @returns {String} Current timezone of user
  */
 export default () => {
+	if (window.Intl && typeof window.Intl === 'object') {
+		const { timeZone } = Intl.DateTimeFormat().resolvedOptions()
```
tcitworld: Not sure if Intl can be defined but not `Intl.DateTimeFormat`. Both seem to have the same percentage on CanIUse.

tcitworld:
We probably might as well ditch jstz (which uses `Intl.DateTimeFormat` itself at first). It seems #496 was caused by a Firefox issue where jstz didn't do anything anyway.
jstz hasn't been updated in a few years. The jstimezonedetect package seems less abandoned, but using the underlying system's settings is probably better anyway.

For reference, Firefox's `privacy.resistFingerprinting` setting will make the returned timezone to be `UTC`.

ChristophWurst:
I guess this still makes sense

https://caniuse.com/?search=Intl.DateTimeFormat looks ok right now

tcitworld:
And jstz is still quite dead.

SebastianKrupinski:
Hi. Is this PR still needed or has the issue been resolved and this can be closed?

tcitworld:
These days we should use `Intl.DateTimeFormat` only and completely remove the use jstz dependency.

So I'd say it should be kept opened and adapted.

SebastianKrupinski:
@tcitworld 

I agree we should drop jstz, but I'm going to close this PR in favour of https://github.com/nextcloud/calendar/pull/7365

PR #1805, opened 2020-01-14
Fix today indicator in day and week view

### LINKED ISSUE #1804: Today in agendaWeek and agendaDay
We don't use the default full-calendar style for styling today, because the yellow color caused too many issues for too many users on different settings of screens.

For month we just put a circle around the day, which works just fine:
0C286F05-3B47-4197-A621-89B451648FDD

In agendaWeek, we currently have the blue box with text:
CE69CB65-466B-4869-A84B-52C30D5B32CF

That used to look somewhat fine, but it's a problem since we replaced the header bar with a gradient.

One option would be to remove the background and just have a bold font:
0F3BEE52-55A2-456B-BC05-6E4963844C56
Please note that in addition to the bold font, today is always including the now indicator.

@nextcloud/designers What's your opinion on this?

georgehrke:
fixes #1804 

98288A3C-4915-4E20-8EBE-592510183EE0

1DCA1064-9997-409E-945A-08A43E5FE2D8

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -99,6 +99,10 @@
 	white-space: inherit !important;
 }
 
+.fc .fc-today {
+	opacity: 1;
+}
+
 .fc-state-highlight.fc-day-number,
 .fc tbody tr,
 .fc tbody tr:hover,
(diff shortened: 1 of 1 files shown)


css/fullcalendar.scss
```diff
@@ -128,10 +128,12 @@
...
 .fc-day-header.fc-today {
-	background-color: var(--color-primary) !important;
-	color: var(--color-primary-text);
-	opacity: 1;
-	font-weight: bold !important;
+	> a, span {
+		background-color: var(--color-primary);
+		border-radius: var(--border-radius-pill);
+		padding: 2px 6px;
+		color: var(--color-primary-text);
```
jancborchardt: We should add `font-weight: bold;` here so it’s like in the month view. :)

Also, we need to add a general:
```
.fc .fc-today {
    opacity: 1;
}
```
So that the blue is the proper color and not slightly transparent (lighter) blue. :)

georgehrke: Done 👍

PR #1809, opened 2020-01-14
Hide vertical scroll bar in Firefox

georgehrke:
On Master, Firefox has a vertical scrollbar, although all content is visible.
This pull-request disables vertical scrolling.

### DIFF
--- a/css/global.scss
+++ b/css/global.scss
@@ -23,3 +23,9 @@
 .toast-calendar-multiline {
 	white-space: pre-wrap;
 }
+
+#content.app-calendar {
+	> div#app-content {
+		overflow-x: hidden;
+	}
+}


tcitworld:
I still get vertical and horizontal ones (on the whole body)
Getting rid of 
```css
.fc * {
  box-sizing: content-box !important;
}
```
removes them https://github.com/nextcloud/calendar/blob/bugfix/noid/firefox_vertical_scroll/css/fullcalendar.scss#L171-L173

georgehrke:
But removing `content-box` causes plenty of issues in Chrome:
Overlapping time-labels:
C80B2A20-7780-4416-A7AA-7407269B7458

Events not properly positioned:
42B1CE6A-6905-4705-A715-39D1CB22EC5F

tcitworld:
Let's get this in right now, and create an issue to try to pin point the exact issue and fix it later (scrollbars are nearly invisible anyway since they're slim by default).

georgehrke:
The problem is that full-calendar is using a mix of `border-box` and `content-box`.

So both the server default as well as our setting is causing issues. Let me ask @jancborchardt

PR #1810, opened 2020-01-14
Always turn DTEND exclusive

georgehrke:
- Open the new event editor
- pick a new end-date with date-picker

master:
- end is one day early

this branch:
- end is correct


georgehrke:
@rullzer This fixes the bug you mentioned yesterday

PR #1811, opened 2020-01-14
Create PULL_REQUEST_TEMPLATE.md

georgehrke:
Introduce Pull-request-template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,34 @@
+## Description
+
+Please give a brief summary of your changes and which issue they address.
+
+Fixes # (issue)
+
+### Type of change
+
+Please delete options that are not relevant.
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] This change requires a documentation update
+
+### How to test / use your changes?
+
+Please provide instructions how to test your changes
+
+1. Step A
+2. Step B
+3. Step C
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+Please provide instructions how to test your changes
+
+1. Step A
+2. Step B
+3. Step C
+
+### Checklist:
+
+- [ ] My code follows the style guidelines of this project
+- [ ] I signed off my changes (`git commit -sm "Your commit message"`)
```
tcitworld: Using `-m` for commit message is bad habit, probably not best to show it here.

raimund-schluessler: Why is this a bad habit? I still do this, what is the alternative?

georgehrke: @raimund-schluessler If you don't provide m, it will open `$EDITOR` and you can type your message. But I have to say that I'm also using `-m`

tcitworld: Yeah, using `$EDITOR` encourages having a long enough message with extra lines that might be useful. Also `$EDITOR` lists (as comments) changed files while you're writing your message.

I use `-m` only for special cases when it doesn't matter much. Anyway, not much important. But here's a blog post about `git commit` if you want to read on. :) https://chris.beams.io/posts/git-commit/

georgehrke: `-m` is actually what we also recommend in https://github.com/nextcloud/server#development-setup-

tcitworld: :man_shrugging:

PR #1813, opened 2020-01-15
Release 2.0.0 RC

georgehrke:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,40 @@
+## 2.0.0 RC1 - 2020-01-15
+### Fixed
+- Hide horizontal scrollbar in Firefox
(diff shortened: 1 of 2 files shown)


tcitworld:
> Release RC only on 18 

Are there some things that prevent having 17 compatibility for final version?

georgehrke:
> Are there some things that prevent having 17 compatibility for final version?

I haven't tested it properly to be honest. All my instances are already running on 18 RC. I wanted to publish this RC on the appstore, that's why i set it to 18 for now.

georgehrke:
Let me actually remove it and create a special PR for the appstore build.
Then we don't break master

tcitworld:
> Let me actually remove it and create a special PR for the appstore build. Then we don't break master

:+1:

PR #1822, opened 2020-01-16
Also set talk_enabled in PublicViewController

georgehrke:
fixes undefined index

### DIFF
--- a/lib/Controller/PublicViewController.php
+++ b/lib/Controller/PublicViewController.php
@@ -97,15 +97,21 @@ class PublicViewController extends Controller {
 	 */
 	private function publicIndex(string $token,
 								 string $renderAs):TemplateResponse {
+		$defaultInitialView = $this->config->getAppValue($this->appName, 'currentView', 'dayGridMonth');
+		$defaultShowWeekends = $this->config->getAppValue($this->appName, 'showWeekends', 'yes');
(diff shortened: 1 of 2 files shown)


tcitworld:
Can you also get the defaults values from `getAppValue` (#1816) here ?

georgehrke:
Sure, will add that.

lib/Controller/PublicViewController.php
```diff
@@ -97,15 +97,21 @@ public function publicIndexForEmbedding(string $token):TemplateResponse {
 	 */
 	private function publicIndex(string $token,
 								 string $renderAs):TemplateResponse {
+		$defaultInitialView = $this->config->getAppValue($this->appName, 'currentView', 'dayGridMonth');
+		$defaultShowWeekends = $this->config->getAppValue($this->appName, 'showWeekends', 'yes');
+		$defaultWeekNumbers = $this->config->getAppValue($this->appName, 'showWeekNr', 'no');
+		$defaultSkipPopover = $this->config->getAppValue($this->appName, 'skipPopover', 'no');
```
tcitworld: We had `'skip_popover' => true` as default previously, but this is fine.

georgehrke: Yeah, let me change that again.

PR #1823, opened 2020-01-16
Set no rewrite for generateUrl, in case URL contains index.php despite configured rewrite

georgehrke:
If you configure URL rewrite, `generateURL` will by default always skip the index.php
Nonetheless, you can still open the URL with the `index.php`

If you access the URL with `index.php`, the router base will be wrong and routing will break:

http://nextcloud.local/index.php/apps/calendar/p/tsQnYebgwxrF67QK/dayGridMonth/now is broken
http://nextcloud.local/apps/calendar/p/tsQnYebgwxrF67QK/dayGridMonth/now works

This branch double checks for index.php and disables rewriting if there is one.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -21,7 +21,7 @@
 
 import Vue from 'vue'
 import Router from 'vue-router'
-import { generateUrl } from '@nextcloud/router'
+import { getRootUrl, generateUrl } from '@nextcloud/router'
 
 import Calendar from './views/Calendar'
 import EditSimple from './views/EditSimple'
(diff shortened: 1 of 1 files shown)


tcitworld:
I'm fine with this, but I wonder why wouldn't it "just" work [like in Contacts](https://github.com/nextcloud/contacts/blob/master/src/router/index.js#L32-L34) ? @skjnldsv

georgehrke:
I was wondering that myself, my best guess is that Contacts only has one root route and we have multiple ones, for normal, embed and public. That's probably causing issues in matching the route when the base is slightly off.

PR #1866, opened 2020-01-20
Sort categories alphabetically

### LINKED ISSUE #1827: List of categories not sorted alphabetically

georgehrke:
Fixes #1827 

| Before | After |
|:---------:|:------:|
|6590F35F-C818-48AF-83FD-A758294A7544|7046E49F-4F19-4FD3-8934-5CA25E9ACBE8|


tcitworld:
Not capitalized categories will be be shown after the capitalized ones, shouldn't we mix them instead ?

georgehrke:
@tcitworld Good point, should be fixed now.

PR #1873, opened 2020-01-21
Illustrations: add a few more activities

shtrom:
* Agile ceremonies
* Lunch, Conference, Pair (as in pair programming)
* Some short-hands/misspellings

Fixes # (issue)

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

1. Create events in the calendar with strings matching the newly added string
2. The chosen illustration should be used rather than the default.

### Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/src/utils/illustration.js
+++ b/src/utils/illustration.js
@@ -164,7 +164,6 @@ const data = [{
 		'Pointing',
 		'Retrospective', 'Retro',
 		'Review',
-		'Stand-up', 'Standup',
 		// TRANSLATORS This string is used for matching the event title to an illustration
 		t('calendar', 'Meeting'),
 		// TRANSLATORS This string is used for matching the event title to an illustration
@@ -177,8 +176,6 @@ const data = [{
 		t('calendar', 'Retrospective'),
 		// TRANSLATORS This string is used for matching the event title to an illustration
 		t('calendar', 'Review'),
-		// TRANSLATORS This string is used for matching the event title to an illustration
-		t('calendar', 'Stand-up'),
 	],
 	illustrationNames: [
 		'meeting',
--- a/src/utils/illustration.js
+++ b/src/utils/illustration.js
@@ -160,10 +160,25 @@ const data = [{
 	strings: [
 		'Meeting',
 		'Meet',
+		'Planning',
+		'Pointing',
+		'Retrospective', 'Retro',
+		'Review',
+		'Stand-up', 'Standup',
 		// TRANSLATORS This string is used for matching the event title to an illustration
 		t('calendar', 'Meeting'),
 		// TRANSLATORS This string is used for matching the event title to an illustration
 		t('calendar', 'Meet'),
+		// TRANSLATORS This string is used for matching the event title to an illustration
+		t('calendar', 'Planning'),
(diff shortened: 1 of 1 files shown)


src/utils/illustration.js
```diff
@@ -160,10 +160,25 @@ const data = [{
...
+		// TRANSLATORS This string is used for matching the event title to an illustration
+		t('calendar', 'Planning'),
+		// TRANSLATORS This string is used for matching the event title to an illustration
+		t('calendar', 'Pointing'),
+		// TRANSLATORS This string is used for matching the event title to an illustration
+		t('calendar', 'Retrospective'),
+		// TRANSLATORS This string is used for matching the event title to an illustration
+		t('calendar', 'Review'),
+		// TRANSLATORS This string is used for matching the event title to an illustration
+		t('calendar', 'Stand-up'),
```
georgehrke: I'm not sure if this is a bit too ambiguous. It will also match "Stand-up comedy" and "Stand-up paddling"

tcitworld: Agreed. Please remove this one for now @shtrom

shtrom: Ah, that's unfortunate, that was the one that motivated to make those changes in the first place. Shall I also remove “standup” above?

georgehrke: Yes please

raimund-schluessler: Maybe `wakeup` is more unambiguous and still fits.

georgehrke: @raimund-schluessler This is not supposed to mean wakeup. It's supposed to be Standup as in agile software development: https://en.wikipedia.org/wiki/Stand-up_meeting

raimund-schluessler: Ah, I see, then `stand-up` *really* is ambiguous 😄

georgehrke:
@shtrom Can you please make the requested changes? Then this is ready to go in! :)

src/utils/illustration.js
```diff
@@ -521,17 +539,20 @@ const data = [{
...
+		// TRANSLATORS This string is used for matching the event title to an illustration
+		t('calendar', 'Conference'),
 	],
 	illustrationNames: [
 		'professor',
 	],
 }, {
 	strings: [
-		'Photograph',
+		'Photograph', 'Photo',
```
georgehrke: Put 'Photo' on a new line please

shtrom: As this one is just a shortened version, I thought it would make sense to specifically not put it on a new line.

Note that there also isn't a translation for it either, for the same reason. Should there be one?

src/utils/illustration.js
shtrom: Done. Though I wonder how we could handle those a bit better. Perhaps some per-calendar or per-user setting to allow disambiguating? Well beyond this PR, but I guess that would be a way to make it work.

src/utils/illustration.js
georgehrke: Ah, that makes sense in that case. Nevermind then. :)

PR #1879, opened 2020-01-21
Increment max-version to 19

georgehrke:
Needed for https://github.com/nextcloud/server/pull/18765

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,12 @@ matrix:
       env: "TEST_SUITE=LINT-PHP"
     - php: 7.4snapshot
       env: "TEST_SUITE=TEST-JS"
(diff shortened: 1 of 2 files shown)


georgehrke:
Let me actually update Travis.

.travis.yml
```diff
@@ -34,6 +34,12 @@ matrix:
       env: "TEST_SUITE=LINT-PHP"
     - php: 7.4snapshot
       env: "TEST_SUITE=TEST-JS"
+    - php: 7.4snapshot
```
tcitworld: 7.4 is still not available?

georgehrke: It is, but using it caused some weird issues with composer

PR #1887, opened 2020-01-23
Up max-version to 19 for latest server master

jancborchardt:
Please check @georgehrke @tcitworld :)


tcitworld:
You missed #1879 :)

jancborchardt:
So strange, I thought I pulled master just before. Must have missed this one cause I updated my whole instance and all apps to the new dev. :D

PR #1917, opened 2020-01-26
Vary illustration when multiple images match

rspencer01:
Randomly select an illustration from the images that match the
title of the event.

## Description

As alluded to in https://github.com/nextcloud/calendar/blob/b9b47a17c1a1bd663011be7079d87533e94abcec/src/utils/illustration.js#L58 there can be more than one valid illustration for an event.  This pull request randomly selects an illustration from those allowed.

### Type of change

1. Create an event that matches a category with multiple illustrations e.g. "Bicycle to work"
2. View event multiple times
3. The illustration should change

### Checklist:


georgehrke:
Hey,

Thx for the pull-request, but my plan was actually to use some sort of hash function to always map to the same image, if there are multiple. Otherwise I would find it confusing if every time you open an event the image changes.

georgehrke:
Something like this: https://github.com/nextcloud/calendar/blob/master/src/utils/uidToColor.js#L42L61

rspencer01:
Ah, I understand.  May I suggest that proper hashing is overkill and something like the last character might suffice?

georgehrke:
> Ah, I understand. May I suggest that proper hashing is overkill and something like the last character might suffice?

If you take the last character, convert it to an int (e.g. with `charCodeAt(0)`) and take that modulo the number of available illustrations, it's also a hashing function, just a very simple one. (which is obviously neither secure nor well distributed).

The function i referenced above is already used all over the place and it doesn't cause any performance issue in case that's what you are worrying about.

rspencer01:
I can only find one reference to `hashToInt` in the codebase, and am reticent to pull it out into a globally available function in the name of complexity.  I've pushed the `charCodeAt` version.  If you'd like the `hashToInt` version, feel free to disregard this PR.

rspencer01:
@rspencer01 fix tests

ChristophWurst:
@rspencer01 I'm fine with this. Could you please rebase and squash? :)

rspencer01:
Squashed + rebased on master

ChristophWurst:
The tests fail again https://github.com/nextcloud/calendar/pull/1917/checks?check_run_id=1805955565 :zipper_mouth_face:

rspencer01:
:grimacing: Ah, whoops.  I'll take a look once I can get the tests running again locally.

tcitworld (approved):

PR #1933, opened 2020-01-29
Properly localize date in document title

### LINKED ISSUE #1871: Tab title not translated
### Steps to reproduce
1. Open the calendar-app in monthly overview.

### Expected behaviour
In the German translation, the tab title of the month in the respective month should be in the same translation.

### Actual behaviour
In the German translation, the English translation of the month is in the tab title.
Bildschirmfoto vom 2020-01-20 21-45-10

### Calendar app
2.0.1

**CalDAV-clients used:** (Thunderbird Lightning, DAVx5, Evolution, macOS Calendar, etc)

### Client configuration
**Browser:** (e.g. Firefox 48)
Firefox 72.0.1

**Operating system:** (e.g. Arch Linux)
Fedora 31

### Server configuration

**Operating system**: (e.g. Debian 8)
Ubuntu

**Web server:** (e.g. Apache, Nginx,...)
Apache

**Database:** (e.g. MariaDB, SQLite or PostgreSQL)
MariaDB

**PHP version:** (e.g. 7.0.3)
7.3.13

**Nextcloud Version:** (see admin page, e.g. 17.0.2)
18.0.0

georgehrke:
fixes #1871

### DIFF
--- a/src/services/windowTitleService.js
+++ b/src/services/windowTitleService.js
@@ -29,8 +29,28 @@ const originalWindowTitle = document.title
  * automatically adjusts the title of the window
  *
  * @param {VueRouter} router The VueJS Router instance
+ * @param {Store} store The vuex store
  */
-export default function(router) {
+export default function(router, store) {
+	/**
+	 * Updates the title of the window
+	 *
+	 * @param {Date} date viewed Date
+	 * @param {String} view Name of the current view
+	 * @param {String} locale Locale to be used for formatting
+	 */
+	function updateTitle(date, view, locale) {
+		const title = dateRangeFormat(date, view, locale)
+		document.title = [
+			title,
+			originalWindowTitle,
+		].join(' - ')
+	}
+
+	/**
+	 * This listens to router changes and automatically
+	 * updates the title
+	 */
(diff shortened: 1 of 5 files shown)


src/services/windowTitleService.js
```diff
@@ -29,23 +29,56 @@ const originalWindowTitle = document.title
...
+	 * @param {Date} date viewed Date
+	 * @param {String} view Name of the current view
+	 * @param {String} locale Locale to be used for formatting
+	 */
+	function updateTitle(date, view, locale) {
+		const title = dateRangeFormat(date, view, locale)
+		document.title = [
+			title,
+			originalWindowTitle,
+		].join(' - ')
```
tcitworld: Maybe use translation system (with a comment for context) instead ? Not sure how this work with RTL languages for instance.

georgehrke: Hm, I think putting up a translation like `{Date} - {windowtitle}` will cause a lot of confusion for translators. 
Imho it would be better if nextcloud/l10n provided an isRTL method, so we can simply flip it around.

PR #1958, opened 2020-02-03
query Application instead of constructing it

icewind1991:
Stops "RuntimeException: App class OCA\Calendar\AppInfo\Application is not setup via query() but directly" log spam

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -22,5 +22,5 @@ declare(strict_types=1);
  */
 namespace OCA\Calendar\AppInfo;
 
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
Couldn't we get rid of `appinfo/app.php` completely by registering the navigation in `appinfo/info.xml`? Like this https://github.com/nextcloud/app-tutorial/blob/master/appinfo/info.xml#L17-L22

tcitworld:
@raimund-schluessler Indeed, done in #1979

PR #1979, opened 2020-02-10
Move navigation to appinfo

tcitworld:
Superseeds #1958 as suggested by @raimund-schluessler


appinfo/info.xml
```diff
@@ -34,6 +34,15 @@
 	<dependencies>
 		<nextcloud min-version="17" max-version="19" />
 	</dependencies>
+	<navigations>
+		<navigation>
+			<id>calendar</id>
+			<name>Calendar</name>
+			<order>5</order>
+			<route>calendar.view.index</route>
+			<icon>calendar.svg</icon>
```
raimund-schluessler: Don't ask me why, but the `info.xml` will not validate in this order. What works is `id->name->route->icon->order`.

PR #1980, opened 2020-02-10
Enable PHP 7.4 stable on Travis

tcitworld:


tcitworld:
https://travis-ci.community/t/some-extensions-are-missing-in-php-7-4-0-zip-gmp-sodium/6320/22

tcitworld:
Can we move to `xenial`?

raimund-schluessler:
Another option would be to move away from travis completely and use Github actions: https://github.com/nextcloud/app-tutorial/tree/master/.github/workflows

georgehrke:
I think we should go for @raimund-schluessler's suggestion and move to GitHub actions.

@tcitworld Any objections?

tcitworld:
> @tcitworld Any objections?

Nope, I never played much with Github Actions, that's why I didn't go ahead.

PR #2000, opened 2020-02-17
Update remaining fullcalendar packages

georgehrke:
Not sure why dependabot missed those


georgehrke:
CI is happy, let's merge

PR #2023, opened 2020-02-25
Autofocus event title on popover

### LINKED ISSUE #1930: new event focus directly on subject
when creating a new event, the focus should be on the subject so that the user can write the content without additional clicks.

cursor on the yellow field (see attach) without other steps (additional click)

nc3

TY

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/87723053-new-event-focus-directly-on-subject?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
fixes #1930

### DIFF
--- a/src/directives/focus.js
+++ b/src/directives/focus.js
@@ -1,0 +1,29 @@
+/**
+ * @copyright Copyright (c) 2020 Georg Ehrke
+ *
+ * @author Georg Ehrke <oc.list@georgehrke.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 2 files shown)


src/directives/focus.js
```diff
@@ -0,0 +1,29 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+export default {
+	inserted(el) {
+		setTimeout(() => {
```
georgehrke: It works with a  timeout, but not without.
There must be something, which takes away the focus immediately.

@tcitworld Do you have any idea what that could be?
My best guess would be that it's related to V-Popover, but that requires further debugging.

laurens94:
Perhaps the autofocus behaviour should only take place whenever the user creates a new event (not when opening an existing event), to prevent mistakes. Or is that already the case with this commit?

georgehrke:
@laurens94 Why would you not focus the title when editing an existing an event?

laurens94:
Nevermind! I often edit an event just to change the time, but I guess it doesn't matter that much if the title is only focused instead of completely selected. 

Thanks for the fix 👍

jancborchardt:
So is this ready to merge? :) @nextcloud/calendar

Zaxooy:
Worked for me, can recommend this. :-)

I'm not very exerienced with github. So we're waiting for @tcitworld to merge this?

ChristophWurst:
@GretaD could you rebase and give this a good test? I think we can integrate this then :)

GretaD (approved):

Issue #2013, opened 2020-02-24
Calendar app cannot add repeats to an event after the event is created


deanforrest:
After an event is created and a repeat is added to it through the web interface, the repeat will not appear in the calendar, and the repeat section of the event will no longer be available. However, running through the steps below using the iOS calendar through CalDAV works perfectly.

### Steps to reproduce
1. Create an event on the calendar, click "save"
2. Click on the previously created event, click "more", click on the "Repeat" tab and add a repeat for any range of time. Adding an "End repeat" option does not affect the resulting outcome.

### Expected behaviour
The event should add multiple occurrences on the calendar based on the repeat selected in the "more" section.

### Actual behaviour
Occurrences are not added to the specified calendar, and if the user navigates to the "Repeat" section of the event, "This event is the recurrence-exception of a recurrence-set. You can not add a recurrence-rule to it." will appear instead of the repeat options. Also, the simplified editor will not show the "Update this occurrence" or "Update this and all future" options in the simplified editor.

If the user follows steps 1 through 2 again but adds a time range for the given day (e.g. 4 PM to 7 PM), it cannot be deleted through the web interface, only through the calendar app on an iPhone connected to Nextcloud through CalDAV.

### Calendar app
**Calendar app version:** 2.0.1

**CalDAV-clients used:** CalDAV via iOS

### Client configuration
**Browser:** Firefox 74.04b & Google Chrome 79.0.3945.130

**Operating system:** Windows 10

### Server configuration

**Operating system**: Ubuntu Server 18.04.3

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.2.24

**Nextcloud Version:** 18.0.0

justin-sleep:
Here's an example of an event Calendar generated that produces this message for me:
```
BEGIN:VCALENDAR
PRODID:-//IDN nextcloud.com//Calendar app 2.0.1//EN
CALSCALE:GREGORIAN
VERSION:2.0
BEGIN:VEVENT
CREATED:20200224T033714Z
DTSTAMP:20200224T033714Z
LAST-MODIFIED:20200224T033714Z
SEQUENCE:1
UID:d287c5f9-e8f2-4323-96c7-9703a9a59b64
DTSTART;VALUE=DATE:20200106
DTEND;VALUE=DATE:20200107
SUMMARY:Test Event
RECURRENCE-ID;VALUE=DATE:20200106
END:VEVENT
END:VCALENDAR
```

[...]

Issue #1929, opened 2020-01-29
Preserve duration when editing start time


castletonroad:
### Steps to reproduce
1. Set up repeating event with duration less than 1 hour: Monday 7.00-7.15am, repeat weekly on Mondays, end after 5 occurrences
2. Edit repeating event: change start date into the future by 10 weeks; select update this and all future events

### Expected behaviour
1. Repeating series of events should 'shift' 10 weeks into the future, with no other changes to the event

### Actual behaviour
1. Duration resets to 1 hour for the series.

### Calendar app
2.0.1

### Client configuration
Firefox 72.0.2

**Operating system:** (e.g. Arch Linux)
Raspbian Debian Buster

### Server configuration

**Operating system**: (e.g. Debian 8)
Debian Buster

**Web server:** (e.g. Apache, Nginx,...)
Apache

**Database:** (e.g. MariaDB, SQLite or PostgreSQL)
MariaDB

**PHP version:** (e.g. 7.0.3)
7.3.11-1

**Nextcloud Version:** (see admin page, e.g. 17.0.2)
17.0.2

**Updated from an older installed version or fresh install:**
Update from 16.x.x

**List of activated apps:**
Enabled:
  - activity: 2.10.1
  - calendar: 2.0.1
  - cloud_federation_api: 1.0.0
  - comments: 1.7.0
  - contacts: 3.1.8
  - dav: 1.13.0
  - federatedfilesharing: 1.7.0
  - files: 1.12.0
  - files_pdfviewer: 1.6.0
  - files_trashbin: 1.7.0
  - files_versions: 1.10.0
  - gallery: 18.4.0
  - logreader: 2.2.0
  - lookup_server_connector: 1.5.0
  - mail: 1.1.1
  - nextcloud_announcements: 1.6.0
  - notifications: 2.5.0
  - oauth2: 1.5.0
  - password_policy: 1.7.0
  - privacy: 1.1.0
  - provisioning_api: 1.7.0
  - theming: 1.8.0
  - twofactor_backupcodes: 1.6.0
  - twofactor_totp: 4.1.2
  - updatenotification: 1.7.0
  - viewer: 1.2.0
  - workflowengine: 1.7.0
Disabled:
  - accessibility
  - admin_audit
  - encryption
  - federation
  - files_external
  - files_rightclick
  - files_sharing
  - files_videoplayer
  - firstrunwizard
  - recommendations
  - serverinfo
  - sharebymail
  - support
  - survey_client
[...]

Issue #577, opened 2017-08-25
[Feature request] Being able to modify the default `minTime`, `maxTime` and `slotDuration` of fullcalendar with a GUI


AlexandreBonneau:
Whenever I upgrade nextcloud, I need to manually fiddle in the `dist/fullcalendar.js` file to modify the *not-so-great* default values.

Using the info from https://fullcalendar.io/docs/agenda/, I usually modify the start (`minTime`) and end (`maxTime`) time of the day (ie. 6:00 and 19:00), and the `slotDuration` to `'00:15:00'` instead of `'00:30:00'`.

Would it be possible to have a simple GUI to do that?

Context: I use a 15 minutes slot duration because I have lots of appointments and those are set every 15 minutes.

georgehrke:
cc @jancborchardt about the defaults.

jancborchardt:
The default of 30min is pretty standard. Introducing a setting for that would obscure more important settings.

Can you provide screenshots of how (if) other widely-used and well-designed software provides an option for that?

(Instead of a visible setting, one thing we could do is notice if there are a lot of <30min appointments, and adjust the default time accordingly.)

AlexandreBonneau:
The 30 minutes duration is indeed standard when you use the calendar to setup meetings and the like.
However, as soon as you use the calendar to set appointments (for instance appointments for a doctor or hospital office), then 15 minutes is mandatory (and I even know ophthalmologists who set appointments every 6 minutes!).

This is important for them, but the two other settings (`minTime` and `maxTime`) are used quite often too: you really do not need to clutter your view with non-office hours. `00:00` to `06:00` and `19:00` to `23:59` are usually never used in most offices, so why bother showing those (or even the scrollbar)?

### About other calendars
Korganizer for instance allows to use `ctrl + mousewheel` to modify the size of the viewport, which allows to actually see the event titles.
This does not change the `slotDuration` per se, since it already default to 15 minutes (see the blue rectangle on the screenshot ; this is the default when clicking on the view), but only zoom in/out.

Before:
![korganizer_ss1](https://user-images.githubusercontent.com/7248877/29947627-5528bc1c-8e46-11e7-93e4-eeb6507a79a7.png)

After:
![korganizer_ss2](https://user-images.githubusercontent.com/7248877/29947626-5519cd1a-8e46-11e7-86a4-92b846408f53.png)

A real-life usecase:
[...]

PR #2032, opened 2020-02-28
Release 2.0.2

georgehrke:
Preparing the 2.0.2 release ...

I would still like to include the following in 2.0.2:
- [x] #2013 
- [x] #1929 
- [x] #1891 
- [x] #71 
- [x] update dependencies
- [x] https://github.com/nextcloud/calendar/pull/2042

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,64 @@
+# 2.0.2 - 2020-03-02
+### Added
+- Recognize Gym as event title for illustrations
+  [#1888](https://github.com/nextcloud/calendar/issues/1888)
+- Improve illustration matching for less false positives
+  [#1916](https://github.com/nextcloud/calendar/issues/1916)
+- Add illustrations keywords related to agile development
+  [#1873](https://github.com/nextcloud/calendar/pull/1873)
+- Add hint to new calendar dropdown that subscriptions are read-only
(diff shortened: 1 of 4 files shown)


georgehrke:
@tcitworld Should we also address https://github.com/nextcloud/calendar/issues/577 by adding a simple select option for now to pick the slot height: Like: 10, 15, 20, 30 mins.

This was the second biggest issue we got negative feedback about after the "Show more"

tcitworld:
> Should we also address #577 by adding a simple select option for now to pick the slot height: Like: 10, 15, 20, 30 mins.

We definitely should, yes. Maybe change back the default value as well, but I don't care much about this one.

At the same time I'm finding the Settings area getting too small for all of the settings, so if we add this option now we'll have to rethink how settings are handled (in a popup? move some options out of the app into general settings?). Can we work on this next?

georgehrke:
> At the same time I'm finding the Settings area getting too small for all of the settings, so if we add this option now we'll have to rethink how settings are handled (in a popup? move some options out of the app into general settings?). Can we work on this next?

Definitely. Especially if we also add BusinessHours (for VAvailability in scheduling) and a UI for caldav proxy.

I think a popover will do fine.

georgehrke:
Requires https://github.com/nextcloud/calendar/pull/2046

PR #2042, opened 2020-02-28
Add slotDuration to ConfigController and ViewControllers

georgehrke:
Related to #577 (not fixing though, as it's not adding automatic detection)

### DIFF
--- a/src/components/AppNavigation/Settings.vue
+++ b/src/components/AppNavigation/Settings.vue
@@ -58,6 +58,16 @@
 				@update:checked="toggleWeekNumberEnabled">
 				{{ $t('calendar', 'Show week numbers') }}
 			</ActionCheckbox>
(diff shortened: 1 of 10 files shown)


src/components/AppNavigation/Settings.vue
```diff
@@ -139,6 +153,30 @@ export default {
 		settingsTitle() {
 			return this.$t('calendar', 'Settings & import').replace(/&amp;/g, '&')
 		},
+		slotDurationOptions() {
+			return [{
+				label: this.$t('calendar', '5 minutes'),
```
tcitworld: MomentJS should be able to provide these translations out of the box, right?

georgehrke: excellent point!

raoulbhatia:
Thanks! 🥳

PR #2086, opened 2020-03-18
Re-add @NoSameSiteCookieRequired to embedding route

georgehrke:
This must have gotten lost during vue rewrite.


georgehrke:
@tcitworld Do you generally have an overview of what's left to make embedding work properly out of the box? 

Didn't we also face some CORS issues?

PR #2132, opened 2020-03-31
Move AppCode check to Github actions

georgehrke:


georgehrke:
CI is happy (except for travis, because there no longer is a `.travis.yml`)
Merging.

PR #2133, opened 2020-03-31
Report test coverage for PHP

georgehrke:


raimund-schluessler:
I don't know if it helps, but this is how the coverage is uploaded for the Tasks app:
https://github.com/nextcloud/tasks/blob/master/.github/workflows/phpunit.yml#L67-L72

georgehrke:
@raimund-schluessler Thx for the link! I tried the codecov GitHub action first, but it's causing issues because we are cloning the calendar repo into a subfolder and you can't apply `working-directory` to a step if there is no `run`.

georgehrke:
CI and Codecov are happy, merging.

PR #2154, opened 2020-04-07
Fix view icons

skjnldsv:
| Before | After |
|:---------:|:------:|
|Kazam_screenshot_00003|Kazam_screenshot_00004|


skjnldsv:
That was fast :p

PR #2155, opened 2020-04-07
Keep settings open while using the import modal

georgehrke:


ChristophWurst (approved):
Works, but also on master

:shrug: @tcitworld what shall we do?

szaimen:
I'll have a look at this

szaimen:
> but also on master

no, it does not work on main

szaimen (approved):
works

st3iny (approved):
Tested and works.

Issue #402, opened 2017-04-02
List view [$45]


RiSK1979:
Please add an agenda view to the nextcloud calendar. In case you don't know, what I'm referring to, an agenda view is basically a list of the upcoming events in your calendar.

The view button could be placed right beside the day view, week view, month view buttons, e.g.

Of course, this view type should also be available for public calendars.

kbeerepoot:
I totally agree with this feature request. I would very much like to be able to use an Agenda view (or List view, as others would call it). The point of it is that I am not much interested in the days without any events, and I assume that the Agenda view (or List view) would allow much more space to instantly display the full name and the specifics of the event (attendees and location).

georgehrke:
Please refrain from commenting when you can't add anything new. Use [Github Reactions](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) instead.

kbeerepoot:
Dear Georg,

I hope you do not mind me using this channel to express my ardent wish for this feature again. I realise it is nothing new (see my comment of 29 April 2017) but my users (and I assume many others all over the world) could really use this feature. Is there any idea about when it could possibly be implemented ?

I would love to do it myself, but unfortunately I can not program my way out of a wet paper bag.

Again, my apologies if I should not have written this comment.

jakobroehrl:
Hints for implementation:
List View: fullcalendar.io/docs/list-view-demo doc: fullcalendar.io/docs/list-view

todo: implement own Full-calendar view: fullcalendar.io/docs/custom-view-with-js
github.com/fullcalendar/fullcalendar/blob/master/packages/list/src/ListEventRenderer.ts#L75L94

phoezelle:
Hello
Is it possible that in the top bar, near the notification icon, an agenda icon could show the next events in a small list view.

Mannshoch:
Embedding a Calendar into a Website is a cool feature.
![grafik](https://user-images.githubusercontent.com/7795891/67858081-9fc84580-fb18-11e9-9290-95551836d1f4.png)

But Inside a Website I don't need a Classic Calendar view. I need a List-view because I do not have Dates all week I wish to present, With this list-view I also wish to present some more Information.
May I could be able to set some settings inside the embed link.
eg.:
`<iframe width="400" height="215" src="/apps/calendar/embed/y5oPHcErXyr2jJ6S?view=list&range=-2d+5w&show-field=Description,category,enddate"></iframe>`

[...]

PR #2195, opened 2020-04-17
Add Keyboard shortcuts

### LINKED ISSUE #157: [Feature Request] Keyboard support

georgehrke:
Related to #157 

- [x] Add keyboard shortcuts

| Action | Shortcut |
|:---------:|:------:|
| Previous time-range | p or k |
| Next time-range | n or j |
| Today | t |
| New event | c |
| Day view | 1 or d |
| Week view | 2 or w |
| Month view | 3 or m |

Possible further shortcuts:
| Action | Shortcut |
|:---------:|:------:|
| List #402 | 4 or l |
| Search #8  | / |
| Refresh #31 | r |

I do not want to add a shortcut for deleting an event, as long as we don't provide an undo.

- [x] provide an overview for all available shortcuts 
- [ ] fix tab order of editor

### DIFF
--- a/src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderViewMenu.vue
+++ b/src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderViewMenu.vue
@@ -20,7 +20,11 @@
   -->
 
 <template>
-	<Actions :default-icon="defaultIcon" menu-align="right">
+	<Actions
+		v-shortkey="shortKeyConf"
+		:default-icon="defaultIcon"
+		menu-align="right"
+		@shortkey.native="selectViewFromShortcut">
 		<ActionButton
 			v-for="view in views"
 			:key="view.id"
@@ -57,6 +61,16 @@ export default {
 				label: this.$t('calendar', 'Month'),
 			}]
 		},
+		shortKeyConf() {
+			return {
+				timeGridDay: ['d'],
+				timeGridDay_Num: [1],
+				timeGridWeek: ['w'],
+				timeGridWeek_Num: [2],
+				dayGridMonth: ['m'],
+				dayGridMonth_Num: [3],
+			}
+		},
 		defaultIcon() {
 			for (const view of this.views) {
 				if (view.id === this.$route.params.view) {
@@ -80,6 +94,9 @@ export default {
 			}
 
 			this.$router.push({ name, params })
+		},
+		selectViewFromShortcut(event) {
+			this.selectView(event.srcKey.split('_')[0])
 		},
 	},
 }
--- a/src/components/AppNavigation/Settings/ShortcutOverview.vue
+++ b/src/components/AppNavigation/Settings/ShortcutOverview.vue
@@ -52,7 +52,7 @@
 							v-if="index2 !== (shortcut.keys.length - 1)"
 							:key="index2"
 							class="shortcut-section-item__spacer">
-							or
+							{{ t('calendar', 'or') }}
 						</span>
 					</template>
 				</span>
--- a/src/components/AppNavigation/Settings/ShortcutOverview.vue
+++ b/src/components/AppNavigation/Settings/ShortcutOverview.vue
@@ -1,0 +1,116 @@
+<!--
+  - @copyright Copyright (c) 2020 Georg Ehrke <oc.list@georgehrke.com>
+  - @author Georg Ehrke <oc.list@georgehrke.com>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+<template>
+	<Modal
+		class="shortcut-overview-modal"
+		size="middle"
+		:title="$t('calendar', 'Shortcut overview')"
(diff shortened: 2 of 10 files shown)


georgehrke:
@tcitworld Should we also add `arrowup` and `arrowdown` for `Previous time-range` and `Next time-range` or does it get confusing if we have three different keys doing exactly the same thing?

georgehrke:
Actually, the nc/vue Actions component is using `arrowup`, `arrowdown`, etc
So we should not override that.

https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Actions/Actions.vue#L79

SuperSandro2000:
I would suggest ``Ctrl + F`` for search. If you press it twice you open the normal browser search.

tcitworld:
> I would suggest Ctrl + F for search. If you press it twice you open the normal browser search.

This one is going to be used anyway since it's used everywhere else in Nextcloud.

SuperSandro2000:
Also what about customizing these controls? Not everyone likes vim style.

georgehrke:
> Also what about customizing these controls? Not everyone likes vim style.

That's why we offer both `n` and `j` (or `p` and `k`), but we are definitely not gonna start making keyboard controls customisable.

georgehrke:
825DB4FB-2C25-4B26-9A1A-8A32DB51524F

georgehrke:
This is good to review from my pov.
Let's move fixing the tab order to a different PR, since it's not really related to this PR.

src/components/AppNavigation/Settings/ShortcutOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+<template>
+	<Modal
+		class="shortcut-overview-modal"
+		size="middle"
```
tcitworld, line 24: Should be one of `'normal', 'large', 'full'` https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Modal/Modal.vue#L269

src/components/AppNavigation/Settings/ShortcutOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+				class="shortcut-section-item">
+				<span class="shortcut-section-item__keys">
+					<template
+						v-for="(keyCombination, index2) of shortcut.keys"
+						class="shortcut-section-item__key-combination">
+						<template v-for="(key, index3) in keyCombination">
+							<kbd :key="index3">{{ key }}</kbd>
+							<span
+								v-if="index3 !== (keyCombination.length - 1)"
+								:key="index3"
```
tcitworld, line 46:
```
Duplicate keys detected: '0'. This may cause an update error.
```
Probably
> Children of the same common parent must have unique keys. Duplicate keys will cause render errors.

Possible fix:
```js
:key="`${index}${index2}${index3}`"
```

src/components/AppNavigation/Settings/ShortcutOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+								class="shortcut-section-item__spacer">
+								+
+							</span>
+						</template>
+						<span
+							v-if="index2 !== (shortcut.keys.length - 1)"
+							:key="index2"
```
tcitworld, line 53:
```js
:key="`${index}${index2}`"
```

src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderViewMenu.vue
```diff
@@ -57,6 +61,16 @@ export default {
 				label: this.$t('calendar', 'Month'),
 			}]
 		},
+		shortKeyConf() {
+			return {
+				timeGridDay: ['d'],
+				timeGridDay_Num: [1],
```
tcitworld, line 67: Why different keys for numbers (`timeGridDay` vs `timeGridDay_Num`) ?

tcitworld, line 67: Seems to be an issue with `vue-shortkey`, but only numpad number keys work, not the combination <kbd>Maj</kbd> + <kbd>&</kbd> on my Azerty keyboard (event code `Digit1` vs `Numpad1`).

src/components/AppNavigation/Settings/ShortcutOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+<template>
+	<Modal
+		class="shortcut-overview-modal"
+		size="middle"
+		:title="$t('calendar', 'Shortcut overview')"
+		@close="$emit('close')">
```
tcitworld, line 26: It would be nice if the `h` shortcut could also close the popup.

src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderViewMenu.vue
georgehrke: This seems to be an issue specific to the AZERTY keyboard :/ 

On my QWERTY keyboard, both Numpad and number keys work just fine.

georgehrke: Because we need two different keys for the object. If we put `d` and `1` into one array, it wouldn't be `d` or `1` but combination of `d` and `1`

PR #2243, opened 2020-05-05
Use nc/dialogs instead of ocp.toast

### LINKED ISSUE #2063: Replace OCP.Toast with Nextcloud/dialogs
https://github.com/nextcloud/server/pull/19346

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90159759-replace-ocp-toast-with-nextcloud-dialogs?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
fixes #2063

### DIFF
--- a/src/views/Calendar.vue
+++ b/src/views/Calendar.vue
@@ -131,6 +131,10 @@ import eventLimitText from '../fullcalendar/eventLimitText.js'
 import windowResize from '../fullcalendar/windowResize.js'
 import dayRender from '../fullcalendar/dayRender.js'
 import { loadState } from '@nextcloud/initial-state'
+import {
+	showWarning,
+} from '@nextcloud/dialogs'
+import '@nextcloud/dialogs/styles/toast.scss'
 
 export default {
 	name: 'Calendar',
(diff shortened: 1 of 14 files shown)


src/views/Calendar.vue
```diff
@@ -131,6 +131,10 @@ import eventLimitText from '../fullcalendar/eventLimitText.js'
 import windowResize from '../fullcalendar/windowResize.js'
 import dayRender from '../fullcalendar/dayRender.js'
 import { loadState } from '@nextcloud/initial-state'
+import {
+	showWarning,
+} from '@nextcloud/dialogs'
+import '@nextcloud/dialogs/styles/toast.scss'
```
tcitworld, line 137: Why does this one needs to import styles as well?

georgehrke: According to https://github.com/nextcloud/nextcloud-dialogs you have to import the styles and Calendar.vue is the root component, so it will always load the styles here.

All the other components are children of Calendar.vue

PR #2263, opened 2020-05-11
Prefill title and description of calendar event from url

nenadalm:
## Description

It's now possible to create new calendar with prefilled title and description. Google calendar has same feature: https://stackoverflow.com/questions/22757908/what-parameters-are-required-to-create-an-add-to-google-calendar-link

Fixes # (issue)

### Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

1. Open calendar with `title` and/or `description`  and/or `alarms` query params (e.g. 
`index.php/apps/calendar/dayGridMonth/now/new/sidebar/0/1620943200/1621029600?title=tit&description=desc&alarms=%5B%7B"type"%3A"EMAIL","totalSeconds"%3A300%7D%5D`)
2. Observe prefilled title and description inputs from url

### Checklist:


src/store/calendarObjectInstance.js
```diff
@@ -1424,7 +1438,12 @@ const actions = {
 		const calendarObject = await dispatch('createNewEvent', { start, end, isAllDay, timezoneId })
 		const startDate = new Date(start * 1000)
 		const eventComponent = getObjectAtRecurrenceId(calendarObject, startDate)
-		const calendarObjectInstance = mapEventComponentToEventObject(eventComponent)
+		const eventProps = objectFromEntries(Object.entries({ title, description }).filter(([k, v]) => v != null))
+		const calendarObjectInstance = Object.assign(
```
georgehrke, line 1442: I didn't yet have a chance to test this, but I doubt this will work.

You are adding the properties to the `calendarObjectInstance`, but not to the actual `eventComponent`. So unless you are editing the title, no title will be set when hitting the save button.

src/store/calendarObjectInstance.js
```diff
@@ -46,6 +46,20 @@ import {
...
 import { getObjectAtRecurrenceId } from '../utils/calendarObject.js'
 
+/**
+ * @param {Array} entries Object entries
+ *
+ * @returns {Object}
+ */
+function objectFromEntries(entries) {
+	const obj = {}
+	entries.forEach(([k, v]) => {
```
georgehrke, line 56: Please use full readable variable names. We have a minifier in place for the production code. The source version should always use readable names.

src/store/calendarObjectInstance.js
```diff
@@ -1413,7 +1427,7 @@ const actions = {
 	 * @param {String} data.timezoneId The timezoneId of the new event
 	 * @returns {Promise<{calendarObject: Object, calendarObjectInstance: Object}>}
 	 */
-	async getCalendarObjectInstanceForNewEvent({ state, dispatch, commit }, { isAllDay, start, end, timezoneId }) {
+	async getCalendarObjectInstanceForNewEvent({ state, dispatch, commit }, { isAllDay, start, end, timezoneId, title, description }) {
```
georgehrke, line 1430: Please add the `title` and `description` to the jsDoc block above.

src/store/calendarObjectInstance.js
```diff
@@ -1424,7 +1438,12 @@ const actions = {
 		const calendarObject = await dispatch('createNewEvent', { start, end, isAllDay, timezoneId })
 		const startDate = new Date(start * 1000)
 		const eventComponent = getObjectAtRecurrenceId(calendarObject, startDate)
-		const calendarObjectInstance = mapEventComponentToEventObject(eventComponent)
+		const eventProps = objectFromEntries(Object.entries({ title, description }).filter(([k, v]) => v != null))
```
georgehrke, line 1441:
```suggestion
		if (title) {
			eventComponent.title = title
		}
		if (description) {
			eventComponent.description = description
		}
		const calendarObjectInstance = mapEventComponentToEventObject(eventComponent)
```

If you add the properties to `eventComponent` like shown above, `mapEventComponentToEventObject` will automatically include them in `calendarObjectInstance`.

georgehrke (changes requested):

src/store/calendarObjectInstance.js
nenadalm: Thanks. Fixed. I was assuming that if there is something in input and I hit save, it's gonna be saved - didn't test it.

src/store/calendarObjectInstance.js
nenadalm: I removed the function as it's not necessary anymore.

I know you have minifier and I didn't use k, v to have final js small :) I'm just used to using `k` as key and `v` as value in generic functions instead of `key` and `value`.

src/store/calendarObjectInstance.js
nenadalm: Fixed. thanks. I just used `title != null` comparison instead if that's ok as it's probably more clear.

It doesn't matter in js for this case as the query string is string: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

But e.g. php (also used in this projects) considers `"0"` to be falsy: https://www.php.net/manual/en/language.types.boolean.php and I find it hard to remember each languages boolean conversions.

nenadalm:
@georgehrke I wonder what's the state of this pr? Anything I can do?

ChristophWurst:
@nenadalm what would be the use case for this enhancement? Do you have an example?

tcitworld:
For instance having a link from other apps allowing to open the event editor prefilled with some details.

PR #2278, opened 2020-05-22
Add unit-tests for certain stores.

georgehrke:
tl;dr:
- Add tests for the following stores:
  - ContactsStore
  - DavRestrictionStore
  - ImportFilesStore
  - ImportStateStore
  - SettingsStore
- `import logger ...` => `import { logError, logInfo } from logger`
- `import client from caldavService` => `import { createCalendarCollection } from caldavService`
- Lots of fixes for jsdoc
- Replace more usage of static strings with constants


georgehrke:
@tcitworld This is ready to review :)

PR #2284, opened 2020-05-25
Move calendar-picker above time-picker

### LINKED ISSUE #2007: UI Tweak: "Calendar" setting in Event Editor not prominent enough
This is a "bug" report on a possible design tweak concerning the event editor. In daily use I find it very inconvenient that the target calendar has been "downgraded" to a setting in the first tab of the event editor side panel. I see this as one of the most important settings, maybe even more important than the date and time.

It happened several times to me and other users that their new calendar entries accidentally ended up in the wrong (potentially a publicly shared!) calendar just because the item is not visible enough.

In the old pre-v2.0 interface, the calendar was one of the most prominent items of the edit panel.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/88949913-ui-tweak-calendar-setting-in-event-editor-not-prominent-enough?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45525646&utm_medium=issues&utm_source=github).
</bountysource-plugin>

georgehrke:
fixes #2007

| Before | After |
|:---------:|:------:|
|popover_before|popover_after|
|sidebar_before|sidebar_after|

### DIFF
--- a/css/app-sidebar.scss
+++ b/css/app-sidebar.scss
@@ -30,6 +30,11 @@
 	.app-sidebar-header__action {
 		margin-top: 0 !important;
 		max-height: none !important;
+		flex-wrap: wrap;
+
+		div {
+			flex-shrink: 0;
+		}
 	}
 
 	.editor-invitee-list-empty-message,
--- a/src/components/Editor/Properties/PropertyCalendarPicker.vue
+++ b/src/components/Editor/Properties/PropertyCalendarPicker.vue
@@ -23,10 +23,6 @@
 <template>
 	<div v-if="display" class="property-select">
 		<div
-			class="property-select__icon icon-calendar-dark"
-			:title="$t('calendar', 'Calendar')" />
-
-		<div
 			class="property-select__input"
 			:class="{ 'property-select__input--readonly-calendar-picker': isReadOnly }">
 			<CalendarPicker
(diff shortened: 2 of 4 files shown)


raimund-schluessler:
How about keeping the icon in front of the calendar picker? (And maybe add an icon in front of start and end date for consistency in a later PR?)

georgehrke:
> (And maybe add an icon in front of start and end date for consistency in a later PR?)

I'm not sure about the icons in front of the start and end date field. If you are on a bigger screen, it will automatically move start and end into one line to save space:
AA8483ED-7943-4CAC-84FB-2FCEDA6263B3

If we add icons, we will have to set the breakpoint for the wrap even higher.

georgehrke:
> Also, what’s the whitespace on the right of the event title input field? Seems that at least in the normal case it doesn’t need to be there.

That's because it's the new event dialog and there are no actions.

We want to show a simple trash bin icon to close it as per #1576

jancborchardt:
> We want to show a simple trash bin icon to close it as per #1576

Yep, but for new events with nothing in it yet (and no actions), the title should expand full width. :)

Issue #2304, opened 2020-06-02
New Setting panel


tcitworld:
Bottom left is getting too crowded.

Needs to handle 
- Existing settings
- Defining a default calendar for creating new events https://github.com/nextcloud/calendar/issues/1835
- Free/Busy view https://github.com/nextcloud/calendar/issues/39
- Customize reminder type for automatically generated birthday events #130?

georgehrke:
There is a WIP PR by Marco in the nc/vue repo: https://github.com/nextcloud/nextcloud-vue/pull/1286

szaimen:
Lets continue in https://github.com/nextcloud/calendar/issues/5892 :)

PR #2331, opened 2020-06-13
WIP: Setting for default calendar when creating new events

### LINKED ISSUE #1835: Defining a default calendar
## Feature Request

As a Calendar user I have more than one calendar. When another users invites me to an event, it is not clear to which calendar the event is copied.

It would be really nice if I could define a default calendar for creating new events, since I am using most of the time a calendar, which is since the 2.0.0 release no longer on top. So as long as you cannot change the list order, I think this would be a nice addition. Thanks!

## Implementation

This option could be added to the settings&import section.

e.g. this is in Nextcloud Tasks already possible. Maybe you could copy some code from there?
image

LukasBoersma:
Signed-off-by: Lukas Boersma <mail@lukas-boersma.com>

## Description

This adds a setting for the default calendar when creating a new event.

To summarize my changes:

- I added a 'disabled' property to the `CalendarPicker` component so that the dropdown for the new setting can be disabled while the setting is saved.
- I added a setting for the default calendar in the settings panel and all the other places where settings need to be handled (I just followed the usage of the existing settings here).
- In the functions `getCalendarObjectInstanceForNewEvent` (`calendarObjectInstance.js`) and `createNewEvent` (`calendarObjects.js`) I added `calendarId` as an new (optional) parameter to specify which calendar to use for a new event.
- In the event editor, I use that parameter to use the configured default calendar.
- When no default calendar is specified yet, the behaviour is the same as before (i.e. the first calendar in `sortedCalendars` is used).

**This is work in progress.** I still have an issue with the calendar picker, it does only close when double clicking it, any help here is greatly appreciated. Also I have yet to write tests for this new setting. However it would be cool if **somebody from the maintainers could let me know if this change has a chance of being merged**.

Fixes #1835

### Type of change

- [X] New feature (non-breaking change which adds functionality)

1. Have more than one calendar in the calendar app.
2. Change the default calendar in the settings panel.
3. Create a new event. The previously selected default calendar will now be preselected for the new event.

### UI Changes

| Before | After |
|:---------:|:------:|
|defaultcalendar-before|defaultcalendar-after|

### Checklist:

- [X] My code follows the style guidelines of this project
- [X] I signed off my changes (`git commit -sm "Your commit message"`)
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/css/app-settings.scss
+++ b/css/app-settings.scss
@@ -55,7 +55,8 @@
 		}
 
 		&--slotDuration,
-		&--timezone {
+		&--timezone,
+		&--default-calendar {
 			width: 100%;
 
 			.multiselect {
--- a/lib/Controller/PublicViewController.php
+++ b/lib/Controller/PublicViewController.php
@@ -120,6 +120,7 @@ class PublicViewController extends Controller {
 		$defaultSkipPopover = $this->config->getAppValue($this->appName, 'skipPopover', 'yes');
 		$defaultTimezone = $this->config->getAppValue($this->appName, 'timezone', 'automatic');
 		$defaultSlotDuration = $this->config->getAppValue($this->appName, 'slotDuration', '00:30:00');
+		$defaultCalendarId = $this->config->getAppValue($this->appName, 'defaultCalendarId', 'none');
 		$defaultShowTasks = $this->config->getAppValue($this->appName, 'showTasks', 'yes');
 
 		$appVersion = $this->config->getAppValue($this->appName, 'installed_version');
@@ -134,6 +135,7 @@ class PublicViewController extends Controller {
 		$this->initialStateService->provideInitialState($this->appName, 'talk_enabled', false);
 		$this->initialStateService->provideInitialState($this->appName, 'timezone', $defaultTimezone);
 		$this->initialStateService->provideInitialState($this->appName, 'slot_duration', $defaultSlotDuration);
+		$this->initialStateService->provideInitialState($this->appName, 'default_calendar_id', $defaultCalendarId);
 		$this->initialStateService->provideInitialState($this->appName, 'show_tasks', $defaultShowTasks === 'yes');
 		$this->initialStateService->provideInitialState($this->appName, 'tasks_enabled', false);
 
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -87,6 +87,8 @@ class SettingsController extends Controller {
 				return $this->setEventLimit($value);
 			case 'slotDuration':
 				return $this->setSlotDuration($value);
+			case 'defaultCalendarId':
+				return $this->setDefaultCalendarId($value);
 			case 'showTasks':
 				return $this->setShowTasks($value);
 			default:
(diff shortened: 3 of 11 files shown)


tcitworld:
Looks good!

I'd also like to have this with https://github.com/nextcloud/calendar/issues/2304 instead of adding another setting item in our crowded setting menu.

LukasBoersma:
@georgehrke Thanks for the feedback! I finally found some time to look at this again.

I managed to get the setting for the initial state from the dav app by loading it along with the calendar's own settings in the calendar's ViewControllers, but I can't figure out how to store the setting in the dav app the right way.

I expect that there is already an API that I can call to store the value, but I can't find it. Or should I just add a function to the calendar's `SettingsController.php` that stores the value for the dav app?

georgehrke:
> I expect that there is already an API that I can call to store the value, but I can't find it. Or should I just add a function to the calendar's SettingsController.php that stores the value for the dav app?

@tcitworld You were working on this, right? Is there already an API?
Otherwise it's fine to use:
```
			$this->config->setUserValue(
 				$this->userId,
 				'dav',
 				'defaultCalendarId',
 				$value
 			);
```

Bolli84:
With more and more collaboration in Nextcloud HUB, this is more and more improtant.

tcitworld:
@ChristophWurst Since we'll have a groupware user settings section with https://github.com/nextcloud/server/pull/27466 wouldn't it make sense to have this option in Server Settings instead of the Calendar app?

szaimen:
> @ChristophWurst Since we'll have a groupware user settings section with [nextcloud/server#27466](https://github.com/nextcloud/server/pull/27466) wouldn't it make sense to have this option in Server Settings instead of the Calendar app?

I'd honestly vote against it because the setting would be very hard to discover this way. And this setting actually only correspends to the calendar so it should be in the calendar app, imo.

ChristophWurst:
I agree with @szaimen. The availability is relevant for *any* CalDAV client. This is only relevant for this app, right? It should be here.

It will help to have a settings modal in this app like we have in Mail and Talk. The current bottom left settings area is already quite crammed.

tcitworld:
> This is only relevant for this app, right?

Nope, it's a dav setting. It should use the setting added in https://github.com/nextcloud/server/pull/19852 instead of the custom one added here, like Georg said above.

szaimen:
Though I still think that having the setting in the calendar app is better because of better discoverability, no?

tcitworld:
I have no strong opinions on this one, other than the fact that it *would be better* for the setting to be available without the app installed, but who doesn't install the calendar app anyway? :sunglasses:

szaimen:
> t _would be better_ for the setting to be available without the app installed

You are right, makes sense because of that. But in this case I am definitely voting for adding a link in the calendar settings that links directly to the corresponding personal setting section like I've done it here: https://github.com/nextcloud/activity/pull/621

st3iny:
We already have a link to the personal groupware settings. It might has to be renamed for this feature.

image

st3iny:
If we implement the setting globally (Personal Setting -> Groupware) the picker (+ its backend) has to move into the dav app at `apps/dav/src/views`.

PR #2286, opened 2020-05-26
Fix Datepicker localization

georgehrke:
fixes #2174

fixes first day of week
fixes title of month in header

| Before | After |
|:---------:|:------:|
|datepicker_before|6C199988-3F02-4778-BEB6-A6E8C03DC456|

PR #2414, opened 2020-07-22
Dashboard widget

juliushaertl:
| No events | No events today | Events | List with ToDo |
|:---------:|:------:|:-------:|:--------:|
|F1BAAB3B-17E6-4D4E-99D1-D725AEB2E2C0|5E4359A5-5180-40B9-AE19-5B79B8085847|95FB5E86-F5E9-4F43-8F16-6F1EB67786D0|95C015E8-7EA2-413E-B9B3-C0BD94CB5A14|

## Description

This PR implements a basic dashboard widget for Nextcloud 20.

### Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I signed off my changes (`git commit -sm "Your commit message"`)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

### ToDo

- [x] Check for backward compatibility on older Nextcloud versions
- [x] Fetch a list of upcoming events from the server
- [x] Determine which information we should show per event

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -23,19 +23,39 @@ declare(strict_types=1);
  */
 namespace OCA\Calendar\AppInfo;
 
+use OCA\Calendar\Dashboard\CalendarWidget;
 use OCP\AppFramework\App;
+use OCP\AppFramework\Bootstrap\IBootContext;
+use OCP\AppFramework\Bootstrap\IBootstrap;
+use OCP\AppFramework\Bootstrap\IRegistrationContext;
 
 /**
  * Class Application
  *
  * @package OCA\Calendar\AppInfo
  */
-class Application extends App {
+class Application extends App implements IBootstrap {
+
+	/** @var string */
+	public const APP_ID = 'calendar';
 
 	/**
 	 * @param array $params
 	 */
 	public function __construct(array $params=[]) {
-		parent::__construct('calendar', $params);
+		parent::__construct(self::APP_ID, $params);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function register(IRegistrationContext $context): void {
+		$context->registerDashboardWidget(CalendarWidget::class);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function boot(IBootContext $context): void {
 	}
 }
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -1,0 +1,312 @@
+<!--
+	- @copyright Copyright (c) 2020 Julius Härtl <jus@bitgrid.net>
+	-
+	- @author Julius Härtl <jus@bitgrid.net>
+	-
+	- @license GNU AGPL version 3 or any later version
+	-
+	- This program is free software: you can redistribute it and/or modify
+	- it under the terms of the GNU Affero General Public License as
+	- published by the Free Software Foundation, either version 3 of the
+	- License, or (at your option) any later version.
+	-
+	- This program is distributed in the hope that it will be useful,
+	- but WITHOUT ANY WARRANTY; without even the implied warranty of
+	- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	- GNU Affero General Public License for more details.
+	-
+	- You should have received a copy of the GNU Affero General Public License
+	- along with this program. If not, see <http://www.gnu.org/licenses/>.
+	-
+	-->
+
+<template>
+	<DashboardWidget
+		id="calendar_panel"
+		:items="items"
+		:loading="loading">
+		<template #default="{ item }">
+			<EmptyContent v-if="item.isEmptyItem"
+				id="calendar-widget-empty-content"
+				class="half-screen"
+				icon="icon-checkmark">
+				<template #desc>
+					{{ t('calendar', 'No more events today') }}
+				</template>
+			</EmptyContent>
+			<DashboardWidgetItem v-else :item="item">
+				<template #avatar>
+					<div
+						v-if="item.componentName === 'VEVENT'"
+						class="calendar-dot"
+						:style="{'background-color': item.calendarColor}"
+						:title="item.calendarDisplayName" />
+					<div v-else
+						class="vtodo-checkbox"
+						:style="{'color': item.calendarColor}"
+						:title="item.calendarDisplayName" />
+				</template>
+			</DashboardWidgetItem>
+		</template>
+		<template #empty-content>
+			<EmptyContent
+				id="calendar-widget-empty-content"
+				icon="icon-calendar-dark">
+				<template #desc>
+					<p class="empty-label">
+						{{ t('calendar', 'No upcoming events') }}
+					</p>
+					<p>
+						<a
+							class="button"
+							:href="clickStartNew">
+							{{ t('calendar', 'Create a new event') }}
+						</a>
+					</p>
+				</template>
+			</EmptyContent>
+		</template>
+	</DashboardWidget>
+</template>
+
+<script>
+import { DashboardWidget, DashboardWidgetItem } from '@nextcloud/vue-dashboard'
+import EmptyContent from '@nextcloud/vue/dist/Components/EmptyContent'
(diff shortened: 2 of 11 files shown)


juliushaertl:
@georgehrke Is there any way we can query the DAV endpoint for something like "the next 10 upcoming events"? Otherwise we probably should have a new HTTP endpoint for that where we collect the relevant list. This would have the benefit that we could also provide the list through the initial state api and reduce loading time when the user sees no information being available in the frontend.

georgehrke:
No, there is no such endpoint. The problem here are recurring events. There is no way for us to easily query "the next 10 events" or something similar. (Neither via CalDAV nor from the database for that matter)

If you send a calendar-query with a time-range, it will return:
- all non-recurring events in that time-range
- all recurring events:
   - where the first occurrence is earlier than the end-date
   - and where the last occurrence is later than the start-date

The server doesn't actually check whether the recurring event is occurring in that time-range. For example, if you query all events for June, the server will just return all birthday events, even if the person was not born in June.

georgehrke:
As discussed:
- query events for the next two weeks
- Out of the queried events, show the next 7
- If there are no events, show: `No events until at least next week! Open Calendar for more`

lib/AppInfo/Application.php
```diff
@@ -37,5 +40,14 @@ class Application extends App {
 	 */
 	public function __construct(array $params=[]) {
 		parent::__construct('calendar', $params);
+
+		$container = $this->getContainer();
+
+		// TODO: Migrate to new bootstrap once Calendar supports only Nextcloud 20+
```
georgehrke, line 46: We can just bump to 2.1 and limit it to Nc 20+

juliushaertl: Sounds good to me :+1:

georgehrke:
FC94FEC9-188B-4091-B210-2822AD5EDC48

georgehrke:
@jancborchardt I also added a small colored circle with the calendar color.

src/views/Dashboard.vue
```diff
@@ -0,0 +1,249 @@
...
+				timezone,
+			} = loadState('calendar', 'dashboard_data')
+			this.$store.commit('loadSettingsFromServer', {
+				timezone,
+				showTasks,
+			})
+
+			const start = dateFactory()
+			const end = dateFactory()
+			end.setDate(end.getDate() + 14)
```
georgehrke, line 108: @jancborchardt Should it be 14 days or end of next week?

src/views/Dashboard.vue
```diff
@@ -0,0 +1,249 @@
...
+				.sort((a, b) => a.start.getTime() - b.start.getTime())
+				.filter(event => !event.classNames.includes('fc-event-nc-task-completed'))
+				.slice(0, 7)
+				.map((event) => {
+					const calendar = calendars.find((calendar) => calendar.id === event.extendedProps.calendarId)
+				  let startDate
+					if (event.allDay) {
+					  startDate = moment(event.start).locale(locale).calendar(null, {
+					    // TRANSLATORS Please translate only the text in brackets and keep the brackets!
+							sameDay: t('calendar', '[Today]'),
```
georgehrke, line 152: TODO: fix formatting.

src/views/Dashboard.vue
jancborchardt: 14 days is good, that way there’s always 2 weeks of relevant events. :)

jancborchardt:
> @jancborchardt I also added a small colored circle with the calendar color.

Very cool! I’d say the subline should be indented a bit to be aligned with the text of the main line then.

What do you think about indenting the whole block further so that the text of the content aligns with the heading text of "Upcoming events"?

------------

And an additional thing: If there are no events left for today, can we have an entry saying "All events done for today!" – it could even take 2 rows so it’s clear that today is done. Otherwise you have to check the subline to make sure you got it all.

src/views/Dashboard.vue
```diff
@@ -0,0 +1,249 @@
...
+		<template #desc>
+			<p />
+			<button>{{ t('calendar', 'Create a new event') }}</button>
+		</template>
+	</EmptyContent>
+	<div v-else>
+		<ul>
+			<li v-for="event in events" :key="event.id">
+				<a :href="event.deepLink" class="event">
+					<div class="event__title">
```
jancborchardt, line 45: @georgehrke if you’re not using the component, this should be a h3 as well like in the component. :) Then you also don’t need to use the additional `strong` tags for the title.

georgehrke: There is a component for it? :) 

I just added the "fetching events part" to Julius' PR.

jancborchardt: Yep, the component is at https://github.com/nextcloud/nextcloud-vue-dashboard, the other widgets use it already.
We will only have to decide how we display the calendar color indicator (small circle, big circle, etc) and possibly adjust the component.

Also since you use custom structure, please make sure it’s keyboard-focusable (via Tab and Shift-Tab) and then accessible via pressing enter.

georgehrke: Let me migrate this code to https://github.com/nextcloud/nextcloud-vue-dashboard in that case :)

juliushaertl:
@georgehrke There seems to be an issue with having only the default personal calendar, in that case there are no requests sent and no events shown. When adding a second calendar only the events of that are shown.

georgehrke:
@juliushaertl Will check that. Was that calendar enabled?

jancborchardt:
> And an additional thing: If there are no events left for today, can we have an entry saying "All events done for today!" – it could even take 2 rows so it’s clear that today is done. Otherwise you have to check the subline to make sure you got it all.

@georgehrke good example of what I meant by this is in the Talk widget: https://github.com/nextcloud/spreed/pull/3890#issuecomment-674903784

georgehrke:
@jancborchardt Please see the updated screenshots in the original post

lib/AppInfo/Application.php
georgehrke: @tcitworld Do you have any objections here?
I would release this as part of Calendar 2.1 and we can always backport bugfixes to 2.0.x when necessary :)

tcitworld: Suits me. There's at least #2286 and #2275 that would be nice to have in a close 2.0.x release (preferably released before 20 gets out with 2.1).

jancborchardt (changes requested):
> @jancborchardt Please see the updated screenshots in the original post

Super super super nice! :) Getting 2 issues though:
- [ ] The "Create a new event" button seems to not do anything?
- [ ] Clicking on an event initially tries to open a deeplink, but then changes to open just the Dashboard. I do have the index.php in my url path, if that makes a difference.
- [ ] The "No more events today" doesn’t seem to take exactly the space as 2 entries – the calendar entries below don’t line up vertically with those of other widgets.

georgehrke:
I would like to get #2498 in first, then we can use the icon-font i added for tasks:
2B83E566-3290-490E-A7E5-446A21FCA998

src/views/Dashboard.vue
```diff
@@ -0,0 +1,274 @@
...
+		 * @returns {Object[]}
+		 */
+		formatEvents(expandedEvents) {
+			return expandedEvents
+				.sort((a, b) => a.start.getTime() - b.start.getTime())
+				.filter(event => !event.classNames.includes('fc-event-nc-task-completed'))
+				.slice(0, 7)
+				.map((event) => ({
+					isEmptyItem: false,
+					targetUrl: generateUrl('apps/calendar') + '/edit/' + event.extendedProps.objectId + '/' + event.extendedProps.recurrenceId,
```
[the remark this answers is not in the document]
georgehrke: Done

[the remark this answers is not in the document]
georgehrke: Done

georgehrke (approved):
Good to merge from my side 👍

georgehrke:
All remarks have been addressed.
Let's get this in, we can always fix smaller issues later.

PR #2450, opened 2020-08-04
Add direct editing route to routes.php

georgehrke:


tcitworld:
What's the benefit over the `calendar.view.indexview.timerange.edit` route (which I previously used in https://github.com/nextcloud/server/pull/19039)?

georgehrke:
One benefit is, that this route will automatically use the user-preferred view.
Later on, my plan is to automatically navigate to the correct date in the grid view.

PR #2478, opened 2020-08-20
Let apps handle clicks on todo entries

juliushaertl:
## Description

Open task entries that are exposed by deck inside of the deck app by emitting an event that can be handled by apps

### Type of change

Requires https://github.com/nextcloud/deck/pull/2683

### Checklist:

- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2173,9 +2173,9 @@
 			"dev": true
 		},
 		"@types/semver": {
-			"version": "7.3.3",
-			"resolved": "https://registry.npmjs.org/@types/semver/-/semver-7.3.3.tgz",
-			"integrity": "sha512-jQxClWFzv9IXdLdhSaTf16XI3NYe6zrEbckSpb5xhKfPbWgIyAY0AFyWWWfaiDcBuj3UHmMkCIwSRqpKMTZL2Q=="
+			"version": "7.3.4",
+			"resolved": "https://registry.npmjs.org/@types/semver/-/semver-7.3.4.tgz",
+			"integrity": "sha512-+nVsLKlcUCeMzD2ufHEYuJ9a2ovstb6Dp52A5VsoKxDXgvE051XgHI/33I1EymwkRGQkwnA0LkhnUzituGs4EQ=="
 		},
 		"@types/stack-utils": {
 			"version": "1.0.1",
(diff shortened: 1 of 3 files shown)


src/fullcalendar/eventClick.js
```diff
@@ -94,17 +94,26 @@ function handleEventClick(event, store, router, route, window) {
...
 	const taskId = davUrlParts.pop()
 	const calendarId = davUrlParts.pop()
-	const url = `apps/tasks/#/calendars/${calendarId}/tasks/${taskId}`
 
+	const deckAppPrefix = 'app-generated--deck--board-'
+	let url
+	if (calendarId.startsWith(deckAppPrefix)) {
+		const board = calendarId.substr(deckAppPrefix.length)
+		const card = taskId.substr('card-'.length).replace('.ics', '')
+		url = `apps/deck/#/board/${board}/card/${card}`
```
juliushaertl, line 111: @georgehrke Would you be fine with such an app related approach or should we just have a route at the deck app that takes the dav calendarId and taskId so that no parsing is done in calendar?

georgehrke: Should we just use nc/event-bus here? Then any app can register custom handlers and calendar doesn't need to know anything about how their routers work.

juliushaertl: Sounds good to me.

ChristophWurst (approved):

raimund-schluessler:
I have a problem understanding how that would work. When you are in the Calendar app, the Deck app isn't loaded, no? How could the Deck app then react to a click on a todo (via the event-bus)?

juliushaertl:
> I have a problem understanding how that would work. When you are in the Calendar app, the Deck app isn't loaded, no? How could the Deck app then react to a click on a todo (via the event-bus)?

See https://github.com/nextcloud/deck/pull/2683/files for the deck implementation. We basically register a small helper script that gets loaded and handles the event.

tcitworld:
Instead of asking admins to install the tasks app, maybe we could fallback to link to the task in the calendar app?

juliushaertl:
> Instead of asking admins to install the tasks app, maybe we could fallback to link to the task in the calendar app?

Would be fine by me, but I'd say that would be separate from that PR.

PR #2495, opened 2020-08-22
Use standardized EmptyContent component

georgehrke:


ChristophWurst (approved):
Looks good!

ChristophWurst:
@tcitworld how many reviews do you want? :)

tcitworld:
I had the review button opened but something else caught by attention, and then you were too quick. :)

PR #2520, opened 2020-08-28
Remove highlighting of weekends

georgehrke:
| Before | After |
|:---------:|:------:|
|before|after|

At the last Calendar Design review session, we discussed whether we should remove the highlighting for weekends. (similar to Google Calendar and Outlook web).
The highlighting draws a lot of attention to weekends, although they are often not as important and always at the the same position anyway.


SuperSandro2000:
I like the overlay. It makes it really easy for me to know when it is weekend even though it is always on the right,

skjnldsv (approved):

ma12-co (approved):
Looks very nice :) Good stuff!

tcitworld (approved):
Fine by me.

PR #2521, opened 2020-08-28
Reduce brightness of events

georgehrke:
| Before | After |
|:---------:|:------:|
|before|after|

At the last Calendar Design review session, we discussed whether we should reduce the brightness of events in the calendar-grid.

| Before | After |
|:---------:|:------:|
|10%|after-0-1|
|20%|after-0-2|
|30%|after-0-3|

### DIFF
--- a/src/fullcalendar/eventSources/eventSource.js
+++ b/src/fullcalendar/eventSources/eventSource.js
@@ -20,7 +20,7 @@
  *
  */
 import {
-	generateTextColorForHex,
+	generateAlphaColor,
 } from '../../utils/color.js'
 import getTimezoneManager from '../../services/timezoneDataProviderService'
 import { getUnixTimestampFromDate } from '../../utils/date.js'
@@ -38,9 +38,9 @@ export default function(store) {
 		const source = {
 			id: calendar.id,
 			// coloring
-			backgroundColor: calendar.color,
+			backgroundColor: generateAlphaColor(calendar.color, 0.1),
 			borderColor: calendar.color,
-			textColor: generateTextColorForHex(calendar.color),
+			textColor: calendar.color,
 			// html foo
 			events: async({ start, end, timeZone }, successCallback, failureCallback) => {
 				let timezoneObject = getTimezoneManager().getTimezoneForId(timeZone)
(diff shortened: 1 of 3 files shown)


georgehrke:
Maybe 10% alpha is too low.

SuperSandro2000:
I am not a designer by any means but the news designs makes the colors really hard to guess.

georgehrke:
I added some screenshots of different alpha values to the original post

nrgeen:
30% looks very good for me

skjnldsv:
Maybe check against WCAG ?

ma12-co (commented):
@georgehrke, could we see a screenshot with opacity 30% and bold event titles?

tcitworld:
I need to have look at how it looks in dark mode, but "outlined" events are quite nice around 30% opacity.

georgehrke:
@ma12-co @tcitworld Here you go: 0.3 with bold font in light and dark:

| Light | Dark|
|:---------:|:------:|
|month_light|month_dark|

ma12-co:
Looks very nice to me :) 
One more thing, what's the margin between events? 2px? seems a little tight to me

georgehrke:
> Maybe check against WCAG ?

Only for 0.1 I'm getting a contract ratio of 4.74, which is good enough to pass `AA`, but not `AAA`.

Setting the alpha value of the background to 1.5, 2, 2.5, or 3 gives a contrast ratio of less than 4.5

ma12-co:
I guess we have to use 0.1 then. Another thing we could do would be darkening the events colors, that should also help.

jancborchardt:
My worry about using such low opacity is like @SuperSandro2000 said:

> the new designs make the colors really hard to guess for me.

And the borders on all the events make it look a bit much to take in, lots of boxes and lines. 2 big approaches are:
- Full color boxes like Google Calendar
- Light boxes without border, like macOS Calendar

ChristophWurst:
Is there anything left to do? Otherwise I can try to rebase, give a smoke test and integrate :)

szaimen:
not needed anymore

PR #2522, opened 2020-08-28
Change highlighting of today

georgehrke:
| Before | After |
|:---------:|:------:|
|before|after|
|Week view: |week_after|

I'm not happy about how it looks in day / week view.

At the last Calendar Design review session, we discussed how to improve the highlighting of today. This is what we came up with. It's similar to what Outlook Web does.

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -71,27 +71,10 @@
 }
 
 // Today highlighting
-.fc-day-today {
-	&.fc-col-header-cell {
-		a, span {
-			background-color: var(--color-primary);
-			color: var(--color-primary-text) !important;
-			border-radius: var(--border-radius-pill);
-			padding: 2px 6px;
-			font-weight: bold;
-		}
-	}
-
-	.fc-daygrid-day-number {
-		background: var(--color-primary);
-		color: var(--color-primary-text);
-		border-radius: 50%;
-		margin: 3px 3px 0 0;
-		width: 24px;
-		height: 24px;
-		text-align: center;
-		font-weight: bold !important;
-		padding: 1px 0 !important;
+.fc-view {
+	&.fc-daygrid .fc-day.fc-daygrid-day.fc-day-today,
+	&.fc-timegrid .fc-day.fc-col-header-cell.fc-day-today {
+		box-shadow: inset 0 4px var(--color-primary);
 	}
 }
 


SuperSandro2000:
I would make the entire top row blue and wouldn't move the number a pixel down.

skjnldsv (approved):

juliushaertl:
Any reason against a full cell highlight with a dimmed primary color? I must say i find both the before and after very hard to spot.

georgehrke:
> Any reason against a full cell highlight with a dimmed primary color? I must say i find both the before and after very hard to spot.

We used to have a full-cell yellow highlighting in the past, but that caused many issues because the visibility was very dependent on the display and screen settings. We got very mixed bug reports from some users for whom it was way too bright and then from others who couldn't even see it on their screens. That's why we stopped using the full cell highlight long time ago.

georgehrke:
For reference: iCloud does it similar to this PR, just that they also highlight the entire current week:
97417048-6636-47E5-86CB-C5B9639D5657

ma12-co (changes requested):
Can we extend the highlight for the whole current week just like apple calendar does in the month view? It just makes it way faster to reach the current day at a glance.
it could be done with var(--color-priomary-light) maybe?

tcitworld:
Also in week view iCloud puts a light background on the whole day column.
image

jancborchardt:
> Any reason against a full cell highlight with a dimmed primary color? I must say i find both the before and after very hard to spot.

Would agree here. The previous highlight of the very present circle on the day number is definitely better (because it also makes clearer to which day it belongs and can’t be confused with the day above it), and an additional background of `var(--color-primary-light)` could do the trick.

th23x:
Great idea as the "Today" does not come out clearly initially.

I also had a look and styled this via some CSS - unfortunately not in the source code format, but only via theming. However maybe somebody would like to take it to develop further/ include in the suggested chage:

Result views:
cal-month
cal-week
cal-list

Source CSS (month):
`.app-calendar .fc .fc-daygrid-day.fc-day-today {
	background: var(--color-primary-light);
}
.app-calendar .fc-day-today .fc-daygrid-day-number {
	border-radius: 0;
	margin: 0 0 6px 0;
	width: 100%;
	height: auto;
}`

Source CSS (week):
`.app-calendar .fc-day-today.fc-col-header-cell a,
.app-calendar .fc-day-today.fc-col-header-cell span {
	width: 100%;
	border-radius: 0;
}
.app-calendar .fc .fc-timegrid-col.fc-day-today {
	background: #fafafa;
}`

Source CSS (list):
`.app-calendar .fc-list-day.fc-day-today .fc-list-day-cushion {
	background: var(--color-primary-element);
}
.app-calendar .fc-list-day.fc-day-today .fc-list-day-cushion .fc-list-day-text,
.app-calendar .fc-list-day.fc-day-today .fc-list-day-cushion .fc-list-day-side-text {
	color: var(--color-primary-text);
	font-weight: bold;
}`

alexanderdd:
Hey, to me the new highlighting does not stick out enough visually. It looks exactly the same as events (rectangle). Please **make a screenshot with 6+ events per day, you will notice that it does not work**. (I notice a tendency in the calendar repo to use empty or nearly empty calendars for testing screenshots -- my calendars are full, and that is when I need highlighting the most). Also, just because Outlook an iCloud are doing it, we don't have to do it the same way.

The circle is better. OR "full cell highlight with a dimmed primary color" as suggested by @juliushaertl (businesscalendar2 android app does that).

szaimen:
not needed anymore

PR #2528, opened 2020-08-31
Improve loading routine of editors

### LINKED ISSUE #2187: Long description box on 2.0.3
### Steps to reproduce
1. Go to month view
2. Uncheck "Enable simplified editor"
3. Create an event that has a description and close it
4. Click on an event with a description

### Expected behaviour
The description box in the editor should be the same size as the event.

### Actual behaviour
The description box is crazy long

### Calendar app
**Calendar app version:** 2.0.3

**CalDAV-clients used:** NA

### Client configuration
**Browser:** Firefox 75.0

**Operating system:** Windows 10 1909

### Server configuration

**Nextcloud Version:** 18.0.3

I did a quick screen recording to demonstrate what I'm seeing because it's kind of hard to describe. Thanks, as always for the great calendar.

Nextcloud Calender Issue - Long Description

### LINKED ISSUE #2459: Show "Event not found" when an event is not available
When trying to open an event which doesn't exist (simply edit the string in the address bar to something random), the app reloads and then shows a kind of broken sidebar:

Screenshot_2020-08-12 August 2020 - Calendar - Nextcloud

Instead of showing such default content and an everlasting spinner, I think it would be good to show that the event was not found. Otherwise the user might think something broke. This should be easily possible once https://github.com/nextcloud/nextcloud-vue/pull/1281 is in.

georgehrke:
fixes #2459 
fixes #2187

also removes the ugly v-if/v-show hack.

| Before | After |
|:---------:|:------:|
|before_popover|after_popover|
|before_sidebar|after_sidebar|


georgehrke:
We should add more detailed error messages later, like:
- Calendar does not exist
- Event exist, but linked recurrence is not valid (maybe with a link to just go to the next occurrence)
- Event does not exist

But that's something for a follow up PR :)

src/views/EditSidebar.vue
```diff
@@ -28,9 +28,27 @@
...
+			<div class="app-sidebar__loading-indicator">
+				<div class="icon icon-loading app-sidebar-tab-loading-indicator__icon" />
+			</div>
+		</template>
+
+		<template v-else-if="isError">
+			<EmptyContent icon="icon-calendar-dark">
+				{{ $t('calendar', 'Event does not exist') }}
+				<template #desc>
+					{{ error }}
```
raimund-schluessler, line 44: Having both `Event does not exist` and `The event does not exist` is a bit doubled, no? I think the main title would be enough.

raimund-schluessler (approved):
Works as it should. Just a minor comment.

src/views/EditSidebar.vue
tcitworld: Otherwise we can add a sentence explaining things like « It might have been deleted, or there was a typo in a link ».

georgehrke: I changed the error message as requested, looks much better now :)

PR #2532, opened 2020-09-01
Add Event to allow other apps to register custom scripts

georgehrke:


lib/Event/BeforeTemplateRenderedEvent.php
```diff
@@ -0,0 +1,83 @@
...
+	public function isEmbedded(): bool {
+		return $this->isEmbedded;
+	}
+
+	/**
+	 * @since 20.0.0
+	 * @return string|null
+	 */
+	public function getTokens(): ?string {
+		return $this->tokens;
```
georgehrke, line 81: ToDo: this should return an array of tokens instead.

ChristophWurst:
Stale

PR #2575, opened 2020-09-11
Dashboard: Filter events before now

georgehrke:
| Before | After |
|:---------:|:------:|
|C284BB19-9D51-4853-8F67-906CCA280D6F|07845959-A616-41CA-B740-09C3F28FE304|

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -141,7 +141,7 @@ export default {
 
 			await this.initializeEnvironment()
 			const expandedEvents = await this.fetchExpandedEvents(start, end)
-			this.events = await this.formatEvents(expandedEvents)
+			this.events = await this.formatEvents(expandedEvents, start)
 			this.loading = false
 		},
 		/**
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
:+1: Makes sense

juliushaertl (approved):

jancborchardt (approved):
Yeah, let’s do it like this for now – simplest fix. :) cc @rullzer

jancborchardt:
@georgehrke only one thing actually, sorry – the "All day today" event should actually show, no? So only stuff that has started before today should be hidden.

georgehrke:
Well, but the "long day from the past event" also takes place today. It's just that it started yesterday already.

jancborchardt:
Yeah, but still there’s a difference if an event e.g. a conference starts _today_ and you put it in so you remember, vs it is running already and you kinda know about it since yesterday or a few days ago already.

georgehrke:
https://github.com/nextcloud/calendar/pull/2579

PR #2578, opened 2020-09-11
Fix opening events from search

georgehrke:
Slipped through in https://github.com/nextcloud/calendar/pull/2528

Without that line, it just tries to load the event without waiting until calendars are loaded.
It fails, since it can't find the calendar.


georgehrke:
Merging, as this is required for 2.1.1

PR #2787, opened 2021-01-11
Bump fullcalendar packages to v5.5

ChristophWurst:
Tested and couldn't find anything that doesn't work anymore :shrug:


tcitworld (commented):
Why the exact version requirement?

ChristophWurst:
> Why the exact version requirement?

Because npm could update package to a compatible version if you `npm i` another package. We've run into this problem in server and Mail. I think if Dependabot is used, there is benefit in having a version *range* in the application's package.json. Now those packages are locked unless `package.json` is touched.

tcitworld:
Oh yeah, I agree that would be npm's behaviour. Fine by me.

tcitworld (approved):

PR #2823, opened 2021-01-17
Add vendor/ and coverage/ to .stylelintignore

tcitworld:
Both folders were considered when building


ChristophWurst (approved):

ChristophWurst:
there's also `/build` to ignore. Can we change it to only look into `/src` and have `/src/fonts` the only exclude? Sounds easier to manage than an exclude for every extra directory

tcitworld:
What about this?

ChristophWurst (approved):
I like it :)

PR #2827, opened 2021-01-18
Add Krankerl configuration

tcitworld:
Add initial krankerl and `.nextcloudignore` configuration, and change Makefile to use it.


ChristophWurst (approved):
:pray:

ChristophWurst:
As a follow-up we could copy and adjust https://github.com/nextcloud/mail/blob/master/.github/workflows/package.yml. It's not the most critical CI check but there were one or two occasions where that pointed out that something would have broken the next release.

PR #2902, opened 2021-02-20
Use npm ci to install front-end dependencies in dev

francoisfreitag:
## Description

The ci command is meant for any situation where you want to make sure
you're doing a clean install of your dependencies.

It is significantly faster than npm install. If a node_modules is
already present, it will be automatically removed before npm ci begins
its install.

Matches the GitHub actions to build the package.

The clean-dev rule is now unused and can be removed.

https://docs.npmjs.com/cli/v7/commands/npm-ci

1. make dev-setup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,10 @@
 all: dev-setup lint build-js-production test test-php
 
 # Dev env management
-dev-setup: clean clean-dev npm-init
+dev-setup: clean npm-init
 
 npm-init:
-	npm install
+	npm ci
 
 composer-init:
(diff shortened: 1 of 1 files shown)


francoisfreitag:
For reference, nextcloud/server uses `npm ci`. https://github.com/nextcloud/server/blob/master/Makefile#L6-L7

Cleanup of the `clean-dev` rule suggested as https://github.com/nextcloud/server/pull/25735.

ChristophWurst (approved):
:+1: :shrug:

PR #2903, opened 2021-02-22
[stable2.0] Search DAV for principal as display name OR email

ChristophWurst:
Backport of https://github.com/nextcloud/calendar/pull/2878


tcitworld (approved):
Tested and works as expected on 19

PR #2918, opened 2021-02-26
Improve free/busy colors

ChristophWurst:
## Design discussions for later reference

<details>

Bildschirmfoto von 2021-02-26 08-38-14
Bildschirmfoto von 2021-02-26 08-38-02
Bildschirmfoto von 2021-02-25 21-04-28

</details>

### DIFF
--- a/css/freebusy.scss
+++ b/css/freebusy.scss
@@ -30,19 +30,19 @@
 	.blocking-event-free-busy {
 		border-color: red;
 		border-style: solid;
-		border-left-width: 1px;
-		border-right-width: 1px;
+		border-left-width: 2px;
+		border-right-width: 2px;
 		background-color: transparent !important;
-		opacity: 1.0 !important;
+		opacity: 0.7 !important;
 		z-index: 2;
 	}
 
 	.blocking-event-free-busy.blocking-event-free-busy--first-row {
-		border-top-width: 1px;
+		border-top-width: 2px;
 	}
 
 	.blocking-event-free-busy.blocking-event-free-busy--last-row {
-		border-bottom-width: 1px;
+		border-bottom-width: 2px;
 	}
 
 	.loading-indicator {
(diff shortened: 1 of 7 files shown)


ChristophWurst:
based on your input the UI would looke like that

Bildschirmfoto von 2021-03-04 12-04-35

but that means we would completely eliminate the different types of *busy*. That is a big no from me. We have to respect what the standard says in keep the four old states.

I'll go ahead and color the striped blocks accordingly.

ChristophWurst:
> I'll go ahead and color the striped blocks accordingly.

That looks terrible as well.

I come to believe it's best if we just add the gray background for the slots where at least one other person is busy.

ChristophWurst:
With some input from @skjnldsv and @GretaD I arrived at 
Bildschirmfoto von 2021-03-05 14-46-33

the colors are up for debate but

* distinctive colors for distinctive states
* solid blocks and no striped blocks in different colors
* blocked background in max contrast -> works with all themes
* even block uses 2px instead of 1px to be a bit stronger

tcitworld:
> distinctive colors for distinctive states

Not against, but that puts ourselves at risk that the calendar color might be the same or really similar as those colors, and you wouldn't be able to notice the rectangle.

Other than this point, colors are indeed debatable (if we implement availability at some point there will probably be yellow every time on the view).

ChristophWurst:
> Not against, but that puts ourselves at risk that the calendar color might be the same or really similar as those colors, and you wouldn't be able to notice the rectangle.

The colors are hard-coded and so is the color of the rectangle. We can prevent a color conflict (to the extent that people are able to distinguish the selected colors, red/orange might be problematic already)

I'm opposed using the calendar color for the color of the rectangle for this reason. We should probably use 6 colors that are far apart each other, 5 for the states and one for the rectangle.

ChristophWurst:
Latest state

Bildschirmfoto von 2021-03-05 15-55-00
Bildschirmfoto von 2021-03-05 15-54-13

jancborchardt (approved):
Let’s go for this for now :+1: 

For the future it would still be good to improve it a bit especially making it simpler to understand rather than needing to check the color explanations (which is also an issue with color blindness). But that also requires some changes in Fullcalendar apparently.

Just as inspiration, this is from Calendar.com – quite simple and focus on the essentials: a) available yes or no and b) clear _green_ highlight of the meeting

PR #2930, opened 2021-03-03
Use Codecov without tokens

ChristophWurst:
They are only required for private repos.

Ref https://github.com/codecov/codecov-action#usage


ChristophWurst:
Oh codecov. Those numbers are so flaky. In the check the show a 0% diff. On the compare overview it shows +64% but the same happens for other unrelated PRs :shrug: 

I'd say at least nothing gets worse with this change

tcitworld (approved):

PR #2939, opened 2021-03-08
Fix right border on today not being visible in month and week view when using Firefox

mwalbeck:
This fixes #2679 

This pull request makes the right border on "today" visible again in month and week view when using Firefox.

before:
Screenshot_20210308_222601
Screenshot_20210308_222611

after:
Screenshot_20210308_222519
Screenshot_20210308_222536

Issue #2679, opened 2020-10-31
Missing border on current day in weekly and monthly view when using firefox


mwalbeck:
### Steps to reproduce
1. Open Firefox
2. Go to Nextcloud Calendar
3. Pick weekly or monthly view and see missing border on current day

The issue is not present in Chromium, specifically tested on Chromium 86.0.4240.111

### Expected behaviour
There should be a border on both left and right side of current day in weekly view and a border on all sides of current day in monthly view.

### Actual behaviour
Right border is missing from current day in weekly view, and right and bottom border is missing from current day in monthly view

### Calendar app
**Calendar app version:** 2.1.2

### Client configuration
**Browser:** Firefox 82.0

**Operating system:** Ubuntu 20.04

### Server configuration

**Operating system**: Ubuntu 20.04

**Web server:** Nginx

**Database:** Sqlite

**PHP version:** 7.4

**Nextcloud Version:** 20.0.1

**Updated from an older installed version or fresh install:** Updated from 20.0.0

![github-nextcloud-calendar-issue](https://user-images.githubusercontent.com/19759293/97785421-2dfc5f00-1ba5-11eb-9f0d-22b139b37301.png)

![github-nextcloud-calendar-issue-2](https://user-images.githubusercontent.com/19759293/97785442-48ced380-1ba5-11eb-93c6-685507ab5699.png)

bryden:
The border is in fact there but it's not visible due to the background color being set 
If you right click on today, inspect element, and then edit the css in the developer console to add a z-index of -1 here:

For the monthly view:
```
.fc .fc-daygrid-day.fc-day-today {
    background-color: rgba(255, 220, 40, 0.15);
    background-color: var(--fc-today-bg-color, rgba(255, 220, 40, 0.15));
    **z-index: -1;**
}
```

And for the weekly view:
```
.fc-theme-standard td, .fc-theme-standard th {
    border: 1px solid #ddd;
    border: 1px solid var(--fc-border-color, #ddd);
    z-index: -1;
}
```

The borders will now show up properly. Unfortunately I don't have the time to submit a code commit right away but if this lingers I may pick it up eventually because it's bothering my eyes.

spiezmaestro:
I noticed that this issue only occurs in Firefox (regardless if Breeze Dark or the default colors are active). When I logon in Chrome or Edge Chromium, all looks well. Hope this helps.

mwalbeck:
[...]

PR #2971, opened 2021-03-24
Fix right border on today not being visible in month and week view when using Firefox, again

mwalbeck:
Alternative fix for #2679 without the regression found with #2939 

Sorry about the previous fix, that regression wasn't great... Hopefully this fix is better.

Unsetting the background colour on today achieves the same result. In Calendar the background colour of today is set to `--color-main-background` which is the same as the parent table, so it isn't necessary to be set again when we want it to be the same colour. I have used `!important` as the original rule from fullcalendar is defined inline. Any change of the background during usage, like background colour change on highlight in monthly view, seems to happen inside the today element and not on the today element itself, so that isn't affected by this fix.

Before:
Screenshot_20210324_155716

After:
Screenshot_20210324_155726

cc @ChristophWurst

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -96,6 +96,12 @@
 	}
 }
 
+// Fix right border on today not being visible in month and week view when using Firefox
+.fc .fc-daygrid-day.fc-day-today,
+.fc .fc-timegrid-col.fc-day-today {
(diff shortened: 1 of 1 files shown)


GretaD (approved):
:partying_face:

PR #2977, opened 2021-03-26
Fix: when a attendee is a second person without email, skip alreadyInvitedEmails check

JackyW83:
In case if multiple attendees have empty email, the second one won't show in search result by 
block of alreadyInvitedEmails. Add email is not empty check in findAttendeesFromContactsAPI

### DIFF
--- a/src/components/Editor/Invitees/InviteesListSearch.vue
+++ b/src/components/Editor/Invitees/InviteesListSearch.vue
@@ -206,7 +206,7 @@ export default {
 						name = email
 					}
 
-					if (this.alreadyInvitedEmails.includes(email)) {
+					if (email && this.alreadyInvitedEmails.includes(email)) {
 						return
 					}
 


src/components/Editor/Invitees/InviteesListSearch.vue
```diff
@@ -220,7 +220,7 @@ export default {
 					return false
 				}
 
-				if (this.alreadyInvitedEmails.includes(principal.email)) {
+				if (this.alreadyInvitedEmails.includes(principal.email) && email) {
```
ChristophWurst, line 223: > 'email' is not defined

eslint is right :)

JackyW83: Exactly, a stupid mistake.

tcitworld: It seems like you reverted your change.

JackyW83: I added wrong place first time, i reverted it and add new code at 189 line of InviteesListSearch.vue

src/components/Editor/Invitees/InviteesListSearch.vue
```diff
@@ -186,7 +186,7 @@ export default {
 						name = email
 					}
 
-					if (this.alreadyInvitedEmails.includes(email)) {
+					if (this.alreadyInvitedEmails.includes(email) && email) {
```
ChristophWurst, line 189: but now it skips the attendee if it has an email, right?

JackyW83: No. If user has email, he/she only skip when this.alreadyInvitedEmails.includes(email) === true, and skip all users don't have email.
I updated the condition check order in latest commit, maybe this looks more intuitive.
In country like China, a lot of people don't have email, so updated this  :)

ChristophWurst (approved):

ChristophWurst:
Squash your commits into a single one and rebase to latest main

tcitworld (approved):
Logic sounds good, didn't test

miaulalala (approved):

PR #3016, opened 2021-04-09
Fix sidebar action menu

### LINKED ISSUE #2966: Calendar menu expands when I click Edit name/color
### Steps to reproduce
1. Open calendar and click the action menu in any of your calendars
2. Click Edit colour/name

### Expected behaviour
The calendar doesn't expand and the menu arrow doesnt "jump" away from the menu
### Actual behaviour
caleditnameexpand

### Calendar app

### Client configuration
**Browser:** FF
**Operating system:** Ubuntu

GretaD:
fixes #2966

before
Kazam-screencast-00015

this branch
Kazam-screencast-00016

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListItem.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListItem.vue
@@ -488,3 +488,11 @@ export default {
 	},
 }
 </script>
+<style lang="scss" scoped>
+::v-deep .action-input {
+  margin-top: 0 !important;
+}
+::v-deep .action-button {
+  width: 217px !important;
+}
+</style>


tcitworld (approved):
Works for me, but shouldn't that be an upstream nextcloud-vue issue instead?

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -488,3 +488,11 @@ export default {
 	},
 }
 </script>
+<style lang="scss" scoped>
+::v-deep .action-input {
+  margin-top: 0 !important;
+}
+::v-deep .action-button {
+  width: 217px !important;
```
raimund-schluessler, line 496: This seems to be a very specific width. Without checking, I guess this might fail for other content!?

GretaD:
> Works for me, but shouldn't that be an upstream nextcloud-vue issue instead?

technically yes, but the problem only happens on calendar. If i change on vue, other apps that use the component break. Or did i test smth wrong?

src/components/AppNavigation/CalendarList/CalendarListItem.vue
GretaD: This with is only applied to calendarlistitem, so only for the "action menu" and i dindt find any think breaking on that. I am fine with other ideas how to fix it if you have any. This worked, and i pushed, im pretty new with calendar code :D

GretaD:
@raimund-schluessler is this good to go? Do you have any other idea?

GretaD:
closing on behalf of nextcloud/nextcloud-vue#1967

Issue #2516, opened 2020-08-28
AppSettings expands to the bottom when opened


raimund-schluessler:
The AppSettings component expands to the bottom when you open it, instead of showing a scroll bar.

![Screenshot_2020-08-28 Woche 36 aus 2020 - Kalender - Nextcloud](https://user-images.githubusercontent.com/2496460/91549815-d41eb500-e927-11ea-92ed-98fd1de6a94b.png)

Happens on Win10, latest Firefox, Firefox 78.0.1 ESR and latest Chrome. The reason is this line: https://github.com/nextcloud/calendar/blob/a62cb69e90157b451f1e7ec58a6c06c5817be617/css/app-navigation.scss#L315 which forces the AppSettings content to be completely visible. Replacing it with `overflow-x: hidden;` (to prevent to long labels from showing a x-scrollbar)  prevents the issue.

georgehrke:
IIRC `overflow-x: hidden;` caused many other issues. (related to the timezone list i think)

I would say for now this is accepted (although very annoying) and close it in favor of https://github.com/nextcloud/calendar/issues/2304

raimund-schluessler:
`overflow-x: hidden` is not really necessary to fix the scrolling. Removing `overflow-y: visible` would be enough. Is there a reason you force it to be visible?

georgehrke:
> `overflow-x: hidden` is not really necessary to fix the scrolling. Removing `overflow-y: visible` would be enough. Is there a reason you force it to be visible?

![BCDFFB62-4123-48EA-A80E-E168A0E3703B](https://user-images.githubusercontent.com/1250540/91550768-58256c80-e929-11ea-99ab-b01f4406efc9.png)

The timezone list is cut off and barely shows 4 entries.

georgehrke:
Same for the timeslot size option:
![8CED4AC0-9F1D-47BB-9E71-2F2D33C6E353](https://user-images.githubusercontent.com/1250540/91551767-01209700-e92b-11ea-9608-3429cbd7335c.png)

raimund-schluessler:
Well, you can scroll to show the rest of the options. But I agree, that is far from perfect.

raimund-schluessler:
I guess #2304 really is the way to go then.

PR #3026, opened 2021-04-16
Fix settings scroll problem

### LINKED ISSUE #2967: Calendar moves/scrolls when I open Settings and Import
### Steps to reproduce
1. Open calendar and click Settings & Import
2. Scroll
3. The calendar scrolls, creating a gap

### Expected behaviour
The calendar doesnt move

### Actual behaviour
The calendar moves/scrolls
settingscalendar

### Calendar app

GretaD:
fixes #2967 

before - check fixes

after
aftercalendarsettings

### DIFF
--- a/css/app-navigation.scss
+++ b/css/app-navigation.scss
@@ -312,7 +312,7 @@
 	#app-settings {
 
 		#app-settings-content {
-			overflow-y: visible !important;
+			overflow-y: auto !important;
 		}
 
 	}


raimund-schluessler:
I think `overflow: visible` is there to prevent issues with the multiselects for timezone and duration. `overflow: auto` makes it difficult to select a value and leads to a scrollbar in y-direction. At least this was the status of the discussion in https://github.com/nextcloud/calendar/issues/2516.

GretaD:
> I think `overflow: visible` is there to prevent issues with the multiselects for timezone and duration. `overflow: auto` makes it difficult to select a value and leads to a scrollbar in y-direction for some locals. At least this was the status of the discussion in #2516.

ohh i see, thank you for pointing  it out, i was trying to find smth if its was done on purpose but failed. So I guess, would be better to put the whole settings into a modal and this way we fix the "crowed" problem and this bug.

GretaD:
We decided on Groupware channel to move settings into a modal, so it will be fixed there.

PR #3080, opened 2021-05-15
l10n: Extract a space from the translation

Valdnet:

### DIFF
--- a/lib/Controller/EmailController.php
+++ b/lib/Controller/EmailController.php
@@ -134,7 +134,7 @@ class EmailController extends Controller {
 		}
 
 		return new JSONResponse([
-			'message' => $this->l10n->t('Successfully sent email to ' . $recipient),
+			'message' => $this->l10n->t('Successfully sent email to %1$s', [$recipient]),
 		]);
 	}
 


tcitworld (changes requested):
Nope, the variable should be brought into the translation instead of concatening.

Valdnet:
Corrected variable, but I don't know if the bot will scan the text string correctly.

@tcitworld Is it ok now?

lib/Controller/EmailController.php
```diff
@@ -134,7 +134,7 @@ public function sendEmailPublicLink(string $recipient,
 		}
 
 		return new JSONResponse([
-			'message' => $this->l10n->t('Successfully sent email to ' . $recipient),
+			'message' => $this->l10n->t('Successfully sent email to' . ' ' . $recipient),
```
raimund-schluessler, line 137: @Valdnet I think @tcitworld meant something like this (not tested):
```suggestion
			'message' => $this->l10n->t('Successfully sent email to %1$s', $recipient),
```

rakekniven (approved):

lib/Controller/EmailController.php
```diff
@@ -134,7 +134,7 @@ public function sendEmailPublicLink(string $recipient,
 		}
 
 		return new JSONResponse([
-			'message' => $this->l10n->t('Successfully sent email to ' . $recipient),
+			'message' => $this->l10n->t('Successfully sent email to %1$s', $recipient),
```
tcitworld, line 137: CI says it should be an array.
```suggestion
			'message' => $this->l10n->t('Successfully sent email to %1$s', [$recipient]),
```

tcitworld (changes requested):

tcitworld (approved):

PR #3098, opened 2021-05-19
Bump @nextcloud/vue to v3, @nextcloud/vue-dashboard to v2

ChristophWurst:
Required for the trashbin UI where I need the `#extra` slot in `AppNavigationItem`. Took me longer than I want to admit finding out the slot doesn't work because it simply doesn't exist in the ancient version of the vue lib we have here in Calendar  …

Tested and can't find anything that doesn't work anymore.


kesselb (approved):
LGTM :+1:

PR #3101, opened 2021-05-20
Fix attendees avatars

tcitworld:


src/store/contacts.js
```diff
@@ -71,8 +75,31 @@ const mutations = {
...
+	 *
+	 * @param {Object} state the store data
+	 * @returns {function({String}): {String}}
+	 */
+	getAvatarForContact: (state) => (uri) => {
+		const contact = state.contactByEMail[stripMailTo(uri)]
+		return contact?.isUser ? undefined : contact?.avatar
+	},
+
+	isContactAnUser: (state) => (uri) => {
```
nickvergessen, line 90:
```suggestion
	isContactAUser: (state) => (uri) => {
```
https://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user

tcitworld: God-dammit English

sazanof:
Hello! I would like to ask. I am currently working on attachments in a parallel branch https://github.com/nextcloud/calendar/pull/4251#discussion_r919033226 . According to the CalDAV specification, we can use our own parameters:  ex  `X-NC-ATTENDEE-PHOTO` (May be it will be usefull)

So, can the avatar url parameter be passed when adding a user? In my case its look like this in .ics file:
image

SebastianKrupinski:
@tcitworld  is this PR still needed? Or can it be closed?

tcitworld:
I think it was fixed elsewhere? Wasn't a definitive fix anyway.

Issue #2644, opened 2020-10-13
Event text is cut off in day/week view


cole8888:
### Steps to reproduce
1. Update to Nextcloud 20.0.0 beta
2. Create an event with a time duration which will cause the text to get cut off. (0mins, 20mins, 40mins, 50mins, etc...)

### Expected behaviour
I expected that despite the time duration of the event, the time and title should be visible. This was the case when using Nextcloud 19.

### Actual behaviour
Various event time duration ranges cause the event text and or time to be completely hidden or cut off.
![image](https://user-images.githubusercontent.com/32819560/95914685-7978d580-0d7c-11eb-910f-ec5fe4139a68.png)

### Calendar app
**App version:** 2.1.2

**CalDAV-clients used:** N/A

### Client configuration
**Browser:** Firefox 81.0

**Operating system:** Kubuntu 20.04

### Server configuration

**Operating system**: Ubuntu Server 20.04
**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.4.3

**Nextcloud Version:** 20.0.0 (beta)

**Updated from an older installed version or fresh install:** Upgraded from 19

**List of activated apps:**
```
Enabled:
  - activity: 2.13.1
  - admin_audit: 1.10.0
  - analytics: 3.0.0
  - breezedark: 19.0.5
  - calendar: 2.1.2
  - cloud_federation_api: 1.3.0
  - contactsinteraction: 1.1.0
  - dav: 1.16.0
  - drawio: 0.9.7
  - extract: 1.2.4
  - federatedfilesharing: 1.10.1
  - files: 1.15.0
  - files_pdfviewer: 2.0.1
  - files_rightclick: 0.17.0
  - files_sharing: 1.12.0
  - files_trashbin: 1.10.1
  - files_versions: 1.13.0
  - files_videoplayer: 1.9.0
  - groupfolders: 8.1.0
  - keeweb: 0.6.3
  - logreader: 2.5.0
  - lookup_server_connector: 1.8.0
  - metadata: 0.12.0
  - oauth2: 1.8.0
  - photos: 1.2.0
  - previewgenerator: 3.0.1
  - provisioning_api: 1.10.0
  - richdocuments: 3.7.4
  - serverinfo: 1.10.0
  - settings: 1.2.0
  - theming: 1.11.0
  - twofactor_backupcodes: 1.9.0
  - twofactor_totp: 5.0.0
  - user_status: 1.0.0
  - video_converter: 0.1.4
[...]

PR #3121, opened 2021-05-28
Add min-height limits for events

mrtumnus:
Both condensed (one-line) and normal (2+ lines) events have no minimum height limit, resulting in cut-off event time and/or text.  This change introduces minimum limits for both event styles, guaranteeing that both time and text will be shown.  The trade-off here is that the event height no longer directly correlates exactly to the length of the event.

Before this change:
image

With this change:
image

This addresses #2644

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -184,3 +184,11 @@
 		margin-top: 0 !important;
 	}
 }
+
+.fc-v-event {
+        min-height: 4em;
+}
+
+.fc-v-event.fc-timegrid-event-condensed {
+        min-height: 2em;
+}


ChristophWurst:
>  The trade-off here is that the event height no longer directly correlates exactly to the length of the event.

I'm not sure if I like that.

How do other calendar solutions handle this? Is there a way of making the font scale better with short events?

mrtumnus:
> > The trade-off here is that the event height no longer directly correlates exactly to the length of the event.
> 
> I'm not sure if I like that.
> 
> How do other calendar solutions handle this? Is there a way of making the font scale better with short events?

I suppose I should rephrase this - "event height no longer directly correlates exactly to the length of the event, _for events that result in vertical cut-off of the text_".  I don't see a way that you could get around having a minimum height; even with font scaling you'd end up with unreadable text at some point (probably even at 15 minute interval).  This change would not affect typical event lengths (30/60 min), or longer events (1+ hours); those would continue to reflect event length.

That said, I haven't answered your question.  I'll look at least at Google and Microsoft approaches & report here.

mrtumnus:
Google Calendar below.  Looks like they do some font scaling AND have a minimum height (1, 5, 10 and 15 min events all have the same, minimum height).

image

mrtumnus:
And Outlook.  No font scaling, minimum size looks to be 30 minutes.  However, it uses a progress bar on the left to represent event length, which does scale.

image

mrtumnus:
And just for kicks, here's Thunderbird.  No font scaling, minimum height equivalent to 15 min. event.

image

ChristophWurst:
Thanks a lot for the thorough research. I guess I'm fine with this then as users are potentially already familiar with the way short events are displayed.

A nice detail from Google is that they use *10 - 11 AM* instead of *10 AM - 11 AM*. We could possibly take some inspiration in another improvement of the event rendering :)

ChristophWurst (approved):

PR #3129, opened 2021-05-31
Add delete button for trash bin

### LINKED ISSUE #3127: Allow users to delete trash bin elements permanently
After https://github.com/nextcloud/calendar/pull/3118 I will see my trash bin elements but I can only restore them. There should be an option for deletion as well.

### LINKED ISSUE #3128: "Trashbin" vs "Trash bin"
The trash bin feature from https://github.com/nextcloud/calendar/pull/3118 has a few naming inconsistencies. We should replace all UI strings that say *trashbin* to *trash bin*.

GretaD:
Fixes #3128
Fixes #3127

- [x] Requires https://github.com/nextcloud/server/pull/27343
- [x] Requires https://github.com/nextcloud/cdav-library/pull/480


ChristophWurst (approved):
Works :)

PR #3130, opened 2021-05-31
Add a footer for trash bin

### LINKED ISSUE #3126: Tell the user when their trashbin elements will be deleted
As a leftover of https://github.com/nextcloud/calendar/pull/3118 there is no info for the users about when elements of the trash bin get deleted

### Steps to reproduce
1. Delete an event in the currently viewed time range
2. Open the trash bin

### Expected behavior

See explanatory text that tells me when elements in the trash bin will be deleted permanently. Wording could be "Elements in the trash bin are deleted after 30 days". Ideally we fetch the exact retention time from the server, so this might need a tiny change there.

### Actual behavior

Users don't know when the elements vanish. They might think they stay there forever.

### Calendar app
**Calendar app version:** Master after https://github.com/nextcloud/calendar/pull/3118

GretaD:
Fixes #3126

- [x] Requires https://github.com/nextcloud/server/pull/27348
- [x] Requires https://github.com/nextcloud/cdav-library/pull/481

### DIFF
--- a/src/components/AppNavigation/CalendarList/Trashbin.vue
+++ b/src/components/AppNavigation/CalendarList/Trashbin.vue
@@ -50,6 +50,9 @@
 							</td>
 						</tr>
 					</table>
+					<p class="footer">
+						{{ n('calendar', 'Elements in the trash bin are deleted after {numDays} day', 'Elements in the trash bin are deleted after {numDays} days', retentionDuration, { numDays: retentionDuration }) }}
+					</p>
 				</div>
 			</Modal>
 		</template>
@@ -115,6 +118,11 @@ export default {
 
 			return formattedCalendars.concat(formattedCalendarObjects)
 		},
+		retentionDuration() {
+			return Math.ceil(
+				this.trashBin._props['{http://nextcloud.com/ns}trash-bin-retention-duration'] / (60 * 60 * 24)
+			)
+		},
 	},
 	methods: {
 		async onShow() {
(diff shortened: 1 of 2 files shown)


src/components/AppNavigation/CalendarList/Trashbin.vue
```diff
@@ -115,6 +118,11 @@ export default {
 
 			return formattedCalendars.concat(formattedCalendarObjects)
 		},
+		retentionDuration() {
+			return Math.ceil(
+				this.trashBin._props['{http://nextcloud.com/ns}trash-bin-retention-duration'] / (60 * 60 * 24)
```
tcitworld, line 123: On the server side, the `calendarRetentionObligation` app config is saved in seconds, so we need to expect this value not to be an exact number of days...and I've noticed the `Math.ceil()` above afterwards, good catch. :see_no_evil:

tcitworld: The check warnings are sometimes hiding things.

ChristophWurst: Yep it's not ideal. Moment.js can't format a time duration right now, but possibly in the future we can try to translate the seconds into something human readable and then also durations like *Two and a half days* could show accordingly instead of the rounded up *3 days*.

ChristophWurst (approved):
@GretaD please give this a final test but I think we can merge this :)

tcitworld:
Is it worth it to handle the property being not found?

Since https://github.com/nextcloud/server/pull/27348 is not yet in the latest beta, there's a very nice `NaN`.
image

ChristophWurst:
Sure, some robustness would be good!

PR #3140, opened 2021-06-02
l10n: Change "at" to "on"

Valdnet:
Reported by a Transifex translator.


Valdnet:
> Should it be just Deleted maybe?

A very good solution. I thought so too, because it's in the table, but I wanted your opinion.

Corrected 😉.

ChristophWurst (approved):

PR #3147, opened 2021-06-04
Hide the trash bin if it's disabled

ChristophWurst:
If an admin sets the retention duration to 0 then deletions will drop
data right away and remaining deleted items will be killed in the next
cron job. So there is nothing for the user to restore and we can safely
hide the item from the sidebar.

### DIFF
--- a/src/store/calendars.js
+++ b/src/store/calendars.js
@@ -401,7 +401,7 @@ const getters = {
 	},
 
 	hasTrashBin(state) {
-		return state.trashBin !== undefined
+		return state.trashBin !== undefined && state.trashBin.retentionDuration !== 0
 	},
 
 	trashBin(state) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -5930,7 +5930,7 @@
 			"dev": true
 		},
 		"cdav-library": {
-			"version": "git+https://github.com/nextcloud/cdav-library.git#00e0479f545cd9d9fc445e70596ec154c106d3c7",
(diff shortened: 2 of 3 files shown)


src/store/calendars.js
```diff
@@ -401,7 +401,7 @@ const getters = {
 	},
 
 	hasTrashBin(state) {
-		return state.trashBin !== undefined
+		return state.trashBin !== undefined && state.trashBin._props['{http://nextcloud.com/ns}trash-bin-retention-duration'] !== 0
```
tcitworld, line 404: `state.trashBin._props['{http://nextcloud.com/ns}trash-bin-retention-duration']` can be undefined as well, in which case we should hide it too.

tcitworld (changes requested):

src/store/calendars.js
ChristophWurst: Should we? I thought about this scenario and this can only mean the server doesn't send the prop. We now have https://github.com/nextcloud/calendar/pull/3146 for that.

tcitworld: Yeah but things can always go wrong in-between. 

I'm already not so fond of accessing `_props` like this, so instead we could use `_exposeProperty` in cdav and do a simpler check here.

ChristophWurst: > e could use `_exposeProperty` in cdav and do a simpler check here

Let me do that.

ChristophWurst: https://github.com/nextcloud/cdav-library/pull/482

tcitworld (approved):

PR #3159, opened 2021-06-06
Show shared calendars by default

### LINKED ISSUE #3022: Shared calender not active by default?!
### Steps to reproduce
1. Share a calender to a group

### Expected behaviour
User of that group should get an active calender, i.e. don't need to click on that calender so the circle changes from grey to color.

### Actual behaviour
The calender is not active, so the user also doesn't get the events in the dashboard.

### Why is that a problem
I then need to activate that calender for every new user by myself.

Calendar:  2.2.0

burnoutberni:
Changes the `enabled` default behavior. Instead of hiding shared calendars by default, now they are shown and users instead have to manually disable shared calendars in order to hide them. Fixes #3022.

### DIFF
--- a/src/models/calendar.js
+++ b/src/models/calendar.js
@@ -122,10 +122,8 @@ const mapDavCollectionToCalendar = (calendar, currentUserPrincipal) => {
 		// If calendar-enabled is set, we will just take that
 		enabled = calendar.enabled
 	} else {
-		// If there is no calendar-enabled,
-		// we will display the calendar by default if it's owned by the user
-		// or hide it by default it it's just shared with them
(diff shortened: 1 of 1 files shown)


ChristophWurst (changes requested):
Let's first discuss https://github.com/nextcloud/calendar/issues/3022#issuecomment-855645920

ChristophWurst (approved):

ChristophWurst:
CI is not happy. Please check the test out and sign off your commits for the DCO

burnoutberni:
Thanks, @rponline, I'm closing my PR in favor of yours.

PR #3222, opened 2021-06-16
l10n: Add dot

Valdnet:


tcitworld (approved):

rakekniven:
@Valdnet You added a dot here. As a heavy Transifex user I am not aware of any rule to add a dot or not.
There is a mix. Do you know any rule for that? So why did you add a dot here?

tcitworld:
I approved because of coherence with above other messages.

Valdnet:
@rakekniven Your question has already been answered by tcitworld 😉.

> There is a mix.

I noticed it too.

PR #3215, opened 2021-06-14
Bump deps / `@nextcloud/webpack-vue-config` and use npm7

skjnldsv:
ref https://github.com/nextcloud/app-tutorial/pull/319

- Feel free to revert all my github actions changes
- There is ONE linting issue that require a bit more attention
- Builds fine BUT didn't look extensively at the ui (seems good at first sight)

PR #3223, opened 2021-06-16
Bump node and npm version in package.json

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,8 @@
 		"extends @nextcloud/browserslist-config"
 	],
 	"engines": {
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

raimund-schluessler:
@ChristophWurst https://github.com/nextcloud/calendar/pull/3215 was not merged yet, so package-lock.json is still at version 1. I guess this will lead to problem now that npm is required to be v7 with this PR.

PR #3251, opened 2021-06-23
Use ISO8601 timestamps for the CalDAV trashbin

ChristophWurst:
- [x] Requires https://github.com/nextcloud/server/pull/27632
- [x] Requires https://github.com/nextcloud/cdav-library/pull/503 but will also be flexible enough to work without

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6760,7 +6760,7 @@
 			"dev": true
 		},
 		"cdav-library": {
-			"version": "git+https://github.com/nextcloud/cdav-library.git#b7a567e2cff9459634e6801719ad780d4a087bbd",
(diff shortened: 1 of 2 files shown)


src/components/AppNavigation/CalendarList/Trashbin.vue
```diff
@@ -58,7 +58,7 @@
 								</div>
 							</td>
 							<td class="deletedAt">
-								<Moment class="timestamp" :timestamp="item.deletedAt" />
+								<Moment class="timestamp" :timestamp="toTimestamp(item.deletedAt)" />
```
tcitworld, line 61: Might as well configure the `<Moment>` component to use Date types instead of timestamps, and then invert the convert logic.

tcitworld (approved):

Issue #3273, opened 2021-06-30
Sidebar cannot be used on smaller desktops


scopseTV:
![image](https://user-images.githubusercontent.com/29678221/124009630-20ba9d00-d9de-11eb-8d16-6cfdcd3d5513.png)

The individual categories such as reminders cannot be used because the upper head of the side bar has a fixed height and the lower important part (highlighted in red) is much too small to adjust. This has only been the case since the latest update, before it was so that the entire side bar was scrollable. Unfortunately, this makes it impossible to select a reminder, for example.

PR #3277, opened 2021-07-01
Fix event repeat multiselect width and position on low resolutions

tcitworld:
Vue-Multiselect [*should* calculate the multi-select position automatically](https://github.com/shentao/vue-multiselect/blob/master/src/multiselectMixin.js#L696-L715), but that fails for some reason on small screen sizes. We can force the position on the bottom.

On small screen sizes as well the input is too small to view the placeholder fully, increasing that a bit.
Before | After
--------|------
image | image

Maybe @nextcloud/vuejs has better knowledge of this kind of Vue-Multiselect issues.

### DIFF
--- a/css/app-sidebar.scss
+++ b/css/app-sidebar.scss
@@ -128,7 +128,7 @@
 
 				.multiselect {
 					width: 100%;
-					min-width: 75px !important; // Set a lower min-width
+					min-width: 100px !important; // Set a lower min-width
 				}
 			}
 
@@ -173,7 +173,7 @@
 
 			.multiselect,
 			input[type=number] {
-				min-width: 75px;
+				min-width: 100px;
 				width: 25%;
 			}
 		}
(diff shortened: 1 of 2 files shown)


tcitworld:
Actually the same applies to https://github.com/nextcloud/calendar/issues/3273 for the Reminders tab (which isn't fixed here, but same issue).

GretaD (approved):
works

tcitworld:
Oh, I had removed the "to review" label because this is a wider issue than I first though, so there's probably a better fix.

tcitworld:
Feel free to revert if you find a better fix, or perform the fix to issue #3273 as well.

GretaD:
> Feel free to revert if you find a better fix, or perform the fix to issue #3273 as well.

ah sorry by bad, i didnt notice the tag. Yes, i will fix the second one :)

raimund-schluessler:
Might be worth checking if the update to nextcloud/vue@4.0.x helps fixing the issue. The Sidebar works well in Tasks and is scrollable and we use 4.0.2 there. Although the header has a higher z-index than the rest of the sidebar there as well, so opening multiselects towards the bottom is still necessary.

PR #3278, opened 2021-07-01
Fix 'add reminder' dropdown position

### LINKED ISSUE #3273: Sidebar cannot be used on smaller desktops

GretaD:
fixes #3273


GretaD:
ref https://github.com/nextcloud/calendar/pull/3277#issuecomment-872038265

st3iny (approved):
Works well

PR #3322, opened 2021-07-12
Replace timezoneselect with timezonepicker

GretaD:
After moving the TimezoneSelect to vue, and renamed it TimezonePicker, we need to replace it on calendar accordingly 

- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/2095

### DIFF
--- a/src/components/AppNavigation/Settings/SettingsTimezoneSelect.vue
+++ b/src/components/AppNavigation/Settings/SettingsTimezoneSelect.vue
@@ -21,10 +21,10 @@
 
 <template>
 	<li class="settings-fieldset-interior-item settings-fieldset-interior-item--timezone">
-		<TimezoneSelect
+		<TimezonePicker
 			:additional-timezones="additionalTimezones"
 			:value="timezone"
-			@change="setTimezoneValue" />
+			@input="setTimezoneValue" />
 	</li>
 </template>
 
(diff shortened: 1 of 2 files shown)


src/components/AppNavigation/Settings/SettingsTimezoneSelect.vue
```diff
@@ -21,7 +21,7 @@
 
 <template>
 	<li class="settings-fieldset-interior-item settings-fieldset-interior-item--timezone">
-		<TimezoneSelect
+		<TimezonePicker
 			:additional-timezones="additionalTimezones"
 			:value="timezone"
 			@change="setTimezoneValue" />
```
ChristophWurst, line 27: I don't think this will fire. You might need `@input="..."` instead

GretaD: i havent tested anything yet. I pushed only because i already start replacing the component

ChristophWurst (approved):

ChristophWurst (changes requested):
~~delete the now unused `TimezoneSelect`~~

tcitworld:
Looks deleted to me :thinking:

ChristophWurst:
> Looks deleted to me

Indeedy. Didn't see it in the collapsed diff :see_no_evil:

ChristophWurst:
> Looks deleted to me

Oh I think it came back. @GretaD could you please fix that?

PR #3041, opened 2021-04-26
Move settings into a modal

GretaD:
WIP

PR #3343, opened 2021-07-19
fix settings design regression

szaimen:
| Before | After |
|---|---|
| image | image |

### DIFF
--- a/css/app-navigation.scss
+++ b/css/app-navigation.scss
@@ -308,12 +308,4 @@
 			}
 		}
 	}
-
-	#app-settings {
-
-		#app-settings-content {
-			overflow-y: visible !important;
-		}
-
-	}
 }


tcitworld:
That's a quick fix but it would be better if the lines too long would wrap.

szaimen:
> That's a quick fix but it would be better if the lines too long would wrap.

Yes, but I'd suggest doing that in a follow up since this issue is much worse than lines not wrapping, imho.

tcitworld:
The proper fix would be https://github.com/nextcloud/calendar/pull/3041 :smirk:

szaimen:
> The proper fix would be #3041 😏

Yes, but can we bring it this small fix until the mentioned PR is merged?

tcitworld (approved):

szaimen:
Thanks!

GretaD:
this is not a regression, but something done on purpose, have a look here: https://github.com/nextcloud/calendar/pull/3026#issuecomment-821099937

szaimen:
Okay, then revert?

GretaD:
@tcitworld  @raimund-schluessler, what do you think?

The thing is that im working on our roudmap now, and I havent found the time to check the modal out, so i dont know when that is going to be done

szaimen:
Btw: if we should decide to leave it this way, we should also merge the still pending backport:
https://github.com/nextcloud/calendar/pull/3345

raimund-schluessler:
I would tend to keep it, since `overflow: visible` also makes the whole calendar grid moving to the top, which indeed looks broken. The increased difficulty selecting a timezone is not so critical I think. But I can live with either decision.

GretaD:
Cool, i also think the same as i already shoot a fixup some time ago :)

PR #3361, opened 2021-07-24
improve editor design

szaimen:
| Before | After |
|---|---|
| image | image |

### DIFF
--- a/src/components/Editor/Properties/PropertySelect.vue
+++ b/src/components/Editor/Properties/PropertySelect.vue
@@ -86,3 +86,16 @@ export default {
 	},
 }
 </script>
+
+<style lang="scss" scoped>
+
+.property-select {
+	margin-bottom: 4px;
+	&__input {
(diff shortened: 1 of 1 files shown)


src/components/Editor/Properties/PropertySelect.vue
```diff
@@ -86,3 +86,15 @@ export default {
 	},
 }
 </script>
+
+<style lang="scss" scoped>
+
+.property-select {
+	margin-bottom: 4px;
+	&__input {
+		width: calc(100% - 34px - 34px);
```
tcitworld, line 95: Can you add a comment to explain the 34px come from the `__icon` & `__info` width (because there might be a case where we don't add an `icon-details` element).

szaimen: done

tcitworld (approved):

PR #3365, opened 2021-07-24
Show loading view while loading deleted objects

### LINKED ISSUE #3332: The trash bin briefly shows `You don't have any deleted calendars or events` until loading has finished.
It should instead indicate that deleted items are currently loaded.

raimund-schluessler:
loading

Closes #3332.

### DIFF
--- a/src/components/AppNavigation/CalendarList/Trashbin.vue
+++ b/src/components/AppNavigation/CalendarList/Trashbin.vue
@@ -29,55 +29,72 @@
 				@close="showModal = false">
 				<div class="modal__content">
 					<h2>{{ t('calendar', 'Trash bin') }}</h2>
-					<span v-if="!items.length">{{ t('calendar', 'You do not have any deleted calendars or events') }}</span>
-					<table v-else>
-						<tr>
-							<th class="name">
-								{{ t('calendar', 'Name') }}
-							</th>
-							<th class="deletedAt">
-								{{ t('calendar', 'Deleted') }}
-							</th>
-							<th>&nbsp;</th>
-						</tr>
-						<tr v-for="item in items" :key="item.url">
-							<td>
-								<div class="item">
-									<div>
-										<div
-											class="color-dot"
-											:style="{ 'background-color': item.color }" />
-									</div>
+					<EmptyContent
+						v-if="loading"
+						icon="icon-loading"
+						class="modal__content__loading">
+						<template #desc>
+							{{ t('calendar', 'Loading deleted elements.') }}
+						</template>
+					</EmptyContent>
+					<EmptyContent
+						v-else-if="!items.length"
+						class="modal__content__empty"
+						icon="icon-delete">
+						<template #desc>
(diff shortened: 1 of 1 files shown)


src/components/AppNavigation/CalendarList/Trashbin.vue
```diff
@@ -29,55 +29,62 @@
...
-						<tr v-for="item in items" :key="item.url">
-							<td>
-								<div class="item">
-									<div>
-										<div
-											class="color-dot"
-											:style="{ 'background-color': item.color }" />
-									</div>
+					<EmptyContent v-if="loading" icon="icon-loading">
+						{{ t('calendar', 'Loading deleted calendars, events and tasks.') }}
```
tcitworld, line 33:
```suggestion
						{{ t('calendar', 'Loading deleted elements.') }}
```

src/components/AppNavigation/CalendarList/Trashbin.vue
```diff
@@ -29,55 +29,62 @@
...
+					</EmptyContent>
+					<EmptyContent v-else-if="!items.length" icon="icon-delete">
+						{{ t('calendar', 'You do not have any deleted calendars, events or tasks.') }}
```
tcitworld, line 36:
```suggestion
						{{ t('calendar', 'You do not have any deleted elements.') }}
```

tcitworld (approved):
Just a wording suggestion.

src/components/AppNavigation/CalendarList/Trashbin.vue
raimund-schluessler: Fine with me.

ChristophWurst (approved):

ChristophWurst:
the only thing that I don't like personally is how the font became bold. it looks like the same weight and size as the header is used, right? could this be lowered?

raimund-schluessler:
> 
> 
> the only thing that I don't like personally is how the font became bold. it looks like the same weight and size as the header is used, right? could this be lowered?

This comes from the EmptyContent component from nextcloud/vue in combination with a server rule for `h2` which makes headlines bold. But I don't mind. Feel free to adjust it.

raimund-schluessler:
> the only thing that I don't like personally is how the font became bold. it looks like the same weight and size as the header is used, right? could this be lowered?

How about we don't show the trash bin header at all while loading the items, but only the empty content header with the icon?

st3iny:
> > the only thing that I don't like personally is how the font became bold. it looks like the same weight and size as the header is used, right? could this be lowered?
> 
> This comes from the EmptyContent component from nextcloud/vue in combination with a server rule for `h2` which makes headlines bold. But I don't mind. Feel free to adjust it.

I fixed that and also lowered the top margin a bit to make the loading and empty messages look more centered.

Screenshot 2021-08-06 at 11-16-30 August 2021 - Calendar - Nextcloud
Screenshot 2021-08-06 at 11-16-54 August 2021 - Calendar - Nextcloud

st3iny (approved):

PR #3373, opened 2021-07-28
Use tz data from @nextcloud/calendar-js v1.1

ChristophWurst:
Incorporates https://github.com/nextcloud/calendar-js/pull/284


tcitworld (approved):
🧹

PR #3447, opened 2021-09-01
Use material-design-icons

raimund-schluessler:
All icons of the app got replaced with the material-design equivalents. This is how the main view looks now:

After:
Screenshot 2021-09-19 at 12-49-27 Woche 35 aus 2021 - Kalender - Nextcloud

Before:
Screenshot 2021-09-19 at 12-53-26 Woche 35 aus 2021 - Kalender - Nextcloud

I think that I replaced all icons. If not, it might be best to do it in a follow up (since this PR is already quite large).

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListNew.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListNew.vue
@@ -27,17 +27,25 @@
 		:menu-open.sync="isOpen"
 		menu-icon="icon-add"
 		@click.prevent.stop="toggleDialog">
+		<template #menu-icon>
+			<Plus :size="20" decorative />
+		</template>
 		<template slot="actions">
 			<ActionButton
 				v-if="showCreateCalendarLabel"
-				icon="icon-new-calendar"
 				@click.prevent.stop="openCreateCalendarInput">
+				<template #icon>
+					<CalendarBlank :size="20" decorative />
+				</template>
 				{{ $t('calendar', 'New calendar') }}
 			</ActionButton>
 			<ActionInput
 				v-if="showCreateCalendarInput"
-				icon="icon-new-calendar"
-				@submit.prevent.stop="createNewCalendar" />
+				@submit.prevent.stop="createNewCalendar">
+				<template #icon>
+					<CalendarBlank :size="20" decorative />
+				</template>
+			</ActionInput>
 			<ActionText
 				v-if="showCreateCalendarSaving"
 				icon="icon-loading-small">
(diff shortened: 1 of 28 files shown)


tcitworld (commented):
Many icons seem bigger to me, this might need an adjustement, but otherwise looks fine.

Is there a sensible build size difference?

szaimen (commented):
I wonder as well if the size differences of the icons are intentional?

Also, can we switch from `InformationVariant` to `Information` or `InformationOutline` everywhere? `InformationVariant` doesn't fit the overall style anymore, imo...

raimund-schluessler:
> Many icons seem bigger to me, this might need an adjustement, but otherwise looks fine.
> 

I used the size which is used in nextcloud/vue. I think an increase was intentional, but I don't know if the size really was decided on. For my taste it looks best with this size now, but of course we can adjust it (needs adjustments in nextcloud/vue as well then).

> 
> Is there a sensible build size difference?

I will check.

raimund-schluessler:
> Also, can we switch from `InformationVariant` to `Information` or `InformationOutline` everywhere? `InformationVariant` doesn't fit the overall style anymore, imo...

I don't mind to much. I didn't want to change it to much, since looks like this in a lot of places all over Nextcloud. And I think as an indicator for an explanation `InformationVariant` is fine. Let's hear what @nextcloud/designers think.

raimund-schluessler:
> Is there a sensible build size difference?

Master gives me a js file of 4.62 MB size, this PR generates a file of 4.65 MB size. I tend to say this doesn't matter 😄

marcoambrosini (commented):
I think it was decided to go for size 20 in the end
cc @jancborchardt & @nimisha-vijay

raimund-schluessler:
> 
> 
> I think it was decided to go for size 20 in the end
> cc @jancborchardt & @nimisha-vijay

20 pix looks quite small to me, but I have to admit I got used to the 24 pix by now 😄 This is 20 pix (top) and 24 pix (bottom):
Screenshot 2021-09-06 at 13-39-48 Woche 36 aus 2021 - Kalender - Nextcloud

src/components/AppNavigation/CalendarList/CalendarListNew.vue
```diff
@@ -27,17 +27,25 @@
 		:menu-open.sync="isOpen"
 		menu-icon="icon-add"
```
raimund-schluessler, line 28: This line needs to be removed once https://github.com/nextcloud/nextcloud-vue/pull/2217 is merged and available here. Otherwise we will see two plus icons (although no one will notice, since they perfectly overlap).

st3iny (commented):
I really like the new icons. They look much cleaner.

However, I would prefer the icons on the details tab to be a bit dimmer. Setting `opacity: 0.8` should work.

How they currently look:
Screenshot 2021-09-08 at 17-51-40 November 2021 - Calendar - Nextcloud

raimund-schluessler:
> 
> 
> I really like the new icons. They look much cleaner.
> 
> However, I would prefer the icons on the details tab to be a bit dimmer. Setting `opacity: 0.8` should work.
> 
> How they currently look:
> Screenshot 2021-09-08 at 17-51-40 November 2021 - Calendar - Nextcloud

Fine with me. Please note, that the icons are already a bit lighter than before, since they take the color of the `color` property, which is `` (not black as before). For the actions icons we use `opacity: .7`, so maybe this is what we should use:

Screenshot 2021-09-09 at 12-11-42 Woche 36 aus 2021 - Kalender - Nextcloud

raimund-schluessler:
I adjusted the icon size to 20 now. Further adjustments (e.g. changing the opacity in the sidebar tabs) I would do in a follow-up PR.

st3iny (approved):
Awesome! I tested the PR locally and there were no issues.

PR #3451, opened 2021-09-03
Set version to latest version, amend changelog

raimund-schluessler:
The version number and changelog of the master branch were lacking behind the latest released versions (stable2.3). This PR updates both.


tcitworld (approved):

ChristophWurst:
Thanks!

FYI we could also set master directly to v2.4.0-alpha.1 or similar. v2.3.x will be built from the stable2.3

PR #3459, opened 2021-09-07
Fix jsdoc/tag-lines

raimund-schluessler:
This fixes the jsdoc/tag-lines warnings after #3410.


st3iny (approved):
I'm glad that this rule got fixed upstream.

Issue #3516, opened 2021-09-28
Resource list action uses wrong icon


raimund-schluessler:
The icon in
https://github.com/nextcloud/calendar/blob/3757ade9e574a16b7a20ed3e0111637dc26fadc4/src/components/Editor/Resources/ResourceListItem.vue#L38-L42
needs to be replaced by a material design icon. Leftover from #3447.

raimund-schluessler:
> id updating class list to `class="mdi mdi-delete"` a valid solution?

No, you need to use the `icon` slot of the `ActionButton`. Have a look here:
https://github.com/nextcloud/calendar/blob/c413cbc0e3914aab8d81716ccb6cf4b9e88997e2/src/views/EditSidebar.vue#L67-L72

PR #3519, opened 2021-09-28
Replace last icons with material design variants, clean up leftovers

raimund-schluessler:
Closes https://github.com/nextcloud/calendar/issues/3516


ChristophWurst (approved):
:shipit:

Issue #3484, opened 2021-09-15
Public appointment booking page (similar to forms public pages)


ChristophWurst:
This ticket is about rendering a page of a specific appointment config of a user. Not to confuse with https://github.com/nextcloud/calendar/issues/3542, which will render an overview page.

  * [x] Show date picker
  * [x] Show slots for the selected date range
  * [x] Ability to book a slot
    * Open modal
    * User has to fill in their data or it's prefilled by the logged in user's data
    * Accept GDPR
    * Submit slot to server
    * Handle possible errors
    * Show confirmation screen

PR #3576, opened 2021-10-27
Add an appointment booking page

ChristophWurst:
- [x] Requires https://github.com/nextcloud/calendar/pull/3568

Fixes https://github.com/nextcloud/calendar/issues/3484

Todo
- [x] New route and template
- [x] Handling of tokens that don't exist
- [x] Fix links on overview page
- [x] Inject config and user info into the Vue view
- [x] Render user info as per mockup (tbd)
- [x] Render time selection as per mockup (tbd)
- [x] Fetch available slots from server for the selected range
- [x] Render slots as per mockup (tbd)
- [x] Show modal or similar as per mockup for the selected slot where the user fills in their details (prefill for registered users)
- [ ] ~~Handle booking errors~~ let's do in a follow-up
- [ ] ~~Show booking success~~ let's do in a follow-up


GretaD:
added this skeleton, will continue improving it tomorrow
Screenshot from 2021-11-01 17-09-13

src/components/Appointments/AppointmentDetails.vue
```diff
@@ -0,0 +1,144 @@
...
+				</div>
+				<input
+					ref="email"
+					type="email"
+					:value="email"
+					autocapitalize="none"
+					autocomplete="on"
+					autocorrect="off"
+					required>
+				<button class="button add-guest"
```
ChristophWurst, line 42: we don't support guests for now. you can drop this button.

src/views/Appointments/Booking.vue
```diff
@@ -0,0 +1,193 @@
...
+			<div class="booking__slots">
+				<AppointmentSlot v-for="slot in slots"
+					:key="slot.start"
+					:start="slot.start"
+					:end="slot.end"
+					:time-zone-id="timeZone" />
+			</div>
+			<button class="button primary"
+				@click.prevent.stop="showAppointmentDetails ">
+				{{ $t('calendar', 'Next') }}
```
ChristophWurst, line 62: what's the purpose of this button @GretaD?

GretaD: to show the modal, or the next step of booking the appointment, i was not sure how to do that. So i did the easiest think that i could think of, but we can remove it/or do smth else.

src/components/Appointments/AppointmentSlot.vue
```diff
@@ -0,0 +1,72 @@
...
+		},
+		timeZoneId: {
+			required: true,
+			type: String,
+		},
+	},
+	computed: {
+		dateTimeFormatter() {
+			// TODO: detect locale as well
+			return Intl.DateTimeFormat('en-US', {
```
tcitworld, line 48: Just pass `undefined` instead of `'en-US'`, it will use the browser's locale.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument

src/components/Appointments/AppointmentSlot.vue
```diff
@@ -0,0 +1,72 @@
...
+				timeZone: this.timeZoneId,
+				timeStyle: 'full',
+				dateStyle: 'short',
+			})
+		},
+		startTime() {
+			// TODO: is there a better way to determine the locale? The one from @nextcloud/l10n is not always supported
```
tcitworld, line 55: Supported by?

ChristophWurst: `Date.toLocaleString`. It doesn't work with `de_AT` for example

tcitworld: That's because it should be `de-AT`. :)

ChristophWurst: Yeah, figured. But then we have to convert and I'm not sure if replacing `_` with `-` works for all possibly locales

tcitworld: That's pretty much what we do in https://github.com/nextcloud/calendar/blob/71be054f7d58556d164bfdd0d72551a8a98ec42e/src/utils/moment.js#L34-L35

But that's also indeed what cause trouble when Nextcloud locales aren't supported by fullcalendar.

ChristophWurst: I discovered `getCanonicalLocale` that does exactly that and added some try-catch around to fall back to `undefined`. This *should* work universally.

miaulalala (approved):

PR #3592, opened 2021-11-02
Modify Migration for Appointments and Add Start and End

miaulalala:
- [x] Add a unique index on the appointment config token
- [x] Add start and end date
- [x] rename buffer field to be more self explanatory

### DIFF
--- a/lib/Db/AppointmentConfig.php
+++ b/lib/Db/AppointmentConfig.php
@@ -23,8 +23,6 @@ declare(strict_types=1);
  *
  */
 
-// @TODO rename and adjust types from migration
-
 namespace OCA\Calendar\Db;
 
 use JsonSerializable;
@@ -53,6 +51,10 @@ use function json_encode;
  * @method void setCalendarFreebusyUris(?string $freebusyUris)
  * @method string getAvailability()
  * @method void setAvailability(?string $availability)
+ * @method int|null getStart()
+ * @method void setStart(?int $start)
+ * @method int|null getEnd()
+ * @method void setEnd(?int $end)
  * @method int getLength()
  * @method void setLength(int $length)
  * @method int getIncrement()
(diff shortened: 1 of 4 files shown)


miaulalala:
Having a few issues with where to add the start & end params to methods. Need to check again tomorrow.

lib/Db/AppointmentConfig.php
```diff
@@ -27,6 +27,7 @@
 
 namespace OCA\Calendar\Db;
 
+use DateTimeInterface;
```
ChristophWurst, line 30: unused, right?

lib/Migration/Version2040Date20210908101001.php
```diff
@@ -83,9 +89,10 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
 		]);
-		$table->addColumn('buffer', Types::INTEGER, [
-			'notnull' => true,
-			'default' => 0
+		// Minimum time before next appointment slot can be booked
+		$table->addColumn('time_before_next_slot', Types::INTEGER, [
+			'notnull' => false,
+			'default' => null
 		]);
 		//Maximum slots per day - if 0, fit as many as possible
```
ChristophWurst, line 97: this is outdated, isn't it? We store `null` if there is no restriction

lib/Migration/Version2040Date20210908101001.php
```diff
@@ -94,6 +101,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		]);
 
 		$table->setPrimaryKey(['id']);
+		$table->addUniqueIndex(['token'], 'token_uniq_idx');
```
ChristophWurst, line 104:
```suggestion
		$table->addUniqueIndex(['token'], 'cal_appt_token_uniq_idx');
```

because I think on some DBs the indexes are not specific to the table but global and `otken_uniq_idx` could potentially lead to a conflict

ChristophWurst (commented):
Nitpicks, the rest looks great :)

Issue #3478, opened 2021-09-15
Provide REST API to book an appointment slot


ChristophWurst:
* Validate if the date corresponds to an actual slot and that the slot is available
  * Create the event
    * Requires https://github.com/nextcloud/server/issues/20154
    * Attach custom property with the appointment slot configuration ID
  * Send emails

PR #3600, opened 2021-11-04
Add the booking backend for Appointments

miaulalala:
Fixes https://github.com/nextcloud/calendar/issues/3478

To Do:

- [x] Test the mechanism with different configurations
- [x] Check if the timezone from the DateTimeImmutable Object in BookingCalendarWriter is resolved properly by Sabre so it is correctly included in the .ical file
- [x] Check if the .ical is in every attendee's calendar
- [x] Create booking DB
  - appt_config_id
  - created_at
  - start
  - end
  - timezone of booking user
  - name
  - email
  - description
  - token (rand32, unique index)
- [x] Create booking method that writes to DB (Entity, Mapper, etc)
- [x] Create Confirmation Email with confirmation token URL
  -  information about having to send an email to the organizer to cancel
  - Follow up fro NC 24: https://github.com/nextcloud/server/issues/29620
- [x] Create confirmation method that looks up booking via token and uses the public calendar write to create the .ical for organizer and attendee (PARTAT ACCEPTED)
- [x] Page for confirmation "Your booking has been confirmed / denied" -> check slots again
  - Event details
- [x] Add background job that removes all expired bookings
- [x] Write Tests
  - [x] MailService
  - [x] BookingService
- [x] Add flag for confirmation instead of deleting
- [x] rework the page for confirmed bookings
  - [x] Expired links should say so and link to appts page
  - [x] Confirmed bookings should say so

Follow Up: https://github.com/nextcloud/calendar/issues/3630


lib/Controller/BookingController.php
```diff
@@ -137,10 +149,12 @@ public function bookSlot(int $appointmentConfigId,
 			return JsonResponse::fail(null, Http::STATUS_NOT_FOUND);
 		}
 
-		// validate slot is available
-		// write slot, preparation duration and follow up duration events
-//		$this->bookingService->book($startTimeInTz, $endTimeInTz, $config, $start);
+		try {
+			$slot = $this->bookingService->book($config, $startTimeInTz, $endTimeInTz, $start, $name, $email, $description);
+		} catch (ServiceException $e) {
```
ChristophWurst, line 154: also catch client exceptions and respond accordingly. and I don't think any service exception should be answered with a 404 :)

lib/Migration/Version3000Date20211109132439.php
```diff
@@ -0,0 +1,78 @@
...
+namespace OCA\Calendar\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
```
ChristophWurst, line 14: 🌝

miaulalala: It wouldn't be a real migration without it

lib/Db/BookingMapper.php
```diff
@@ -0,0 +1,55 @@
...
+
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Db\MultipleObjectsReturnedException;
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\Exception as DbException;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+
+/**
+ * @template-extends QBMapper<AppointmentConfig>
```
ChristophWurst, line 36: :thinking:

lib/Service/Appointments/BookingService.php
```diff
@@ -84,4 +212,131 @@ public function updateBooking() {
...
+
+		$bookingUrl = $this->urlGenerator->linkToRouteAbsolute('calendar.booking.confirmBooking', ['token' => $booking->getToken()]);
+		$template->addBodyButton($this->l10n->t('Confirm'), $bookingUrl);
+
+		$bodyText = $this->l10n->t('This confirmation link expires in 24 hours.');
+		$template->addBodyText($bodyText);
+
+		$bodyText = $this->l10n->t("If you wish to cancel the appointment after all, please contact your organizer:");
+		$template->addBodyText($bodyText);
+		$template->addBodyText($this->l10n->t("Message $fromEmail"));
```
ChristophWurst, line 262: :thinking:

lib/Service/Appointments/BookingService.php
```diff
@@ -84,4 +212,131 @@ public function updateBooking() {
...
+				Http::STATUS_NOT_FOUND
+			);
+		}
+	}
+
+	/**
+	 * @param $booking
+	 * @throws DbException
+	 */
+	public function deleteEntity($booking) {
```
ChristophWurst, line 337: :thinking: type hint

lib/Service/Appointments/Mailer.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Calendar\Service\Appointments;
+
+class Mailer {
+
```
ChristophWurst, line 30: :thinking:

src/views/Appointments/Confirmation.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+	<div class="update">
+		The slot for your appointment from
```
ChristophWurst, line 3: translations missing

src/views/Appointments/Conflict.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+	<div class="update">
+		The slot for your appointment from
```
ChristophWurst, line 3: translations missing

ChristophWurst (commented):
early feedback :)

lib/BackgroundJob/CleanUpJob.php
```diff
@@ -0,0 +1,60 @@
...
+	/** @var BookingMapper */
+	private $mapper;
+	/** @var ILogger */
+	private $logger;
+
+	public function __construct(ITimeFactory $time,
+								BookingMapper $mapper,
+	ILogger $logger) {
+		parent::__construct($time);
+		$this->setInterval(24);
```
ChristophWurst, line 49: https://github.com/nextcloud/server/blob/d706c7e0625c87d1c3a5cde683d09abc9e920567/lib/public/BackgroundJob/TimedJob.php#L45

miaulalala: uh oh :scream:

miaulalala:
Would it be good to have a follow up that allows setting a limit for when bookings are still possible? say - bookings are only allowed 24 / 48 / etc  hours in the future and the booking page will not show any slots for the next 24 hours?

lib/Service/Appointments/Mailer.php
miaulalala: I still don't understand what the issue is here :slightly_frowning_face:

ChristophWurst:
> Would it be good to have a follow up that allows setting a limit for when bookings are still possible? say - bookings are only allowed 24 / 48 / etc hours in the future and the booking page will not show any slots for the next 24 hours?

You mean the TTL for a booking to be confirmed? Let's use a sensible default.

For the time before next booking we already have a setting.

miaulalala:
Some tests are still missing:

- [ ] MailService
- [ ] BookingService

appinfo/info.xml
```diff
@@ -15,7 +15,7 @@
 * ☑️ Tasks! See tasks with a due date directly in the calendar
 * 🙈 **We’re not reinventing the wheel!** Based on the great [c-dav library](https://github.com/nextcloud/cdav-library), [ical.js](https://github.com/mozilla-comm/ical.js) and [fullcalendar](https://github.com/fullcalendar/fullcalendar) libraries.
 	]]></description>
-	<version>2.4.0-RC.1</version>
+	<version>3.0.0-alpha</version>
```
ChristophWurst, line 18: 3.0.0-alpha.1

ChristophWurst (dismissed):

lib/BackgroundJob/CleanUpJob.php
```diff
@@ -0,0 +1,56 @@
...
+								BookingMapper $mapper,
+								LoggerInterface $logger) {
+		parent::__construct($time);
+		$this->setInterval(24*60*60);
+		$this->mapper = $mapper;
+		$this->logger = $logger;
+	}
+
+	protected function run($argument): void {
+		$outdated = $this->mapper->deleteOutdated();
```
ChristophWurst, line 53: nit: we might want to move this into a method of the booking service, and move the actual retention time into an argument of `deleteOutdated`. Then the background job and the mapper are dumb and the logic ("how long do we keep bookings?" is in the service

lib/Db/AppointmentConfig.php
```diff
@@ -37,8 +37,8 @@
  * @method void setToken(string $token)
  * @method string getName()
  * @method void setName(string $name)
- * @method string getDescription()
- * @method void setDescription(string $name)
+ * @method string|null getDescription()
+ * @method void setDescription(?string $name)
```
ChristophWurst, line 41:
```suggestion
 * @method void setDescription(?string $description)
```

copy pasta from the first setter all the way down to the last ;-)

lib/Migration/Version3000Date20211109132439.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
```
ChristophWurst, line 3: license header missing

ChristophWurst (commented):
Looks great otherwise :rocket:

lib/BackgroundJob/CleanUpJob.php
```diff
@@ -0,0 +1,56 @@
...
+class CleanUpJob extends TimedJob {
+
+	/** @var BookingMapper */
+	private $mapper;
+
+	/** @var ILogger */
+	private $logger;
+
+	public function __construct(ITimeFactory $time,
+								BookingMapper $mapper,
```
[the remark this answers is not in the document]
miaulalala: :grey_question: i is confushon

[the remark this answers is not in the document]
ChristophWurst: my bad, read the code wrong.

ChristophWurst:
- [ ] BUG the token is leaked to the front-end in the json sent for the successful booking. A malicious user could use it to generate the confirmation link without access to the email. I'd suggest we don't give the booking json to the front-end. It's not used after all.

ChristophWurst:
- [ ] Link from confirmation email leads to internal server error due to `Slot for booking is not available any more`. On reload it suddenly works with "Thank you, your booking ...".

ChristophWurst:
* [ ] BUG: *See all available slots* leads to invalid URL

ChristophWurst:
- [ ] BUG: empty bullet point for empty description

Bildschirmfoto von 2021-11-22 09-09-38

ChristophWurst:
- FOLLOW-UP: Rename event description from just the appointment name to "{appointmentName} with {attendeeName}" or similar. Don't know if this works with all languages, we might have to set the appointment name in quotes.

Bildschirmfoto von 2021-11-22 09-12-39

lib/Controller/BookingController.php
```diff
@@ -86,60 +113,143 @@ public function getBookableSlots(int $appointmentConfigId,
...
+			);
+		}
 
-		return JsonResponse::success();
+		try {
+			// This happens most likely bc of conflicting slots
+			$booking = $this->bookingService->confirmBooking($booking, $config);
+		} catch (ClientException $e) {
+			$this->logger->warning($e->getMessage(), ['exception' => $e]);
+			$link = $this->urlGenerator->linkToRouteAbsolute( 'calendar.booking.getBookableSlots', [ 'appointmentConfigId' => $config->getId() ]);
```
ChristophWurst, line 227:
```suggestion
			$link = $this->urlGenerator->linkToRouteAbsolute( 'calendar.appointment.show', [ 'token' => $config->getToken() ]);
```

because you want to redirect to the *page*, not the REST API

lib/Controller/BookingController.php
```diff
@@ -86,60 +113,143 @@ public function getBookableSlots(int $appointmentConfigId,
...
+				[],
+				TemplateResponse::RENDER_AS_GUEST
+			);
 		}
 
-		// validate slot is available
-		// write slot, preparation duration and follow up duration events
-//		$this->bookingService->book($startTimeInTz, $endTimeInTz, $config, $start);
+		try {
+			// This should never happen
```
ChristophWurst, line 210: this comment is misleading. should this be inside the catch block?

lib/Controller/BookingController.php
```diff
@@ -86,60 +113,143 @@ public function getBookableSlots(int $appointmentConfigId,
...
+				Application::APP_ID,
+				'appointments/404-booking',
+				[],
+				TemplateResponse::RENDER_AS_GUEST
+			);
+		}
 
-		return JsonResponse::success();
+		try {
+			// This happens most likely bc of conflicting slots
```
ChristophWurst, line 223: same here. this sounds like you describe the possibly thrown exception

lib/Db/Booking.php
```diff
@@ -0,0 +1,113 @@
...
+	protected $apptConfigId;
+
+	/** @var int */
+	protected $createdAt;
+
+	/** @var string */
+	protected $token;
+
+	/** @var string */
+	protected $displayName = '';
```
ChristophWurst, line 69: don't set a default value for entity properties. this won't be used in production code.

lib/Db/Booking.php
```diff
@@ -0,0 +1,113 @@
...
+ * @method void setDescription(?string $description)
+ * @method string getEmail()
+ * @method void setEmail(string $email)
+ * @method int getStart()
+ * @method void setStart(int $start)
+ * @method int getEnd()
+ * @method void setEnd(int $end)
+ * @method string getTimezone()
+ * @method void setTimezone(string $timezone)
+ * @method bool getConfirmed()
```
ChristophWurst, line 54:
```suggestion
 * @method bool isConfirmed()
```

possible because its a boolean

lib/Db/Booking.php
```diff
@@ -0,0 +1,113 @@
...
+			'created_at' => $this->getCreatedAt(),
+			'apptConfigId' => $this->getApptConfigId(),
+			'token' => $this->getToken(),
+			'displayName' => $this->getDisplayName(),
+			'description' => $this->getDescription(),
+			'email' => $this->getEmail(),
+			'start' => $this->getStart(),
+			'end' => $this->getEnd(),
+			'timezone' => $this->getTimezone(),
+			'confirmed' => $this->getConfirmed()
```
ChristophWurst, line 110:
```suggestion
			'confirmed' => $this->getConfirmed(),
```

we use trailing commas when possible because it allows for a cleaner diff when another entry is added below

lib/Db/BookingMapper.php
```diff
@@ -0,0 +1,74 @@
...
+			->where($qb->expr()->eq('token', $qb->createNamedParameter($token, IQueryBuilder::PARAM_STR), IQueryBuilder::PARAM_STR));
+		return $this->findEntity($qb);
+	}
+
+	/**
+	 * @param int $validFor is subtracted from time() and then compared against 'created_at'.
+	 * @throws DbException
+	 */
+	public function deleteOutdated(int $validFor = 0) : int {
+		$limit = $this->time->getTime() - ($validFor);
```
ChristophWurst, line 66:
```suggestion
		$limit = $this->time->getTime() - $validFor;
```

lib/Db/BookingMapper.php
```diff
@@ -0,0 +1,74 @@
...
+			->from($this->getTableName())
+			->where($qb->expr()->eq('token', $qb->createNamedParameter($token, IQueryBuilder::PARAM_STR), IQueryBuilder::PARAM_STR));
+		return $this->findEntity($qb);
+	}
+
+	/**
+	 * @param int $validFor is subtracted from time() and then compared against 'created_at'.
+	 * @throws DbException
+	 */
+	public function deleteOutdated(int $validFor = 0) : int {
```
ChristophWurst, line 65: drop the default param. the method is not intended to be used without an explicit value, is it?

lib/Service/Appointments/BookingCalendarWriter.php
```diff
@@ -0,0 +1,142 @@
...
+			'ORGANIZER',
+			'mailto:' . $organizer->getEMailAddress(),
+			[
+				'CN' => $organizer->getDisplayName(),
+				'CUTYPE' => 'INDIVIDUAL',
+				'PARTSTAT' => 'ACCEPTED'
+			]
+		);
+		$vcalendar->VEVENT->add(
+		'ATTENDEE',
```
ChristophWurst, line 111: indentation is off

lib/Service/Appointments/MailService.php
```diff
@@ -0,0 +1,178 @@
...
+
+
+		try {
+			$failed = $this->mailer->send($message);
+			if (count($failed) > 0) {
+				$this->logger->warning('Mail delivery failed for some recipients.');
+				foreach($failed as $fail) {
+					$this->logger->debug('Failed to deliver email to ' . $fail);
+				}
+			}
```
ChristophWurst, line 138: I think you want to throw your own exception here because if you reach this line it means no email was sent out. We only have one email recipient.

lib/Service/Appointments/MailService.php
```diff
@@ -0,0 +1,178 @@
...
+		} catch (\Exception $ex) {
```
ChristophWurst, line 139: import the class

lib/Service/Appointments/MailService.php
```diff
@@ -0,0 +1,178 @@
...
+			$booking->getStart(),
+			'long',
+			'short',
+			new \DateTimeZone($booking->getTimezone()),
+			$this->lFactory->get('calendar',$l)
+		);
+
+		$template->addBodyListItem($relativeDateTime, $l10n->t('Date:'));
+
+		if (isset($location)) {
```
ChristophWurst, line 163:
```suggestion
		if (!empty($location)) {
```

templates/appointments/confirmation.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * @copyright 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
```
ChristophWurst, line 6: it wasn't me

tests/php/integration/Db/BookingMapperTest.php
```diff
@@ -0,0 +1,129 @@
...
+	/** @var IDBConnection */
+	private $db;
+
+	/** @var BookingMapper */
+	private $mapper;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->db = \OC::$server->getDatabaseConnection();
```
ChristophWurst, line 52: deprecated. use `OC::$server->get(IDBConnection::class)`

tests/php/unit/BackgroundJob/CleanUpJobTest.php
```diff
@@ -0,0 +1,81 @@
...
+use OCA\Calendar\BackgroundJob\CleanUpJob;
+use OCA\Calendar\Service\Appointments\BookingService;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Calendar\ICalendarQuery;
+use OCP\ILogger;
+use PHPUnit\Framework\MockObject\MockObject;
+use Psr\Log\LoggerInterface;
+use Test\TestCase;
+
+/**
```
ChristophWurst, line 14: license header should be after `declare` and before the imports

ChristophWurst (changes requested):
Works!

appinfo/info.xml
```diff
@@ -36,6 +36,9 @@
 	<dependencies>
 		<nextcloud min-version="20" max-version="24" />
 	</dependencies>
+	<background-jobs>
+		<job>OCA\Calendar\BackgroundJob\CleanupJob</job>
```
tcitworld, line 40: Should be a better class/file name, to make it explicit that we're cleaning up outdated booking confirmations

lib/Controller/AppointmentConfigController.php
```diff
@@ -42,26 +43,36 @@ class AppointmentConfigController extends Controller {
...
 		$this->userId = $userId;
+		$this->logger = $logger;
 	}
 
 	/**
 	 * @NoAdminRequired
 	 */
 	public function index(): JsonResponse {
 		if ($this->userId === null) {
+			$this->logger->warning('User not found');
```
tcitworld, line 65: Log `userId` as well. :)

lib/Controller/AppointmentConfigController.php
```diff
@@ -71,16 +82,16 @@ public function index(): JsonResponse {
 	 */
 	public function show(int $id): JsonResponse {
 		if ($this->userId === null) {
+			$this->logger->warning('User not found');
```
tcitworld, line 85: Log `userId` as well.

lib/Controller/AppointmentConfigController.php
```diff
@@ -227,13 +241,16 @@ public function update(
 	 */
 	public function destroy(int $id): JsonResponse {
 		if ($this->userId === null) {
+			$this->logger->warning('User not found');
```
tcitworld, line 244: Log `userId` as well.

lib/Controller/BookingController.php
```diff
@@ -86,60 +113,134 @@ public function getBookableSlots(int $appointmentConfigId,
...
+			$this->logger->error($e->getMessage(), ['exception' => $e]);
+			return new TemplateResponse(
+				Application::APP_ID,
+				'appointments/404-booking',
+				[],
+				TemplateResponse::RENDER_AS_GUEST
+			);
+		}
+
+		$link = $this->urlGenerator->linkToRouteAbsolute( 'calendar.booking.getBookableSlots', [ 'appointmentConfigId' => $config->getId() ]);
```
tcitworld, line 222: Odd spacing before and after arguments

lib/Migration/Version2040Date20210908101001.php
```diff
@@ -22,79 +43,79 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		$schema = $schemaClosure();
 
 		$table = $schema->createTable('calendar_appt_configs');
-		$table->addColumn('id', Types::BIGINT, [
+		$table->addColumn('id', 'bigint', [
```
tcitworld, line 46: Why don't you use the `Types` constants anymore? You can remove `use OCP\DB\Types;` if you don't use it anymore.

lib/Migration/Version3000Date20211109132439.php
```diff
@@ -0,0 +1,100 @@
...
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Calendar\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
```
tcitworld, line 30: Not used

lib/Service/Appointments/BookingCalendarWriter.php
```diff
@@ -0,0 +1,142 @@
...
+		$vcalendar->VEVENT->add(
+			'ORGANIZER',
+			'mailto:' . $organizer->getEMailAddress(),
+			[
+				'CN' => $organizer->getDisplayName(),
+				'CUTYPE' => 'INDIVIDUAL',
+				'PARTSTAT' => 'ACCEPTED'
+			]
+		);
+		$vcalendar->VEVENT->add(
```
tcitworld, line 110: Can't remember why the organizer has to be added twice as an `ORGANIZER` and as an `ATTENDEE`. For email notifications ?

lib/Service/Appointments/MailService.php
```diff
@@ -0,0 +1,178 @@
...
+			$template->addBodyListItem($booking->getDescription(), $l10n->t('Description:'));
+		}
+	}
+
+	/**
+	 * @return string
+	 */
+	private function getSysEmail(): string {
+		$instanceName = $this->defaults->getName();
+		return \OCP\Util::getDefaultEmailAddress($instanceName);
```
tcitworld, line 176: Usually apps add something like `no-reply` in the user part. Look for uses of `getDefaultEmailAddress` in a whole Nextcloud setup.

lib/Migration/Version2040Date20210908101001.php
ChristophWurst: was added only for 21

lib/Controller/AppointmentConfigController.php
miaulalala: I can't if it's null :)

lib/Service/Appointments/BookingCalendarWriter.php
miaulalala: newline between function calls here?

lib/Service/Appointments/BookingCalendarWriter.php
miaulalala: because the organizer is not automatically also an attendee

lib/Service/Appointments/MailService.php
miaulalala: good idea!

ChristophWurst (approved):

lib/Controller/BookingController.php
```diff
@@ -86,60 +113,143 @@ public function getBookableSlots(int $appointmentConfigId,
...
 		try {
 			$config = $this->appointmentConfigService->findById($appointmentConfigId);
 		} catch (ServiceException $e) {
+			$this->logger->error('No appointment config found for id ' . $appointmentConfigId, ['exception' => $e]);
+			return JsonResponse::fail(null, Http::STATUS_NOT_FOUND);
+		}
+		try {
+			$booking = $this->bookingService->book($config, $start, $end, $timeZone, $displayName, $email, $description);
+		} catch (ClientException $e) {
+			$this->logger->warning('No slot available for start: ' . $start . ', end: ' . $end . ', config id: ' . $appointmentConfigId , ['exception' => $e]);
```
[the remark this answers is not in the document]
miaulalala: I added a NoSlotFoundException and changed the catch condition to listen to this. Client Exceptions are now caught generally the same way as ServiceExceptions.

lib/Controller/AppointmentConfigController.php
tcitworld: 🙈 I was probably looking somewhere else.

PR #3631, opened 2021-11-10
Fix appointments availability handling: start, end and time before next slot

ChristophWurst:
* Factor in *start time*
  * Do not let the user select times before that
  * Do not generate available slot before
* Factor in *end time*
  * Do not let the users select times after that
  * Do not generate available slots after
* Factor in *Minimum time before next slot*
  * Do not let the user select dates without a possible first bookable slot
  * Do not generate available slots before

`start` and `end` are missing from the UI. You have to generate timestamps and add them to the DB manually.


miaulalala (approved):
LGTM

PR #3635, opened 2021-11-11
Fix appointments daily max filter

ChristophWurst:
* Look into all of the principal's calendars
* Do the filtering for our special property in PHP

## How to test
* Create an appointment config with a daily limit of 1 appointment
* Create an event on an upcoming day. Mangle with the oc_calendar_objects data so this event has `X-NC-APPOINTMENT:xyz` where xyz is the token (check URL) of your appointment config
* Try to find slots on the day of that event vs on others

On master: slots were always available
With this patch: the whole day does not show any slots anymore

Requires https://github.com/nextcloud/server/pull/29660. This won't work with Nextcloud 23 RC1 therefore.


miaulalala (approved):
LGTM

PR #3638, opened 2021-11-11
Update npm and node engines versions

nextcloud-command:
Automated update of the npm and node engines versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,145 +1,145 @@
 {
-	"name": "calendar",
-	"description": "A calendar app for Nextcloud. Easily sync events from various devices, share and edit them online.",
-	"version": "3.4.0",
-	"author": "Georg Ehrke <oc.list@georgehrke.com>",
-	"contributors": [
-		"Georg Ehrke <oc.list@georgehrke.com>",
-		"Thomas Citharel <tcit@tcit.fr>"
-	],
-	"keywords": [
-		"nextcloud",
-		"calendars",
-		"vevent",
-		"caldav"
-	],
-	"bugs": {
-		"url": "https://github.com/nextcloud/calendar/issues"
-	},
-	"repository": {
-		"url": "https://github.com/nextcloud/calendar",
-		"type": "git"
-	},
-	"homepage": "https://github.com/nextcloud/calendar",
-	"license": "agpl",
-	"private": true,
-	"scripts": {
-		"build": "webpack --node-env production --progress",
-		"dev": "webpack --node-env development --progress",
-		"watch": "webpack --node-env development --progress --watch",
-		"lint": "eslint --ext .js,.vue src",
-		"lint:fix": "eslint --ext .js,.vue src --fix",
-		"stylelint": "stylelint src css",
-		"stylelint:fix": "stylelint src css --fix",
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

ChristophWurst (changes requested):
Actually, no. Our app also works with newer node

skjnldsv:
> Actually, no. Our app also works with newer node
@ChristophWurst 
It's not about this, it's about making sure we use the same environment across NC so we don't have to change versions every repo switch

tcitworld:
Why not use nvm instead ?

skjnldsv:
> Why not use nvm instead ?

I do, but even that is really annoying when you switch 20 repos a day.
It's the same, why not use yarn in some places?
Met's be consistent and professional so we're safe and in control of whatever new versions might appear where we'll have to do this whole process all over again.

ChristophWurst:
> It's not about this, it's about making sure we use the same environment across NC so we don't have to change versions every repo switch

Could you link us to the ticket where this has been discussed? What is the problem with newer node/npm?

skjnldsv:
> Could you link us to the ticket where this has been discussed? What is the problem with newer node/npm?

Discussed internally in TD and engineering channels. It was blowing some CI. We had this issue with npm 7 and we had to secure this, why are we debating again for the same topic we solved months ago?
But I don't even get the arguing against it, you're ok having such a wild selector you can leave it, I'm just helping, feel free to do as you please :shrug:

skjnldsv:
PS: everyone comes to be to complain when npm breaks, but don't want to listen when I give recommendations :confused:

tcitworld (approved):
All right, fine by me.

Is it acceptable if we bundle [an .nvmrc](https://github.com/nvm-sh/nvm#nvmrc) file as well for easy switching?

ChristophWurst:
> Discussed internally in TD and engineering channels. It was blowing some CI.
> We had this issue with npm 7 and we had to secure this, why are we debating again for the same topic we solved months ago?

My bad if I' missing something. If I remember correctly the issue was that the *server repo* didn't work with latest node/npm because of one dependency. In that case it makes sense to set hard requirements for the engines for that repo. However, I do not understand that every library and app that isn't effected by this also has to go into lockstep restrictions.

We are not restricting anything if we leave the engines open to newer version. You can still use your LTS node. It *will just work*. We will not limit you or anyone else.

**But**: if a contributor just wants to work on Calendar or similar component that isn't restricted by factual node engine problem, it doesn't make sense to greet them with tons of incompatible engine warnings when they could actually use their current setup without any issues.

So, I do agree that we should have a baseline of the minimum node/npm version that has to be supported. But I would certainly allow apps and libs to have a wider support range.

:v:

skjnldsv:
I disagree, we're on a stable direction.
npm and node are dependencies and should be treated as such. When a new version is available, we test against and if it works we allow the migration and usage.

Currently this selector is just prone to failure like it did for npm7. We're favouring fancy uncontrolled updates over stability

skjnldsv (approved):

ChristophWurst (approved):

PR #3695, opened 2021-11-24
Fix appointment availability structure

miaulalala:
Change the structure to account for more than one slot per day.

### DIFF
--- a/lib/Service/Appointments/AvailabilityGenerator.php
+++ b/lib/Service/Appointments/AvailabilityGenerator.php
@@ -27,7 +27,6 @@ namespace OCA\Calendar\Service\Appointments;
 
 use DateInterval;
 use DatePeriod;
-use DateTime;
 use DateTimeImmutable;
 use DateTimeZone;
 use JsonException;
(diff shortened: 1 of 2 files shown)


lib/Service/Appointments/AvailabilityGenerator.php
```diff
@@ -123,15 +122,17 @@ public function generate(AppointmentConfig $config,
...
+			if(empty($availabilitySlots)) {
 				$availabilityRules[$key] = [];
 				continue;
 			}
-			$availabilityRules[$key] = [
-				'start' => (new DateTimeImmutable())->setTimestamp($availabilitySlot['start']),
-				'end' => (new DateTimeImmutable())->setTimestamp($availabilitySlot['end'])
-			];
+			foreach($availabilitySlots as $slot) {
+				$availabilityRules[$key][] = [
```
ChristophWurst, line 131: still not a fan of using uninitialized array indices but since php allows this :+1:

ChristophWurst (approved):

PR #3697, opened 2021-11-25
Sort appointment configs by their name

ChristophWurst:


src/components/AppNavigation/AppointmentConfigList.vue
```diff
@@ -106,6 +106,17 @@ export default {
...
 		},
+		sortedConfigs() {
+			return [...this.configs].sort((config1, config2) => {
+				if (config1.name < config2.name) {
+					return -1
+				}
+				if (config1.name > config2.name) {
+					return 1
+				}
+				return 0
```
raimund-schluessler, line 117: I would propose to use `localCompare` here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
```suggestion
				return config1.name.localCompare(config2.name)
```

ChristophWurst: I knew something like that existed :D

st3iny (approved):

Issue #3604, opened 2021-11-05
Clean up appointment configs when the user gets deleted


ChristophWurst:
### Steps to reproduce
1. Create appointment configs
2. Delete the user
3. Check the database

### Expected behaviour

Appointment configs vanish with the user

### Actual behaviour

Appointment configs remain

Ref https://github.com/nextcloud/calendar/issues/3482

PR #3698, opened 2021-11-25
Delete appointment configs when the user is deleted

ChristophWurst:
Fixes https://github.com/nextcloud/calendar/issues/3604

To test create a new user on your dev instance, then create at least one appointment config. Delete the user after and check the database. There shouldn't be any configs of that user.

### DIFF
--- a/lib/Listener/UserDeletedListener.php
+++ b/lib/Listener/UserDeletedListener.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * @copyright 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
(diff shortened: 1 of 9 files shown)


lib/Listener/UserDeletedListener.php
```diff
@@ -0,0 +1,57 @@
...
+								LoggerInterface $logger) {
+		$this->appointmentConfigService = $appointmentConfigService;
+		$this->logger = $logger;
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof UserDeletedEvent)) {
+			return;
+		}
+
```
miaulalala, line 52: Also delete all booking confirmation links here so users can't confirm a booking that doesn't exist any more.

miaulalala (changes requested):

miaulalala (approved):

PR #3700, opened 2021-11-25
Fetch the appointment availability default from caldav, if possible

ChristophWurst:
Fixes https://github.com/nextcloud/calendar/issues/3694

PR #3701, opened 2021-11-25
Fix appointment availability time zone from caldav

ChristophWurst:
And add a fallback as well.

Bug introduced with https://github.com/nextcloud/calendar/pull/3700. The time zone was missing from the availability data structure and the validation did not allow creation of new configs.


miaulalala (approved):
LGTM

PR #3702, opened 2021-11-25
Allow longer time before the next possible appointment slot

ChristophWurst:
Because the previous max of 1h is very short notice.


GretaD (approved):
code looks ok

PR #3720, opened 2021-11-26
Do not allow shared calendars for appointments

ChristophWurst:
They don't work right now, so they have to be filtered out.


tcitworld (approved):

st3iny (commented):
Still shows editable shares in the picker.

Peek 2021-11-26 14-36

**EDIT:** I tried again and it worked.

Issue #3711, opened 2021-11-26
Default reminder setting ignored for appointments


ChristophWurst:
### Steps to reproduce
1. Set a default reminder
2. Create an appointment config
3. Have another user book and confirm an appointment

### Expected behavior

The event in my calendar has the default reminder setting applied

### Actual behavior

The event has no reminder.

### Calendar app
**Calendar app version:** v3

PR #3731, opened 2021-11-26
Add default reminders to appointments

ChristophWurst:
Fixes https://github.com/nextcloud/calendar/issues/3711

### DIFF
--- a/lib/Service/Appointments/BookingCalendarWriter.php
+++ b/lib/Service/Appointments/BookingCalendarWriter.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace OCA\Calendar\Service\Appointments;
 
 use DateTimeImmutable;
+use OCA\Calendar\AppInfo\Application;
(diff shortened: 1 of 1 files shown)


lib/Service/Appointments/BookingCalendarWriter.php
```diff
@@ -61,6 +63,22 @@ public function __construct(IConfig $config,
 		$this->random = $random;
 	}
 
+	private function secondsToIso8601Duration(int $secs): string {
```
tcitworld, line 66: I would have simply added seconds to a timestamp, then DateTime::diff, then something like https://stackoverflow.com/a/61088115

But this is small too and works. :)

tcitworld (approved):

lib/Service/Appointments/BookingCalendarWriter.php
ChristophWurst: There are so many solutions for this problem 🙈

PR #3745, opened 2021-11-29
Log debug infos of appointment slots

ChristophWurst:
The logic is a black box otherwise and impossible to debug in
production.

Example log

```json
{
  "reqId": "9lovcMHpqpTdxWVQQJlX",
  "level": 0,
  "time": "2021-11-29T13:02:32+00:00",
  "remoteAddr": "127.0.0.1",
  "user": "admin",
  "app": "calendar",
  "method": "GET",
  "url": "/apps/calendar/appointment/27/slots?startTime=1638190952&timeZone=Europe%2FVienna",
  "message": "Appointment config QByBoCW64gbs has 1 intervals that result in 11 possible slots. 11 slots remain after the daily limit. 11 available slots remain after conflict checking.",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0",
  "version": "24.0.0.0"
}
```

### DIFF
--- a/lib/Service/Appointments/BookingService.php
+++ b/lib/Service/Appointments/BookingService.php
@@ -40,6 +40,8 @@ use OCP\DB\Exception;
 use OCP\DB\Exception as DbException;
 use OCP\IUser;
 use OCP\Security\ISecureRandom;
+use Psr\Log\LoggerInterface;
+use function count;
 
 class BookingService {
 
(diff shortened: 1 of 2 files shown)


lib/Service/Appointments/BookingService.php
```diff
@@ -163,7 +170,16 @@ public function getAvailableSlots(AppointmentConfig $config, int $startTime, int
 		// 3. Filter out the daily limits
 		$filteredByDailyLimit = $this->dailyLimitFilter->filter($config, $allPossibleSlots);
 		// 4. Filter out booking conflicts
-		return $this->eventConflictFilter->filter($config, $filteredByDailyLimit);
+		$available = $this->eventConflictFilter->filter($config, $filteredByDailyLimit);
```
miaulalala, line 173: I would also add some logs inside the generators

miaulalala (changes requested):

miaulalala (approved):

PR #3773, opened 2021-12-01
Check for TRANSP and CANCELLED property when conflict checking appointment slots

miaulalala:


lib/Service/Appointments/EventConflictFilter.php
```diff
@@ -56,14 +60,31 @@ public function filter(AppointmentConfig $config, array $slots): array {
...
 		return array_filter($slots, function (Interval $slot) use ($followUpDuration, $preparationDuration, $query): bool {
 			$query->setTimerangeStart($slot->getStartAsObject()->sub($preparationDuration));
 			$query->setTimerangeEnd($slot->getEndAsObject()->add($followUpDuration));
 
-			$objects = $this->calendarManager->searchForPrincipal($query);
+			$events = $this->calendarManager->searchForPrincipal($query);
+			$objects = array_filter($events, function (array $event) {
+				$isCancelled = ($event['objects'][0]['STATUS'][0] ?? null) === 'CANCELLED';
+				$isFree = ($event['objects'][0]['TRANSP'][0] ?? null) === 'TRANSPARENT';
+				if(!$isCancelled && !$isFree) {
```
tcitworld, line 79: Add a variable before to avoid double condition check twice, here and in the return.
```php
$conflict = !$isCancelled && !$isFree;
```

lib/Service/Appointments/EventConflictFilter.php
```diff
@@ -56,14 +60,31 @@ public function filter(AppointmentConfig $config, array $slots): array {
 		}
 		// Always check the target calendar for conflicts
 		$query->addSearchCalendar($config->getTargetCalendarUri());
+		$query->addType('VEVENT');
 		$preparationDuration = DateInterval::createFromDateString($config->getPreparationDuration() . ' seconds');
 		$followUpDuration = DateInterval::createFromDateString($config->getFollowupDuration() . ' seconds');
-
+		$this->logger->debug('Appointment config ' . $config->getToken() . ' is looking with preparation duration {preparation} and follow up duration {followup} in calendar {calendarUri}', [
+			'preparation' => $preparationDuration->s,
```
tcitworld, line 67: It would be nice if we could have BookingCalendarWriter::secondsToIso8601Duration in a helper class and call it here instead of giving the number of seconds that don't make a lot of sense.

lib/Service/Appointments/EventConflictFilter.php
```diff
@@ -56,14 +60,31 @@ public function filter(AppointmentConfig $config, array $slots): array {
...
 			$query->setTimerangeStart($slot->getStartAsObject()->sub($preparationDuration));
 			$query->setTimerangeEnd($slot->getEndAsObject()->add($followUpDuration));
 
-			$objects = $this->calendarManager->searchForPrincipal($query);
+			$events = $this->calendarManager->searchForPrincipal($query);
+			$objects = array_filter($events, function (array $event) {
+				$isCancelled = ($event['objects'][0]['STATUS'][0] ?? null) === 'CANCELLED';
+				$isFree = ($event['objects'][0]['TRANSP'][0] ?? null) === 'TRANSPARENT';
+				if (!$isCancelled && !$isFree) {
+					$this->logger->debug('Appointment config conflicting with event with UID {uid}', [
```
ChristophWurst, line 80: could you move the log outside of this `array_filter`, log all UIDs at once (implode with `:` or similar) and include the time stamp of `$slot->getStartAsObject()->sub($preparationDuration)` and `$slot->getEndAsObject()->add($followUpDuration)`? Those timestamps will tell more than the preparation/followup second above because we'll know the precise window that is searched :)

ChristophWurst: Absolutely. I meant all *conflicting* UIDs :)

lib/Service/Appointments/EventConflictFilter.php
```diff
@@ -63,28 +65,30 @@ public function filter(AppointmentConfig $config, array $slots): array {
...
-		]);
-		return array_filter($slots, function (Interval $slot) use ($followUpDuration, $preparationDuration, $query): bool {
+		return array_filter($slots, function (Interval $slot) use ($followUpDuration, $preparationDuration, $query, $config): bool {
 			$query->setTimerangeStart($slot->getStartAsObject()->sub($preparationDuration));
 			$query->setTimerangeEnd($slot->getEndAsObject()->add($followUpDuration));
 
 			$events = $this->calendarManager->searchForPrincipal($query);
-			$objects = array_filter($events, function (array $event) {
+
+			$uids = '';
```
ChristophWurst, line 74: you can do better than this :wink: 

revert to the previous filtered array. `$object` contains all the conflict events. Array map them to their UID/id (or URI?) and implode the result.

Mutations are not our friends. Let's avoid them :)

miaulalala: Yes sorry I'm super tired 😅

ChristophWurst (approved):
Suuuper clean :shipit:

PR #3794, opened 2021-12-04
Fix SQL error when deleting a user

nickvergessen:
Fix https://github.com/nextcloud/server/issues/30094
Fix https://github.com/nextcloud/server/issues/30093


ChristophWurst (approved):

enoch85:
Thanks for a quick fix! <3

PR #3810, opened 2021-12-13
Add flex-wrap for better mobile design

### LINKED ISSUE #3808: Appointment booking mobilepage vertical scrolling
If i share an appointment Page, the datepicker and slots aren't vertical aligned, but horizontal aligned. On mobile Devices (e.g. iPhone 12 mini, IOS 15) only the first part is visible, the datepicker and timeslots aren't visible, only after horizontal scrolling. 

The three containers with user-info, date and slot selection should be vertically aligned and not horizontal. Could be fixed with flex-wrap.

Not sure if it helps, but the userstory would be:
I as an unexperienced user want to open the appointment booking page on a Phone, and see everything required to book an appointment without scrolling 

### Steps to reproduce
1. Share an Appointment booking Page
2. Open Page on Phone Browser
3. Don't see the second and third container

### Expected behaviour
All three containers should be visible

### Actual behaviour
Only the user-info Container is visible

### Calendar app
**Calendar app version:** newest

**CalDAV-clients used:** None, Browser

### Client configuration
**Browser:** Safari, Chrome (newest versions)

**Operating system:** iOS, MacOS

### Server configuration

**Operating system**: Debian 8

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.4.26

**Nextcloud Version:** 23.0.0

**Updated from an older installed version or fresh install:** Updated

### Logs
It's CSS Bug, so no logs

Ben-Ro:
Added "flex-warp: wrap;" to prevent unusable mobile design since flex-wrap default is nowrap which leads to horizontal scrolling on mobile devices

I hope this pull request is right here. It's my first pull request, so i'm sorry if it doesn't belongs here :)

Fixes #3808

### DIFF
--- a/src/views/Appointments/Booking.vue
+++ b/src/views/Appointments/Booking.vue
@@ -234,6 +234,7 @@ export default {
 	padding-top: 50px;
 	display: flex;
 	flex-direction: row;
+	flex-wrap: wrap;
 	max-width: 800px;
 
 	&__date-selection {


GretaD (approved):
nice improvement. Would be better tho to have the avatar in the middle when in mobile. but this already improves it

ChristophWurst:
@Ben-Ro please see https://github.com/nextcloud/calendar/pull/3810/checks?check_run_id=4504192130 for why CI fails

Ben-Ro:
> @Ben-Ro please see https://github.com/nextcloud/calendar/pull/3810/checks?check_run_id=4504192130 for why CI fails

Should be fixed now :)

PR #3831, opened 2021-12-20
adjust week view

szaimen:
This is best reviewed like this: https://github.com/nextcloud/calendar/pull/3831/files?diff=unified&w=1

### Goal:
**Making today better recognizeable on big screens**

| Before | After |
|---|---|
| image |image |
| image | image |

<details>

```
docker run -it \
-e CALENDAR_BRANCH=enh/noid/adjust-week-view \
-p 8443:443 \
--name nextcloud-easy-test \
```

</details>

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -60,6 +60,11 @@
 	color: var(--color-text-lighter) !important;
 }
 
+// Remove dotted half-lines
+.fc .fc-timegrid-slot-minor { 
+	border-top-style: none !important; 
+}
+
 // Center the date in month view
 .fc-daygrid-day-top {
 	justify-content: center;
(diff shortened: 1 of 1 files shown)


jancborchardt:
Nice – only 2 pieces of feedback:
- The color seems a bit too light / harsh. A color like the one of the buttons in the left nav would work better?
- The table header background (except the blue bubble ofc) could have the same background color.

tcitworld:
Do you mind adding screenshots of the white theme as well?

szaimen:
> Do you mind adding screenshots of the white theme as well?

I'll update the screenshots when I addressed the feedback from Jan :)

szaimen:
> * The color seems a bit too light / harsh. A color like the one of the buttons in the left nav would work better?

The problem is that the border lines of the grid have the same grey tone like the buttons in the left nav. So if we make the background this color, too, we loose the border lines, which looks very odd, imo.

I've for now adjusted the grid to be a bit brighter so that we can use the color nonetheless:
| Before | After |
|---|---|
| image | image |
| image | image |

What do you think? Which one is better? (regarding background of today and border lines)

szaimen:
This is ready now! Updated the screenshots in the PR-description :)

jancborchardt (approved):
Looks good design-wise, today is now much clearer. :)

GretaD (approved):
tested and works.

PR #3859, opened 2022-01-02
Change config to enable dev on Windows

dartcafe:
I would like to add a proposal for the dev configuration, so that it is possible to compile the js part on a windows machine. 

* use webpack's `--node-env` switch instead of `NODE_ENV=`
  produces error 'NODE_ENV not found'
* rename `webpack.js` to `webpack.config.js` (default config file)
  webpack.js gets openend due to the assotiation to the default .js extension
  (Possibly this is a configuration problem on my side)

Any concerns?


szaimen:
> * rename `webpack.js` to `webpack.config.js` (default config file)
>   webpack.js gets openend due to the assotiation to the default .js extension
>   (Possibly this is a configuration problem on my side)

You removed `--config webpack.js` ?

szaimen:
> * use webpacks `--node-env` switch instead of `NODE_ENV=`
>   produces error 'NODE_ENV not found'

Fine by me 👍

But should probably also be changed here https://github.com/nextcloud/documentation/blob/master/developer_manual/digging_deeper/npm.rst and here https://github.com/nextcloud/app-tutorial/blob/master/package.json

raimund-schluessler (changes requested):
Renaming `webpack.js` won't work like this, because the file is provided with this name by https://github.com/nextcloud/webpack-vue-config.

dartcafe:
> You removed `--config webpack.js` ?

webpack uses webpack.config.js by default.

dartcafe:
> Renaming `webpack.js` won't work like this, because the file is provided with this name by https://github.com/nextcloud/webpack-vue-config.

I don't get the point. Do you refer to the docs there or will there be a conflict upon renaming here?

raimund-schluessler:
> > Renaming `webpack.js` won't work like this, because the file is provided with this name by https://github.com/nextcloud/webpack-vue-config.
> 
> 
> 
> I don't get the point. Do you refer to the docs there or will there be a conflict upon renaming here?

Sorry, what I mean is that you should also adjust it in this upstream repository where the general webpack configuration for all of Nextcloud is located. Adjusting it only here for Calendar will lead to confusion and is against the goal of unification.

dartcafe:
@raimund-schluessler Ah, OK. I see. At least the readme has to be updated. I can add a PR there as well. But do you think, the webpack.js should be renamed as well, since it is no part of a build process there?

dartcafe:
See referenced repos. I am hungry now...

dartcafe:
Any further comments or concerns? 
Does the change have an impact on open branches?

szaimen (approved):

st3iny (approved):

PR #3917, opened 2022-01-27
Integrate with User Migration

Pytal:
TODO drop all code for testing purposes and integrate with user_migration app API


tcitworld:
This should be in the `dav` app instead.

It will be pretty much the same for the Tasks and Contacts app, so the common parts should be together.

Pytal:
Superseded by https://github.com/nextcloud/server/pull/30963

PR #3918, opened 2022-01-27
Fix: Inconsistencies due to ALTREP parameter

max65482:
This PR addresses issue #3863: In addition to the plain text description of an event, Thunderbird also saves a formatted HTML version inside the ALTREP parameter. NC calendar does not alter the ALTREP parameter when the plain text description is changed. This results in inconsistencies.

The proposed solution deletes the ALTREP parameter upon modification. This prevents inconsistencies. Thunderbird keeps accepting plaintext-only descriptions.

I will have opened a [similar PR](https://github.com/nextcloud/tasks/pull/1916) in nextcloud/tasks as it is equally concerned.

### DIFF
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -1527,6 +1527,18 @@ const actions = {
 		const calendarObject = state.calendarObject
 
 		if (eventComponent.isDirty()) {
+			// ALTREP parameter is not set by NC calendar, but by CalDAV clients like Thunderbird.
+			// To avoid inconsistencies, remove ALTREP parameter upon modification.
+			const veventComponent = calendarObject.calendarComponent.getFirstComponent('VEVENT')
+			if (veventComponent) {
+				const descriptionProperty = veventComponent.getFirstProperty('Description')
+				if (descriptionProperty) {
+					if (descriptionProperty.hasParameter('ALTREP')) {
+						descriptionProperty.deleteParameter('ALTREP')
+					}
+				}
+			}
+
 			const isForkedItem = eventComponent.primaryItem !== null
 			let original = null
 			let fork = null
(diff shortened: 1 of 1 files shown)


tcitworld:
I think this could be handled here instead, which would be cleaner.
https://github.com/nextcloud/calendar/blob/f39e31fec11708970f975dc83e49246decce95dd/src/store/calendarObjectInstance.js#L322-L325

max65482:
This method is called upon every keystroke in the description field. Deleting ALTREP here would cause the loss of formatting even if the user discards changes by closing the dialog. I would prefer to delete ALTREP only if the user explicitly clicks "Update".

max65482:
Sorry, I misunderstood how `changeDescription` works. The same behavior can be achieved there as well. I now moved the fix to this function.

st3iny (approved):
Thanks a lot for the fix. I tested it and it works fine.

Please sign off your commits before we can merge them. To add your Signed-off-by line to every commit in this branch:

1. Ensure you have a local copy of your branch by checking out the pull request locally via command line.
2. In your local branch, run: git rebase HEAD~2 --signoff
3. Force push your changes to overwrite the branch: git push --force-with-lease origin fix_altrep

max65482:
Perfect! The commits are signed off now.

tcitworld:
As [`DESCRIPTION` can have more parameters](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.1.5) than `ALTREP`, such as [`LANGUAGE`](https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.10) or vendors parameters (surely there must be some client that uses `X-VENDOR-DESC-HTML`), should we remove all of them instead of just `ALTREP`?

max65482:
I agree that it makes sense to remove all parameters that are not updated consistently (so, currently all of them).
I got the implementation ready. Let me know if I should submit it. And where (this PR?).

benbucksch:
@max65482 Thanks so much for the fix! :+1: 

> should we remove all of them instead of just ALTREP?

+1

This is also what I recommended in https://github.com/nextcloud/calendar/issues/3863#issuecomment-1007533913

max65482:
As this PR is merged and fixes the issue, I think we should consider it complete.
For the deletion of all parameters I opened #3924.

Issue #3905, opened 2022-01-22
ics import failure when VEVENT includes X-APPLE-STRUCTURED-LOCATION


flexibeast:
Calendar 3.0.5 on NC 22.2.3 failed to import an `.ics` file, sent from a recent iPhone, with the UI errors:

```
No valid files found, aborting import
```

and

```
file.ics could not be parsed
```

with the `nextcloud.log` error:

```
"Sabre\\VObject\\ParseException","Message":"This parser only supports VCARD and VCALENDAR files"
```

Bisecting the file revealed that the presence of the X-APPLE-STRUCTURED-LOCATION property within the VEVENT component was the cause; without this property, the file was imported successfully.

max65482:
I just successfully imported an `.ics` file that includes one `X-APPLE-STRUCTURED-LOCATION` property:

```
X-APPLE-STRUCTURED-LOCATION;VALUE=URI;X-ADDRESS="Berlin, Deutschland";X-AP
 PLE-MAPKIT-HANDLE=CAESkAEIrk0Q0ozqrpusuYyyARoSCZ3dtd81Q0pAESBumLIdvSpAIikK
 C0RldXRzY2hsYW5kEgJERRoGQmVybGluKgZCZXJsaW4yBkJlcmxpbioTQmVybGluIEhhdXB0Ym
 FobmhvZjIGQmVybGluMgtEZXV0c2NobGFuZDgvUAFaEwoRCNKM6q6brLmMsgEYrk2QAwE=;X-A
 PPLE-RADIUS=301.0391264594954;X-APPLE-REFERENCEFRAME=1;X-TITLE=Berlin Haup
 tbahnhof:13.369367
```

There are some reports about colon/semicolon mixups and missing escape characters with apple devices:
https://github.com/collective/icalendar/issues/116
https://discussions.apple.com/thread/250761617

Could you share a sample property that causes the error?

max65482:
What Apple is doing is "interesting", anyway. Properties that start with `X-` are for "experimental use" according to [RFC5545](https://datatracker.ietf.org/doc/html/rfc5545#section-3.1). Logically, `X-APPLE-STRUCTURED-LOCATION` supplements the regular `LOCATION` property.

Think of the following situation (similar to #3863): You create an event with an Apple device (includes `X-APPLE-STRUCTURED-LOCATION` and `LOCATION`) . Then you modify the location with a non-Apple device. Only `LOCATION` is updated. The event properties are inconsistent.

[...]

PR #3930, opened 2022-02-03
Fix: Inconsistencies due to Apple's additional location property

max65482:
Similar to #3924, this PR fixes an inconsistency issue. This time it is about the additional location information that Apple applications store in  the `X-APPLE-STRUCTURED-LOCATION` property. This is a supplement to the regular `LOCATION` property. It includes coordinates and a custom map reference. Upon change of `LOCATION` in a non-Apple application, `X-APPLE-STRUCTURED-LOCATION` becomes invalid. This PR solves this inconsistency by removing the `X-APPLE-STRUCTURED-LOCATION` property upon change of the location.

Discussion can be found in #3905.

### DIFF
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -307,6 +307,9 @@ const mutations = {
 	 * @param {string} data.location New location to set
 	 */
 	changeLocation(state, { calendarObjectInstance, location }) {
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Makes sense

tcitworld (approved):

Issue #3993, opened 2022-02-24
Ability to remove certain menu options on Calendars entries


AndyScherzinger:
### Is your feature request related to a problem? Please describe.

On Shared environments notifications make less sense than on dedicated and with shared environment/devices Download confuses users since there is no use for that in general since everything is done "online"

### Describe the solution you'd like

Ability to remove some menu items in Calendar app to avoid user's confusion (i.e., the functions that not applicable an certain/shared environments). 

1. Remove the "Download" button inside Event's Details:

![image1](https://user-images.githubusercontent.com/1315170/151061505-037ccc02-ab5e-4bfe-b275-0fb0c9cf4394.jpeg)

2. Remove the "Notification" and set "Email" by default in the Event details->Reminder (added a reminder first):

![image2](https://user-images.githubusercontent.com/1315170/151061519-b7b981bb-7b0c-407b-a5c0-ee214e130996.jpeg)

---

> Remove the "Notification" and set "Email" by default in the Event details

Design-wise, both of these could be removed. The method of reminder should be a user setting to be put in the "Notifications" user settings, which could be set to mail, push, neither or both. Default should probably be both, cause mail covers the folks which don’t have Nextcloud open all day every day:

![image](https://user-images.githubusercontent.com/925062/151171010-65e94f7d-d553-4907-8168-10f5eff71ad1.png)

> Remove the "Download" button inside Event's Details

Can understand that, even for us this is more of an advanced "Export" function which is not so important?

### Describe alternatives you've considered

no

### Additional context

no

szaimen:
Could this ability be somehow wrapped into a config setting? For the normal user are these options very useful...

AndyScherzinger:
> Could this ability be somehow wrapped into a config setting? For the normal user are these options very useful...

Yes, that is the idea 👍

tcitworld:
Can you explain more what's shared environments so that I can understand more what are the encountered issues ?

I think renaming the Download buttons to Export makes sense in any case.

On the topic of reminders, some remarks:
* This would apply only to events created through the NC calendar app, not from 3rd-party clients that may have some other reminder defaults settings (or none). Might add confusion if those are not the same than the new global settings.
* Having both email and notification providers enabled as a default seems fine to me, apart from when the reminder is sent ~10 minutes before the event and we've just sent an useless email you'll probably only see during or after the event (email being sent before the start of the event doesn't mean you'll receive it straight away, contrarily to a push notification or your phone's own calendar system).
[...]

PR #3995, opened 2022-02-25
Allow admins to hide event download buttons

ChristophWurst:
This contributes the first part of https://github.com/nextcloud/calendar/issues/3993.

The patch adds a simple config switch. There are shared environments where Nextcloud is only used on web and users are not supposed to export/download event data to a local disk.

I will update the admin calendar docs accordingly. The config switch can be done via CLI with

```shell
php -f occ config:app:set calendar hide_event_download --value=yes
```


raimund-schluessler:
Maybe I don't understand it, but how does that help without a change of server? The url is still accessible and serves the data.

lib/Controller/ViewController.php
```diff
@@ -93,6 +93,7 @@ public function index():TemplateResponse {
 		$slotDuration = $this->config->getUserValue($this->userId, $this->appName, 'slotDuration', $defaultSlotDuration);
 		$defaultReminder = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminder', $defaultDefaultReminder);
 		$showTasks = $this->config->getUserValue($this->userId, $this->appName, 'showTasks', $defaultShowTasks) === 'yes';
+		$hideEventDownload = $this->config->getAppValue($this->appName, 'hide_event_download', 'no') === 'yes';
```
tcitworld, line 96: Other config values are using camelCase, so let's be consistent.

tcitworld (changes requested):

ChristophWurst:
> Maybe I don't understand it, but how does that help without a change of server? The url is still accessible and serves the data.

That is correct from a technical PoV. This is really about what options we give the users explicitly. The Files app has a similar feature. You can hide the download button, while technically files could still be acquired. It's :shrug: but also :see_no_evil: to put it into emojis :)

ChristophWurst:
> Maybe I don't understand it, but how does that help without a change of server? The url is still accessible and serves the data.

https://github.com/nextcloud/calendar/issues/3993#issuecomment-1050840836 for more context :)

GretaD (approved):
looks good

PR #3996, opened 2022-02-25
Add PHP8.1 support

ChristophWurst:
Ref https://github.com/nextcloud/groupware/issues/27

* [x] Update info.xml requirements
* [x] Update composer.json requirements
* [x] Update test matrix
* [x] Update lint matrix
* [x] Fix the code

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.2|^8.0"
+        "php": ">=7.3 <=8.1"
     },
     "require-dev": {
         "christophwurst/nextcloud_testing": "0.12.4",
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -6,7 +6,7 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.2|^8.0"
+        "php": ">=7.2 <=8.1"
```
tcitworld, line 9: Is that useful since `config.platform.php` is at 7.3 ?

ChristophWurst: I think composer still checks during installation

tcitworld: I meant for 7.2 compatibility.

PR #3997, opened 2022-02-25
Optionally enforce a event alarm type

ChristophWurst:
This contributes the second part of https://github.com/nextcloud/calendar/issues/3993.

* Admins can enforce `DISPLAY` or `EMAIL` as alarm type
* The value is used as new default for an reminder
* If a reminder matches the enforced type, we hide the radio group all together (there would only be one option after all)
  Bildschirmfoto von 2022-02-25 11-49-12
* If a reminder doesn't match the enforced type (event created before enforcement or another client) then the user decide to stay on the old value or switch to the enforced one
  Bildschirmfoto von 2022-02-25 11-49-05

I will update the Calendar admin docs for this.

```shell
occ config:app:set calendar forceEventAlarm_type --value=EMAIL
```

### DIFF
--- a/lib/Controller/ViewController.php
+++ b/lib/Controller/ViewController.php
@@ -30,6 +30,7 @@ use OCP\AppFramework\Http\TemplateResponse;
 use OCP\AppFramework\Services\IInitialState;
 use OCP\IConfig;
 use OCP\IRequest;
+use function in_array;
 
 class ViewController extends Controller {
 
@@ -94,6 +95,10 @@ class ViewController extends Controller {
 		$defaultReminder = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminder', $defaultDefaultReminder);
 		$showTasks = $this->config->getUserValue($this->userId, $this->appName, 'showTasks', $defaultShowTasks) === 'yes';
 		$hideEventExport = $this->config->getAppValue($this->appName, 'hideEventExport', 'no') === 'yes';
+		$forceEventAlarmType = $this->config->getAppValue($this->appName, 'forceEventAlarmType', '');
+		if (!in_array($forceEventAlarmType, ['DISPLAY', 'EMAIL'], true)) {
+			$forceEventAlarmType = null;
+		}
 
 		$talkEnabled = $this->appManager->isEnabledForUser('spreed');
 		$talkApiVersion = version_compare($this->appManager->getAppVersion('spreed'), '12.0.0', '>=') ? 'v4' : 'v1';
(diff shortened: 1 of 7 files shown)


tcitworld:
> Admins can enforce DISPLAY or EMAIL as alarm type

Not both, then ?

lib/Controller/ViewController.php
```diff
@@ -93,6 +94,10 @@ public function index():TemplateResponse {
 		$slotDuration = $this->config->getUserValue($this->userId, $this->appName, 'slotDuration', $defaultSlotDuration);
 		$defaultReminder = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminder', $defaultDefaultReminder);
 		$showTasks = $this->config->getUserValue($this->userId, $this->appName, 'showTasks', $defaultShowTasks) === 'yes';
+		$forceEventAlarmType = $this->config->getAppValue($this->appName, 'force_event_alarm_type', '');
```
tcitworld, line 97: camelCase as well

tcitworld (changes requested):

ChristophWurst:
> Not both, then ?

That would be equal to no enforcement. Right now you can also only pick between EMAIL/DISPLAY and nothing else unless that was already present in the event.

GretaD (approved):

max65482:
By default I cannot select the reminder type anymore (no config switch set). Is this the expected behavior?

Otherwise, I would have expected this condition...
https://github.com/nextcloud/calendar/blob/bba3cd5c97ab164e7af196600200208004f5b5d0/src/components/Editor/Alarm/AlarmListItem.vue#L269-L271

... to be
`return this.forceEventAlarmType === null || this.alarm.type !== this.forceEventAlarmType `

lib/Controller/ViewController.php
```diff
@@ -114,6 +119,7 @@ public function index():TemplateResponse {
 		$this->initialStateService->provideInitialState('show_tasks', $showTasks);
 		$this->initialStateService->provideInitialState('tasks_enabled', $tasksEnabled);
 		$this->initialStateService->provideInitialState('hide_event_export', $hideEventExport);
+		$this->initialStateService->provideInitialState('force_event_alarm_type', $forceEventAlarmType);
```
szaimen, line 122: Based om the docs is null not allowed as key:
https://github.com/nextcloud/server/blob/3e067edcbe80797a70118f613ac4b752ce512992/lib/public/AppFramework/Services/IInitialState.php#L45

max65482: Yes, as mentioned in #4051, this causes log entries.

miaulalala:
Not sure if your occ command is intentionally using `forceEventAlarm_type` but the initial state checks for `forceEventAlarmType` just FYI

pm-pm:
Was hab ich den Kaffee auf! == What I have the coffee on!

It took me about 1 hour to understand the problem a little, then find a workaround. The one above.

It then took me 3.5 hours to figure out that there was a typo in the workaround! I'm not a software expert so it's hard to find that! Since the underscore also occurs again and again somewhere else the error was not so obvious!!!
So I just copied and pasted the workaround.

WRONG!! "occ config:app:set calendar forceEventAlarm_**type** --value=EMAIL"

CORRECT!! "occ config:app:set calendar forceEventAlarm**Type** --value=EMAIL"

Please correct in original Posting on top!

pm-pm

PS:
If you follow the link, you may understand how the typo came about. But it doesn't bring back the lost time:
https://github.com/nextcloud/calendar/blob/main/lib/Controller/ViewController.php#L122


> I will update the Calendar admin docs for this.
> 
> ```shell
> occ config:app:set calendar forceEventAlarm_type --value=EMAIL
> ```

PR #4010, opened 2022-03-01
Use CircleOutline for disabled calendar icon

### LINKED ISSUE #4005: Use CheckboxBlankCircleOutline if calendar is disabled

After https://github.com/nextcloud/calendar/pull/3960 when a calendar is disabled, only the color of the icon-bullet is changed to `var(--color-text-lighter)`. This makes it not obvious that the calendar is disabled, because it's just a different color. If you happen to have calendars with a similar color, or have a color-impaired vision, you are a bit lost.
See here for example:
Screenshot 2022-03-01 at 09-52-36 Woche 9 aus 2022 - Kalender - Nextcloud

Use a different icon for disabled calendars: `CheckboxBlankCircleOutline` here
https://github.com/nextcloud/calendar/blob/b9576fcab2e0d1f20c974128f673057a1a9dc976/src/components/AppNavigation/CalendarList/CalendarListItem.vue#L33-L36

This would solve the problem:
Screenshot 2022-03-01 at 09-54-22 Woche 9 aus 2022 - Kalender - Nextcloud

_No response_

### Additional context

_No response_

@JuliaKirschenheuter @nimishavijay For the design here.

raimund-schluessler:
This PR changes the icon for disabled calendars to `CheckboxBlankCircleOutline` so it can be better distinguished from enabled calendars. Since the color is not needed anymore as a difference, I figured we can keep it the same, even if the calendar is disabled. This is how it looks now:
2022 - Calendar - Nextcloud

Compared to before:
Screenshot 2022-03-01 at 20-34-45 Week 10 of 2022 - Calendar - Nextcloud

@jancborchardt @nimishavijay In case you prefer to still have disabled calendars using `var(--color-text-lighter)` instead of the calendar color, let me know, and I will change it.

Closes #4005.

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListItem.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListItem.vue
@@ -31,9 +31,15 @@
 				<ActionButton @click.prevent.stop="toggleEnabled">
 					<template #icon>
 						<CheckboxBlankCircle
-							:title="calendar.enabled ? $t('calendar', 'Disable calendar') : $t('calendar', 'Enable calendar') "
+							v-if="calendar.enabled"
+							:title="$t('calendar', 'Disable calendar')"
 							:size="20"
-							:fill-color="calendar.enabled ? calendar.color : 'var(--color-text-lighter)'" />
+							:fill-color="calendar.color" />
+						<CheckboxBlankCircleOutline
+							v-else
+							:title="$t('calendar', 'Enable calendar')"
+							:size="20"
+							fill-color="var(--color-text-lighter)" />
 					</template>
 				</ActionButton>
 			</Actions>
@@ -189,6 +195,7 @@ import CalendarListItemSharingSearch from './CalendarListItemSharingSearch.vue'
 import CalendarListItemSharingPublishItem from './CalendarListItemSharingPublishItem.vue'
 import CalendarListItemSharingShareItem from './CalendarListItemSharingShareItem.vue'
 import CheckboxBlankCircle from 'vue-material-design-icons/CheckboxBlankCircle.vue'
+import CheckboxBlankCircleOutline from 'vue-material-design-icons/CheckboxBlankCircleOutline.vue'
 import Close from 'vue-material-design-icons/Close.vue'
 import Delete from 'vue-material-design-icons/Delete.vue'
 import Download from 'vue-material-design-icons/Download.vue'
@@ -211,6 +218,7 @@ export default {
 		CalendarListItemSharingPublishItem,
 		CalendarListItemSharingShareItem,
 		CheckboxBlankCircle,
+		CheckboxBlankCircleOutline,
 		Close,
 		Delete,
 		Download,
--- a/src/components/AppNavigation/CalendarList/CalendarListItem.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListItem.vue
@@ -39,7 +39,7 @@
 							v-else
 							:title="$t('calendar', 'Enable calendar')"
 							:size="20"
-							fill-color="var(--color-text-lighter)" />
+							:fill-color="calendar.color" />
 					</template>
 				</ActionButton>
 			</Actions>


raimund-schluessler:
Ping @nextcloud/designers

skjnldsv (approved):

nimishavijay:
Nice work! Looks much nicer with the outline!

> In case you prefer to still have disabled calendars using var(--color-text-lighter) instead of the calendar color, let me know, and I will change it.

I'd go with this, since the calendar is disabled it makes sense to keep it grey :)

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -31,9 +31,15 @@
...
+							v-if="calendar.enabled"
+							:title="$t('calendar', 'Disable calendar')"
 							:size="20"
-							:fill-color="calendar.enabled ? calendar.color : 'var(--color-text-lighter)'" />
+							:fill-color="calendar.color" />
+						<CheckboxBlankCircleOutline
+							v-else
+							:title="$t('calendar', 'Enable calendar')"
+							:size="20"
+							:fill-color="calendar.color" />
```
raimund-schluessler, line 42:
```suggestion
							fill-color="var(--color-text-lighter)" />
```

raimund-schluessler:
> Nice work! Looks much nicer with the outline!
> 
> > In case you prefer to still have disabled calendars using var(--color-text-lighter) instead of the calendar color, let me know, and I will change it.
> 
> I'd go with this, since the calendar is disabled it makes sense to keep it grey :)


@nimishavijay Hm, I liked having the color still visible, especially since it is not really guaranteed that the color of disabled calendars (Gray here) is unique with respect to enabled calendars. Anyway, here it is with `--color-text-lighter`, see the adjusted screenshot in https://github.com/nextcloud/calendar/pull/4010#issue-1155733487.

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -31,9 +31,15 @@
...
+							v-if="calendar.enabled"
+							:title="$t('calendar', 'Disable calendar')"
 							:size="20"
-							:fill-color="calendar.enabled ? calendar.color : 'var(--color-text-lighter)'" />
+							:fill-color="calendar.color" />
+						<CheckboxBlankCircleOutline
+							v-else
+							:title="$t('calendar', 'Enable calendar')"
+							:size="20"
+							fill-color="var(--color-text-lighter)" />
```
raimund-schluessler, line 42: @jancborchardt This would set the color to the calendars color.
```suggestion
							:fill-color="calendar.color" />
```

JuliaKirschenheuter (approved):

Ornanovitch:
> @nimishavijay Hm, I liked having the color still visible, especially since it is not really guaranteed that the color of disabled calendars (Gray here) is unique with respect to enabled calendars. But anyway, here it is with `--color-text-lighter`, see the adjusted screenshot in [#4010 (comment)](https://github.com/nextcloud/calendar/pull/4010#issue-1155733487).

Hey calendar teem and thanks for your work. I was following this PR as I find it really nice. Just wanted to say that IMO the outline style is largely enough to symbolize a disabled item in comparison with a fully filled circle. With this improvement, it could be worth to keep the color as the text is already lighter and as you can more quickly eye-track the calendar you want to enable when you have a lot of.

raimund-schluessler:
> > @nimishavijay Hm, I liked having the color still visible, especially since it is not really guaranteed that the color of disabled calendars (Gray here) is unique with respect to enabled calendars. But anyway, here it is with `--color-text-lighter`, see the adjusted screenshot in [#4010 (comment)](https://github.com/nextcloud/calendar/pull/4010#issue-1155733487).
> 
> Hey calendar teem and thanks for your work. I was following this PR as I find it really nice. Just wanted to say that IMO the outline style is largely enough to symbolize a disabled item in comparison with a fully filled circle. With this improvement, it could be worth to keep the color as the text is already lighter and as you can more quickly eye-track the calendar you want to enable when you have a lot of.

Yes, this is what we aggreed on now. Thanks for the feedback.

PR #4045, opened 2022-03-11
Remove Action from CheckboxBlankCircle and add more description text

### LINKED ISSUE #3955: Calendar visibility toggle (entry itself) is not focusable via keyboard

JuliaKirschenheuter:
fixes #3955 

- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/2548

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListItem.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListItem.vue
@@ -22,24 +22,17 @@
 <template>
 	<AppNavigationItem v-click-outside="closeShareMenu"
 		:loading="calendar.loading"
+		:aria-description="descriptionAppNavigationItem"
 		:title="calendar.displayName || $t('calendar', 'Untitled calendar')"
 		:class="{deleted: !!deleteTimeout, disabled: !calendar.enabled, 'open-sharing': shareMenuOpen}"
 		@click.prevent.stop="toggleEnabled">
 		<template slot="icon">
-			<Actions>
-				<ActionButton @click.prevent.stop="toggleEnabled">
-					<template #icon>
-						<CheckboxBlankCircle v-if="calendar.enabled"
-							:title="$t('calendar', 'Disable calendar')"
-							:size="20"
-							:fill-color="calendar.color" />
-						<CheckboxBlankCircleOutline v-else
-							:title="$t('calendar', 'Enable calendar')"
-							:size="20"
-							:fill-color="calendar.color" />
-					</template>
-				</ActionButton>
-			</Actions>
+			<CheckboxBlankCircle v-if="calendar.enabled"
+				:size="20"
+				:fill-color="calendar.color" />
+			<CheckboxBlankCircleOutline v-else
+				:size="20"
+				:fill-color="calendar.color" />
 		</template>
 
 		<template v-if="!deleteTimeout" slot="counter">
@@ -326,6 +319,22 @@ export default {
 			}
 
 			return ''
+		},
+		/**
+		 * compute aria-description for AppNavigationItem link
+		 *
+		 * @return {string}
+		 */
+		descriptionAppNavigationItem() {
+			if (this.calendar.enabled && this.calendar.displayName) {
+				return t('calendar', 'Disable calendar "{calendar}"', { calendar: this.calendar.displayName })
+			} else if (this.calendar.enabled && !this.calendar.displayName) {
+				return t('calendar', 'Disable untitled calendar')
+			} else if (!this.calendar.enabled && this.calendar.displayName) {
+				return t('calendar', 'Enable calendar "{calendar}"', { calendar: this.calendar.displayName })
+			} else {
+				return t('calendar', 'Enable untitled calendar')
+			}
 		},
 	},
 	methods: {


src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -27,22 +27,16 @@
...
-							v-else
-							:title="$t('calendar', 'Enable calendar')"
-							:size="20"
-							:fill-color="calendar.color" />
-					</template>
-				</ActionButton>
-			</Actions>
+			<CheckboxBlankCircle
+				v-if="calendar.enabled"
+				:title="$t('calendar', `Disable ${calendar.displayName.toLowerCase()} calendar`)"
```
raimund-schluessler, line 32: This won't work with the translations. You have to use it like this (untested code):
```suggestion
				:title="$t('calendar', 'Disable calendar {calendar}', {calendar: calendar.displayName})"
```

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -27,22 +27,16 @@
...
+				:size="20"
+				:fill-color="calendar.color" />
+			<CheckboxBlankCircleOutline
+				v-else
+				:title="`Enable ${calendar.displayName.toLowerCase()} calendar`"
```
raimund-schluessler, line 37: Translation missing, and see above.

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -27,22 +27,16 @@
 		:class="{deleted: !!deleteTimeout, disabled: !calendar.enabled, 'open-sharing': shareMenuOpen}"
 		@click.prevent.stop="toggleEnabled">
 		<template slot="icon">
-			<Actions>
-				<ActionButton @click.prevent.stop="toggleEnabled">
```
raimund-schluessler, line 31: There is now no way to toggleEnabled anymore. It is missing on the Icons.

raimund-schluessler (changes requested):
I have to admit, I don't get the problem this solves, but this anyway needs changes to work. Please see my comments.

src/components/AppNavigation/CalendarList/CalendarListItem.vue
raimund-schluessler: See
https://github.com/nextcloud/calendar/blob/0467c6e5571a9346b33756da43fea050b30c5f59/src/views/Calendar.vue#L279
for an example how to properly replace a variable in a translated string.

JuliaKirschenheuter: @raimund-schluessler Thank you!

src/components/AppNavigation/CalendarList/CalendarListItem.vue
raimund-schluessler: Ah, nevermind, you are right. I missed the @click handler on the `AppNavigationItem`. Sorry.

raimund-schluessler (approved):
Looks good code wise.

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -23,26 +23,22 @@
 	<AppNavigationItem
 		v-click-outside="closeShareMenu"
 		:loading="calendar.loading"
+		:aria-description=" calendar.enabled ? $t('calendar', 'Disable {calendar} calendar', {calendar: calendar.displayName.toLowerCase()}) :
```
raimund-schluessler, line 26: I would like to propose changing the strings a little bit. Especially in German making the name lowercase will not work well. E.g. a calendar named `Feiertage` (holidays) will produce the string `Deaktiviere feiertage Kalendar`, which does not make sense.
I would propose to use `$t('calendar', 'Disable calendar "{calendar}"', {calendar: calendar.displayName})` which will give `Deaktiviere Kalendar "Feiertage"`.

JuliaKirschenheuter:
> I have to admit, I don't really understand why this shouldn't be an Action, but this anyway needs changes to work. Please see my comments.

After review with @jancborchardt we have decided not to to have a same action on two different components: AppNavigationItem and CheckboxBlankCircle. User click on AppNavigationItem -> toggle will be triggered, and it is enough.

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -342,6 +335,16 @@ export default {
...
 			return ''
 		},
+	  /**
+	   * compute aria-description for AppNavigationItem link
+	   *
+	   * @return {string}
+	   */
+	  descriptionAppNavigationItem() {
+			return this.calendar.enabled
+				? t('calendar', 'Disable calendar "{calendar}"', { calendar: this.calendar.displayName })
```
tcitworld, line 345: `this.calendar.displayName` can be undefined, see how `calendar.displayName || $t('calendar', 'Untitled calendar')` is used above.

tcitworld (changes requested):

src/components/AppNavigation/CalendarList/CalendarListItem.vue
```diff
@@ -342,6 +335,17 @@ export default {
 
 			return ''
 		},
+	  /**
+	   * compute aria-description for AppNavigationItem link
+	   *
+	   * @return {string}
+	   */
+	  descriptionAppNavigationItem() {
+		  const untitledCalendar = t('calendar', '{calendar}', { calendar: 'Untitled calendar' })
```
tcitworld, line 344: I'm not sure what you're doing there, but the translation for 'Untitled calendar' won't be used that way.

In any case, it's hard to assume the order of the words will stay the same, it's possible there's a language where the position of the calendar name depends if the calendar has a real name or not. Imagine a situation like :
* `Disable calendar "MyCalendar"` :arrow_right: `"Disable calendar "MyCalendar"` :heavy_check_mark: 
* `Disable calendar "Untitled Calendar"` :arrow_right: `"Untitled Calendar" Disable calendar` :no_entry_sign: 

It's better to have all 4 cases, even though it's long:
* `Disable calendar "{calendar}"`
* `Disable untitled calendar`
* `Enable calendar "{calendar}"`
* `Enable untitled calendar`

tcitworld (changes requested):

raimund-schluessler (approved):

tcitworld (approved):

ChristophWurst:
Marking as draft to prevent an accidental merge before main ships `@nextcloud/vue` with https://github.com/nextcloud/nextcloud-vue/pull/2548.

JuliaKirschenheuter:
> Marking as draft to prevent an accidental merge before main ships `@nextcloud/vue` with [nextcloud/nextcloud-vue#2548](https://github.com/nextcloud/nextcloud-vue/pull/2548).

i think we can merge it: https://github.com/nextcloud/nextcloud-vue/releases/tag/v4.4.0

ChristophWurst (approved):

PR #3994, opened 2022-02-24
Show attendee icon on event and reenable notification icon too

tcitworld:
image

image

Closes #3162

Issue #1758, opened 2020-01-06
Add CONFERENCE property to events linked to Talk rooms


tcitworld:
> I tried that, there is a problem with the encoding of FEATURE param in ical.js
> Have to fix / report that upstream. :)

_Originally posted by @georgehrke in https://github.com/nextcloud/calendar/pull/1732#issuecomment-566555305_

Luncheon3462:
Nextcloud talk is flaky for so many users.   Would the calendar team consider enhancing the app to allow for the automatic creation of jitsi links for a new calendar event.

keunes:
Hi @Luncheon3462

> Would the calendar team consider enhancing the app to allow for the automatic creation of jitsi links for a new calendar event.

Note that this feature request is about something rather different. So if you want your thing you should probably create your own feature request.

PR #4085, opened 2022-03-18
[WIP] Move event rendering to vue

tcitworld:
Full calendar has a slot named `eventContent` that we can use to render the event.

Previously the rendering was handled using classes in `src/fullcalendar/eventSources/eventSourceFunction.js` and `src/fullcalendar/rendering/eventDidMount.js`.

- [x] Show the bell icon when the event has alarms
- [ ] Add the attendee icon from #3994
- [ ] Show the checkbox icon when the event is a task
   - [ ] Adapt to the view (check logic in `eventDidMount.js`)
   - [ ] Adapt to the task being done
- [ ] Add the location and description content in the event in the list view
- [ ] Set the event text color to `eventDetails.textColor` and background to `eventDetails.backgroundColor`
- [ ] Adapt the CSS to make the icons look right
- [ ] Remove the addition of now unneeded classes in both `eventSourceFunction.js` and `eventDidMount.js` files (some need to be kept, such as `fc-event-nc-cancelled`)
- [ ] Clear the corresponding classes from the SCSS code

Addresses #3162 and replaces #3994

### DIFF
--- a/src/components/CalendarGrid.vue
+++ b/src/components/CalendarGrid.vue
@@ -23,7 +23,11 @@
 <template>
 	<FullCalendar
 		ref="fullCalendar"
-		:options="options" />
+		:options="options">
+		<template #eventContent="arg">
+			<EventRender :event-details="arg" />
+		</template>
+	</FullCalendar>
 </template>
 
 <script>
@@ -56,6 +60,7 @@ import dayHeaderDidMount from '../fullcalendar/rendering/dayHeaderDidMount.js'
 import eventDidMount from '../fullcalendar/rendering/eventDidMount.js'
 import eventOrder from '../fullcalendar/rendering/eventOrder.js'
 import noEventsDidMount from '../fullcalendar/rendering/noEventsDidMount.js'
+import EventRender from './EventRender.vue'
 
 // Import timezone plugins
 import VTimezoneNamedTimezone from '../fullcalendar/timezones/vtimezoneNamedTimezoneImpl.js'
@@ -71,6 +76,7 @@ export default {
 	name: 'CalendarGrid',
 	components: {
 		FullCalendar,
+	  EventRender,
 	},
 	props: {
 		/**
--- a/src/components/EventRender.vue
+++ b/src/components/EventRender.vue
@@ -1,0 +1,53 @@
+<!--
+  - @copyright Copyright (c) 2022 Thomas Citharel <nextcloud@tcit.fr>
+  -
+  - @author Thomas Citharel <nextcloud@tcit.fr>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+<template>
+	<div>
+		<Bell v-if="hasAlarms" />
+		<b>{{ eventDetails.timeText }}</b>
+		<span class="fc-event-title">{{ eventDetails.event.title }}</span>
+	</div>
+</template>
+
+<script>
+import Bell from 'vue-material-design-icons/Bell.vue'
+
+export default {
+	name: 'EventRender',
+	components: {
+		Bell,
+	},
+	props: {
+		eventDetails: {
+			type: Object,
+			required: true,
+		},
+	},
+	computed: {
+		viewType() {
+			return this.eventDetails?.view?.type
+		},
+		hasAlarms() {
+			return this.eventDetails?.event?._def?.extendedProps?.hasAlarms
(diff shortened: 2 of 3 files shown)


ChristophWurst:
This will also come in handy for https://github.com/nextcloud/calendar/issues/1758 :rocket:

JuliaKirschenheuter:
@nimishavijay @jancborchardt  i have redesigned a bit events in calendar, please check it out. Some examples are: 

image
image
image
image

Please give some feedback)

nimishavijay:
Looks great! 

> Question: should we show a task event in this way:

If the icon for a task is a checkbox then I would want to click on it. If it is possible we should definitely try to make an interactive checkbox that we can click on and mark the task as completed. 
If that is not possible then we could just use another icon. We could use the icon for Tasks itself which looks like this so people would know that it is a task
image

tcitworld:
> If the icon for a task is a checkbox then I would want to click on it. If it is possible we should definitely try to make an interactive checkbox that we can click on and mark the task as completed.

The current behavior is to open the task in question into the task app.

Alex-decode:
> Question: should we show a task event in this way:
> or in this way? 

The second way. With the presence of a checkbox it is more clear that this is a task.

raimund-schluessler:
> > If the icon for a task is a checkbox then I would want to click on it. If it is possible we should definitely try to make an interactive checkbox that we can click on and mark the task as completed.
> 
> 
> 
> The current behavior is to open the task in question into the task app.

When we anyway do the rendering in Vue now, it should be possible to toggle the completed state when you click the checkbox and open the task in the Tasks app if you click somewhere else, no? Could be done as a further enhancement.

nimishavijay:
> When we anyway do the rendering in Vue now, it should be possible to toggle the completed state when you click the checkbox and open the task in the Tasks app if you click somewhere else, no? Could be done as a further enhancement.

If this is possible then we should do it this way. I will open a new issue about it

jancborchardt:
Very cool, only small things:
- [ ] The time shouldn't be bold as it's not more important than the event name. Actually making it `color-text-maxcontrast` might be better
- [ ] The icons also seem very present, I'd say less opacity, like we do for menu icons in the left navigation
- [ ] For long events, the content is vertically centered in the box which looks a bit strange. Would look better top-aligned

ChristophWurst:
@JuliaKirschenheuter let's finish this. If it helps lower the complexity of this scope and the discussion, we can split the refactoring and the redesign part.

1) Refactor: move the rending to Vue but have no visual changes
2) Redesign: ask the design colleagues for specs and then implement them

Right now it's a mix of both and the constant feedback loop leads to stagnation.

Is that ok for everyone?

JuliaKirschenheuter:
> If it helps lower the complexity of this scope and the discussion, we can split the refactoring and the redesign part.
You mean, that we should merge & close this one and then move on with 2 separate tickets: the refactoring and the redesign?

I think the refactoring could be done here https://github.com/nextcloud/calendar/issues/4113

In my opinion the redesign part could be done here, that we have a clean implementation design-wise (improve only things, which were discussed above).

ChristophWurst:
> I think the refactoring could be done here #4113

That is also a feature.

Let's do the refactoring in isolation and the two features will be done later.

JuliaKirschenheuter:
@tcitworld sorry for rewriting your commit, i had to do it for the final one.
I've removed all new style changes from this PR and put them all to https://github.com/nextcloud/calendar/issues/4235

JuliaKirschenheuter:
I have fixed tests.
I have done a reminder icon via material icons, because it was not possible anymore to use icons via class (see https://github.com/nextcloud/server/issues/32072). 

Now it seems like this

image
image

There will be some design issues (design bugs), they could be fixed within https://github.com/nextcloud/calendar/issues/4235. Unfortunately it was not possible to create this PR completely without visual changes.

Please let me know, if we could close this issue and merge "[Move event rendering to vue](https://github.com/nextcloud/calendar/pull/4085#)"

src/components/EventRender.vue
```diff
@@ -0,0 +1,74 @@
...
+	  Bell,
+	},
+	props: {
+		eventDetails: {
+			type: Object,
+			required: true,
+		},
+	},
+	computed: {
+		viewType() {
```
ChristophWurst, line 56: indentation

JuliaKirschenheuter: tabulation? in a EventRender.vue looks right

src/components/CalendarGrid.vue
```diff
@@ -71,6 +76,7 @@ export default {
 	name: 'CalendarGrid',
 	components: {
 		FullCalendar,
+	  EventRender,
```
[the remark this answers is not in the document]
ChristophWurst: Check https://github.com/nextcloud/calendar/blob/7716e0193673ae4aba3094c79a7bda420aaa4cd1/src/components/CalendarGrid.vue#L78-L79. L78 uses tab, L79 uses two spaces

src/components/EventRender.vue
ChristophWurst: See https://github.com/nextcloud/calendar/blob/7716e0193673ae4aba3094c79a7bda420aaa4cd1/src/components/EventRender.vue#L55-L65

JuliaKirschenheuter: i don't know what is wrong here with tabulation, on my local machine is everything right

Jerome-Herbinet:
Hello @ChristophWurst,
Not sure to understand … are the reminder bells supposed to appear since this PR has been merged ... or not ?

st3iny:
@Jerome-Herbinet This PR was reverted due to multiple problems caused by it. See https://github.com/nextcloud/calendar/pull/4351.

PR #4105, opened 2022-03-24
add ok-button to datetime picker

### LINKED ISSUE #2936: Implicit Action on Datetimepicker
When editing an event, either combination of time&date, the dialog can only be closed by clicking outside the dialog.

I'd like to propose an intentional way of closing this dialog, via a save button.

**Additional context**
I think a similar topic was discussed in #899
I agree that it is good to not discard changes when the user clicks somewhere else, but an action that changes something should always have an explicit trigger (like an OK button)
Only having implicit triggers, like closing or clicking somewhere else, does not communicate to the user when something is going to happen. Imho, a clear way to save would be better, as to prevent confusion. (I find myself quite often switching between date&time because i expect a save button in the lower right)

newhinton:
This closes #2936.

The goal is to create a more "guided" user experience by providing explicit actions to confirm an action

### DIFF
--- a/src/components/Shared/DatePicker.vue
+++ b/src/components/Shared/DatePicker.vue
@@ -39,7 +39,8 @@
 		v-on="$listeners"
 		@close="close"
 		@change="change"
-		@pick="pickDate">
+		@pick="pickDate"
+		confirm>
 		<template #icon-calendar>
 			<button class="datetime-picker-inline-icon icon"
 				:class="{'icon-timezone': !isAllDay, 'icon-new-calendar': isAllDay, 'datetime-picker-inline-icon--highlighted': highlightTimezone}"


szaimen:
@newhinton thanks for this PR! Could you please fix the CI? Thank you! :)

newhinton:
I have found out how the contacts-app does it. However: This comes with a drawback.

Clicking outside now dismisses the picker, without changing the date.
Selecting a date/time does not save it. The user now HAS to click the confirm button.

While they present a drastic change to the current behaviour, i still think this is the way forward, to keep consistency and a clear functionality which the current implementation does not provide.

szaimen:
> I have found out how the contacts-app does it. However: This comes with a drawback.
> 
> Clicking outside now dismisses the picker, without changing the date. Selecting a date/time does not save it. The user now HAS to click the confirm button.
> 
> While they present a drastic change to the current behaviour, i still think this is the way forward, to keep consistency and a clear functionality which the current implementation does not provide.

cc @jancborchardt and @nimishavijay on that

jancborchardt:
I'd say that's ok. @nimishavijay what do you think?

nimishavijay:
Yep this sounds like a good idea! Suggestion: we could use "Done" for the wording in the button.

newhinton:
I dont really know if that is easily possible. Even the `confirm` element is not documented in the nextcloud/vue project, and i have not yet found a way to access the `confirmText` value in the underlying datetimepicker library. Maybe we need to first expose this in the vue project?

szaimen (approved):
I like it and it works :)
| image | image |

szaimen:
@newhinton can you please resolve DCO? Afterwards we can merge this, imo :)

newhinton:
@szaimen Yes i will fix that! But i wanted to add a 5px margin between the buttons beforehand if thats fine with you

szaimen:
> But i wanted to add a 5px margin between the buttons beforehand if thats fine with you

Of course it is fine, please go ahead :)

newhinton:
it's a 4 pixel margin now, to keep it centered. However, as you can see in your own screenshot, we may have to adjust the popup width so that the controls always fit inside the popup. My margin has worsened it. Should i try to fix it here aswell?

nimishavijay:
Looks good to me! I have one important point though: Considering the date/time will be selected only after you click "OK", the button should really be in the primary button style, i.e `--color-primary` background with `--color-primary-text` text. Currently it looks more like a disabled button.

szaimen:
> Looks good to me! I have one important point though: Considering the date/time will be selected only after you click "OK", the button should really be in the primary button style, i.e `--color-primary` background with `--color-primary-text` text. Currently it looks more like a disabled button.

In general this should probably be adressed in the vue repo but @newhinton do you think we are able to overwrite the color somehow?

newhinton:
@szaimen probably. Though maybe we should just fix it in the vue repo, so that the contacts app (and everyone else) benefits aswell. 

I think i will create a pr for the vue repo, and not touch it here if it is okay, (that includes the misaligned text&margins, though we already have one now. Should i remove that then?)

szaimen:
> Though maybe we should just fix it in the vue repo, so that the contacts app (and everyone else) benefits aswell.
> 
> I think i will create a pr for the vue repo, and not touch it here if it is okay. (that includes the misaligned text&margins, though we already have one now. Should i remove that then?)

Sounds good to me!
@nimishavijay also fine to you?

st3iny (changes requested):
Tested and works. Please squash your commits into one before we can merge.

newhinton:
I have removed the margin for now.

PR #4132, opened 2022-04-03
Improve the trashbin layout

szaimen:
## Desktop

| Before | After |
|---|---|
| image | image |

## Mobile

| Before | After |
|---|---|
| image | image |

<details>

```
docker run -it \
-e CALENDAR_BRANCH=enh/noid/improve-trashbin-modal-size \
-p 8443:443 \
-e TRUSTED_DOMAIN=192.168.146.130 \
--name nextcloud-easy-test \
```

</details>

### DIFF
--- a/src/components/AppNavigation/CalendarList/Trashbin.vue
+++ b/src/components/AppNavigation/CalendarList/Trashbin.vue
@@ -28,6 +28,7 @@
 		</template>
 		<template #extra>
 			<Modal v-if="showModal"
+				size="large"
 				@close="showModal = false">
 				<div class="modal__content">
 					<h2>{{ t('calendar', 'Trash bin') }}</h2>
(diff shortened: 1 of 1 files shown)


GretaD:
I think Jan should have a look at this.

I have only one comment, the name. the time and button should be aligned, right now the time its a bit low compare to name of the event

jancborchardt:
Could we move the "deleted" time into the subline as well? Then it's:

> **Nextcloud Contributor Week planning meeting**
> Calendar name · 9 April 2022 · deleted yesterday

Much simpler and it already reduces width.
Then for the rest I'd say the event name can be ellipsized instead of making the modal large?

st3iny (approved):
Tested and works. Mobile version is not perfect but that is a topic for another time.

jancborchardt:
@szaimen did you address my comment about moving from table layout to subline? Otherwise could you put what’s left in an issue so it can be picked up later? :) Thanks!

szaimen:
Done in https://github.com/nextcloud/calendar/issues/4369

PR #4154, opened 2022-04-14
Update main target versions

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


tcitworld (approved):
We can refactor it better later

PR #4183, opened 2022-04-29
fix sidebar scrolling

szaimen:
This bug is as far as I could see only present on master but I might be wrong.


szaimen:
Unfortunately this issue now exists in the 3.3.0 release. So it just got pretty important!

szaimen:
Thanks Christoph! :)

PR #4198, opened 2022-05-05
Compare lower case email when checking invitees

artonge:
When comparing email to check if it is present in the invitees list, we do not normalize the value.
This leads to inconsistencies:
1. the user will see two options when searching for Alice@exampe.com if the stored email is alice@example.com
2. the user will see one option when searching for alice@exampe.com if the stored email is alice@example.com

1 | 2
-- | --
Screenshot from 2022-05-05 16-19-37 | Screenshot from 2022-05-05 16-20-40

This PR transform the values to lower cases to consolidate the number of options offered to the user.


tcitworld (approved):

ChristophWurst:
Is this a feature or a fix? Backport to the v3.3 line?

artonge:
Fix, yes backport could be nice.

Issue #298, opened 2017-01-17
[question] Is there attachment option


somenxavier:
Is there an option to add an attachments to calendar events? If not, please put it as enhancement.

georgehrke:
Other CalDAV clients like thunderbird can already add attachments. The web calendar can't. Definitely something we should consider for the future.
Instead of writing the attachment into the actual calendar data I would suggest to upload it to the nextcloud and just link there.

abditus:
I  agree. Attachments would be a great feature. Syncing by CalDAV should include the files.

georgehrke:
@abditus please refrain from commenting when you can't add anything new. Use [Github Reactions](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) instead.

DavidVentura:
Is this on a roadmap / milestone / anything ? Having link support (like thunderbird) would be enough

georgehrke:
@DavidVentura I have no intention to implement this soon. There are way more important features missing. 

But as always: Pull requests welcome :)

szaimen:
Any update on this?

tcitworld:
Was giving a few thoughts to this issue recently.

* Should the editor offer the possibility to pick an existing file?
* When uploading a new file, should the editor let the user pick where to upload the file (or to a fixed `.calendar-attachments` hidden folder)?
* Since the event may be shared/exported to people without access to the file, should the link to the file always be a public link? Or do we add a « public link » checkbox/toggle?

We can probably reuse the files app file picker to make things easier.

@georgehrke @jancborchardt

georgehrke:
> - Should the editor offer the possibility to pick an existing file?
> - When uploading a new file, should the editor let the user pick where to upload the file (or to a fixed .calendar-attachments hidden folder)?

As a first iteration of implementing attachments, i would not allow uploading new documents at all. Only picking existing files from Nextcloud for now.

Later on we can add the possibility to upload new files. There is a related discussion going on in Talk right now, see https://github.com/nextcloud/spreed/issues/1811#issuecomment-505096933

> Since the event may be shared/exported to people without access to the file, should the link to the file always be a public link? Or do we add a « public link » checkbox/toggle?

I think the link should always be public, but I'm not sure if we need a new share-type for that in the server. This can be a bit tricky. Even when you remove the attachment from a desktop or mobile client, the public link should automatically be unshared.

[...]

PR #4251, opened 2022-06-01
Add attachments to events

sazanof:
Fix https://github.com/nextcloud/calendar/issues/298

Signed-off-by: Mikhail Sazanov <m@sazanof.ru>

I tried to implement attachment files in events. The parameter is described here https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.1.1

Probably it will be a draft, although in fact everything works.

I decided to do this: there are several ways to add attachments:
1) From the Files application. Then a common link is formed (here I think to make it possible to provide a link not shared, but closed. but so far I have done so).
2) File upload. The file is uploaded to the root directory and a public link is formed to it

Probably here you can use a separate folder "Calendar". And in the future, the place for attachments may grow, but it is more convenient for users to download from a computer, rather than choose from "Files".

I also plan to delete the file itself when deleting the event (if the owner of the event does this action), if it is not needed in the repository.

Well, the loading indication (start - end) it should also be added.

Please tell me: where is the email notification template generated when an event notification arrives (I want to add attachments section)?

I would be grateful if you would take a look and give your comments. Thanks!=)

image
empty

image
with files

image
file options


ChristophWurst:
This PR addresses https://github.com/nextcloud/calendar/issues/298

tcitworld:
> Please tell me: where is the email notification template generated when an event notification arrives (I want to add attachments section)?

This is handled by the server: https://github.com/nextcloud/server/blob/master/apps/dav/lib/CalDAV/Schedule/IMipPlugin.php

src/services/attachmentService.js
```diff
@@ -0,0 +1,96 @@
...
+	let promises = [];
+	
+	files.forEach(file => {
+		if(componentAttachments.map(attachment => attachment.fileName).indexOf(file.name) !== -1){
+			showError(t('calendar', 'Attachment {fileName} already exists!',{
+				fileName: file.name
+			}))
+		}
+		else {
+			const url = `/remote.php/dav/files/${dav.userId}/${file.name}`
```
tcitworld, line 67: There's the brand new `nextcloud-upload` package from @skjnldsv that might be a better fit: https://github.com/nextcloud/nextcloud-upload.

src/store/calendarObjectInstance.js
```diff
@@ -1350,6 +1350,65 @@ const mutations = {
...
+
+		calendarObjectInstance.eventComponent.addProperty(attachment)
+		calendarObjectInstance.attachments.push(attachment)
+		console.log('addAttachmentBySharedData', attachment)
+	},
+
+	addAttachmentByLocalFile(state, { calendarObjectInstance, file }){
+		
+		const attachment = AttachmentProperty.fromLink(file.url)
+		// hot-fix needed temporary, becase calendar-js has no fileName get-setter
```
tcitworld, line 1373: Can you open a PR there? https://github.com/nextcloud/calendar-js

src/components/Editor/Attachments/AttachmentsList.vue
```diff
@@ -0,0 +1,222 @@
...
+			style="display:none"
+			@change="onLocalAttachmentSelected">
+		<div v-if="attachments.length > 0">
+			<ul>
+				<ListItem v-for="attachment in attachments"
+					:key="attachment.value"
+					:force-display-actions="true"
+					:title="attachment.fileName">
+					<template #icon>
+						<Paperclip :size="24" />
```
tcitworld, line 15: As an improvement, we *should* be able to show the file preview here instead.

tcitworld:
> I also plan to delete the file itself when deleting the event (if the owner of the event does this action), if it is not needed in the repository.

Note : this will only affect events deleted through the web UI.

In the future we need to implement CalDAV-Managed Attachments ([RFC 8607](https://datatracker.ietf.org/doc/html/rfc8607)) https://github.com/sabre-io/dav/issues/837 and let the server handle everything instead. We have time though, I'm not aware of any client implementing this (at least not [Evolution](https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/125), KOrganizer or even Thunderbird - this should interest @devvv4ever btw)

src/components/Editor/Attachments/AttachmentsList.vue
sazanof: @tcitworld  That would be really great! However, I don't really know =( yet how to display file previews (it would also be relevant for downloading attachments in the mail application).  

I know that links like " are generated http://nextcloud.dev.loc/index.php/core/preview?fileId=607478&x=250&y=250&forceIcon=0&a=0"  

I read the description here, but there was a question of generating office documents. 

https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/previews_configuration.html 

**I would be grateful for a hint in this direction.**

tcitworld: I did not find a cleaner way than https://github.com/nextcloud/viewer/blob/43e7003b3095bba553d71f59e73aa8ec1dbdff29/src/mixins/PreviewUrl.js#L57-L82

We should have this exacted in a new package.

sazanof: **Thank you very much. I will apply this method.**

That is, do I understand correctly that if I don't have a `collabora + nextcloud office` bundle, then I get previews only for pictures, mp3 and text files? There is also no way to get a preview at the time of download. 

That's all I'm getting at: I probably need to replace the missing images with "something" (file icons, for example).
BTW
There are still such addresses.
`/index.php/apps/theming/img/core/filetypes/x-office-document.svg`
Probably I should also look for how icons are generated according to the mime type.

sazanof: https://github.com/nextcloud/server/blob/fd5f7486da8192b6ebf69ee3e6640cb7662028ca/apps/files/src/views/Sidebar.vue#L310-L337
As I understand it, so far there is no single way (or component) to get a preview or icon

sazanof: In general, I had to use `MANAGED-ID` and `split()` magic =)
Perhaps there are better solutions. But I can only see this one so far.
image

_In general, the topic of the preview is quite ticklish_

tcitworld:
> I decided to do this: there are several ways to add attachments:
>  1. From the Files application. Then a common link is formed (here I think to make it possible to provide a link not shared, but closed. but so far I have done so).
>  2. File upload. The file is uploaded to the root directory and a public link is formed to it
>
> Probably here you can use a separate folder "Calendar". And in the future, the place for attachments may grow, but it is more convenient for users to download from a computer, rather than choose from "Files".

I'm not fond of uploading directly to the user root folder, it might get crowded real quick.

One possibility would be to handle this in the app backend (it would be irrelevant once [RFC 8607](https://datatracker.ietf.org/doc/html/rfc8607) would be implemented in dav server - see above) and put files in user's appdata. However the users wouldn't have access to files apart from accessing them directly from the event.

In the case of Talk the app creates a `Talk` folder in the user root to put attachments. Is that a pattern we should reproduce here with a `Calendar` folder ? Sounds the best to me. Opinions @nextcloud/designers @nickvergessen

Later and with the help of file tracking for RFC 8607 we can improve these files discoverability just like for Talk with a « all files attached to events in this calendar » section.

tcitworld:
@sazanof Tests have to be updated for the new `attachments` property.

src/services/attachmentService.js
sazanof: > @sazanof Tests have to be updated for the new `attachments` property.

Tell me, is it enough to add a parameter or do I need to write a full test? If it's a complete test, then I ask for help, since **I've never written tests and don't understand how to write them** (It's a matter of time and effort, I don't argue). =( 
I tried to write my own block for checking (by analogy with other tests) `it()`
This error occurs if I add my parameters to 
```
expect(getDefaultEvent Object()).to Equal({
....
attachments: [
	'ATTACHMENT1',
	'ATTACHMENT2',
	'ATTACHMENT3',
	'ATTACHMENT4',
	]
})
```
https://gist.github.com/sazanof/ecb2aaa8c760f93e74d2952294ba09ba

When checking, I get an error. I can't understand why it occurs...
```
 ● Test suite failed to run

    TypeError: Converting circular structure to JSON
        --> --> --> --> --> --> --> --> starting at object with constructor 'CalendarComponent'
        --- property '_root' closes the circle
        at stringify (<anonymous>)

```

sazanof:
> @sazanof Tests have to be updated for the new `attachments` property.

Tell me, is it enough to add a parameter or do I need to write a full test? If it's a complete test, then I ask for help, since I've never written tests and don't understand how to write them (It's a matter of time and effort, I don't argue). =( 
I tried to write my own block for checking (by analogy with other tests) `it()`
This error occurs if I add my parameters to 
```
expect(getDefaultEvent Object()).to Equal({
....
attachments: [
	'ATTACHMENT1',
	'ATTACHMENT2',
	'ATTACHMENT3',
	'ATTACHMENT4',
	]
})
```
https://gist.github.com/sazanof/ecb2aaa8c760f93e74d2952294ba09ba

When checking, I get an error. I can't understand why it occurs...
```
 ● Test suite failed to run

    TypeError: Converting circular structure to JSON
        --> --> --> --> --> --> --> --> starting at object with constructor 'CalendarComponent'
        --- property '_root' closes the circle
        at stringify (<anonymous>)
```

devvv4ever:
@tcitworld Thank you for thinking on us here too :) We would definitely love to implement attachment sync in DAVx⁵! Managed Attachments would be a really really cool feature also on Android. The problem here is that the Android Calendar storage does not offer attachments storage capabilities. So we'd need to find other apps that can provide an interface for it. We need a collaboration with others then, too (etar, acalendar+, jtx board, etc) because DAVx⁵ is not a calendar app itself so that people can also access their attachments then on Android.

nickvergessen:
I think it shouldn't be in the root anymore. I think Deck and etc. use an Attachment/ root folder

src/services/attachmentService.js
skjnldsv: Yes, clearly.

sazanof: @skjnldsv **Well, since we are not in a hurry, we can discuss =).** 

I would like to be able to upload to a specific folder + have a callback to provide shared access or something else. Maybe a subscription to some events (loading before, progress, after) ...
By buttons - you need a universal button that can be entered into any standard NC component. Ex.: ListItem,ActionButton, Button and so on. 

It would also be cool to be able to set the template of the uploaded file (before uploading, at the moment and after).

If you make the component universal, then in the future consider using it to upload not only to DAV, but also when writing a letter in the mail, for example...

nimishavijay:
Dunno if the frontend is ready for reviewing but I have some design feedback that could be considered, feel free to incorporate it :) The general flow looks really great! The only feedback is regarding wording and icons.

In the "empty" screenshot:
- The empty content icon could be the [`create-new-folder` MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%3Acreate_new_folder%3A) or just the [`folder` MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%3Afolder%3A)
- Color of the empty content icon should be `var(color-text-maxcontrast)`
- "Choose" --> "Add from Files"
- Upload icon --> [`file-upload` MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%3Afile_upload%3A)

In the "with files" screenshot:
- Instead of a 3dot menu icon that opens a menu with "Choose" and "Upload" we could make it just another list item with an [`add` icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aadd%3A) and text "New attachment"
- If it is possible to show a preview thumbnail we could do that
- If it is possible to display a filetype icon we could do that 
- Currently the subline shows the type of file it is, but that is already understood from the extension in the name of the file. Instead we could show the size of the file (for eg. 300 KB)
- "Choose" --> "Add from Files"
- Upload icon --> [`file-upload` MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%3Afile_upload%3A)

In the "file options" screenshot:
- Not sure if this is already the case but clicking on the file should open it in the viewer if possible. If not it could be downloaded.
- Hence we would not need the "View file" option in the action menu. It could be changed to "Download". Icon stays the same
- "Delete file" --> "Delete"
- Delete icon --> [`delete` MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%3Adelete%3A)

Super nice work! cc @jancborchardt

sazanof:
@nimishavijay Hello!

Thank you for the detailed answer and your wishes. I will definitely take them into account when updating the design. 

Regarding the preview for attachments: I have already implemented this, I just closed the discussion, and this branch is no longer shown. My joint.=) Here is a link to this comment 

https://github.com/nextcloud/calendar/pull/4251#discussion_r889991161    

file options actions, in fact, will not be needed if you open a file by clicking on its name and add only one delete button to the right of it. 

But what if we need to introduce a document preview function? Preview in the modal window, download and delete.  Of course it won't be soon, and then you can fix the design, but still. 

Thanks again for your comments!

jancborchardt:
Good stuff! Agree with @nimishavijay's feedback, just 2 points/questions:
- To keep keeping it simple, we can put the attachment section on the bottom of "Details" instead of adding a separate "Attachments" section. Details is visible by default.
- As to where the Files are stored: Regardless of how we do it, it would mean that only 1 copy exists and collaboration is possible, right? Or will files uploaded from the cloud be duplicated?

nimishavijay:
Just checked Google on how attachments are handled and they are uploaded into the root directory of your Google Drive, and the attachment basically links to that file. If you email another person the the event, they are able to see the attachment and if they click on it, it opens the file in Drive. 

cc @jancborchardt

sazanof:
@nimishavijay, @jancborchardt 
I tried to take into account all your comments and added a little of my own. What came out of it - you can see in the screenshots and https://github.com/nextcloud/calendar/pull/4251/commits/a4e764c82cb4774c9ffb4a2f2b742bccd7a53e44.

image
_no attachments_

image
_with attachments_

What do you say?

jancborchardt:
@sazanof really cool! :) Some additional feedback:
- [ ] "Upload from PC" → better "Upload from device" since people might be on tablet or mobile
- [ ] The preview needs some border radius (like in Files), `var(--border-radius)`
- [ ] The subline is too technical and too much information, it can just be cut

jancborchardt (changes requested):
@sazanof really cool! :) Some additional feedback:
- [x] "Upload from PC" → better "Upload from device" since people might be on tablet or mobile
- [x] The preview needs some border radius (like in Files), `var(--border-radius)`
- [ ] The subline is too technical and too much information, it can just be cut

jancborchardt:
@sazanof could you add a screenshot for easier design review? Thank you! :)

sazanof:
> @sazanof could you add a screenshot for easier design review? Thank you! :)

No problem =)
image

jancborchardt:
@sazanof really nice! Only 2 things actually:
- Better than the trash can icon would be the simple remove "x", cause otherwise people who added files from the cloud could misunderstand it as deletion from there. But it's just a removal. :)
- The indentation of the text is a bit further in than the text below, of xourse cause of the nice thumbnails. Not sure if anything needs to be done about it actually.

What do you think @nimishavijay @marcoambrosini?

IgorA100:
When I create an event from Thunderbird and attach an attachment as a link to a document stored in NC, when I go to the web version of NC, I see the created event, but the link I added is not displayed anywhere.
I consider this a problem that is highly desirable to resolve.
I would be very grateful!

nimishavijay:
Looks good once you replace the trash icon :) 🚀

IgorA100:
> PS I really want attachments in the calendar. =)

Super great!
> PSPS @IgorA100 Please send the source of the event

I don't quite understand what I need to do :(

sazanof:
> > PS I really want attachments in the calendar. =)
> 
> Super great!
> 
> > PSPS @IgorA100 Please send the source of the event
> 
> I don't quite understand what I need to do :(

an .ics file of an event

sazanof:
@jancborchardt @nimishavijay 

https://github.com/nextcloud/calendar/pull/4251/commits/3c62c0ef1ed8ca8dc8fe209d380594f892ac268f

image

What is next? =)

nimishavijay (approved):
🚀🚀🚀 :)

IgorA100:
> What is next? =)

Show size and change time? :)

sazanof:
@IgorA100 
А вы точно используете версию NC с моими изменениями? Потому что если нет, то ваши пропавшие вложения вполне объяснимы =)

Are you sure you're using the NC version with my changes? Because if not, then your missing attachments are quite understandable =)

IgorA100:
> NC с моими изменениями?

Вероятно нет, т.к. я ставил только стандартно через приложения в NC, а там крайняя версия 3.4.1 и в ней вообще нет функции аттача.
Мне нужно собрать из исходников версию "Release 3.4.2-RC1", верно?

sazanof:
contact me via telegram @Sazanof   

and these comments can be marked as offtop or deleted , since they are not related to PR

src/store/calendarObjectInstance.js
```diff
@@ -1351,6 +1351,61 @@ const mutations = {
...
 		}
 	},
+
+	addAttachmentBySharedData(state, { calendarObjectInstance, sharedData }) {
+		const attachment = AttachmentProperty.fromLink(sharedData.url)
+		const fileName = sharedData.path
+
+		// hot-fix needed temporary, becase calendar-js has no fileName get-setter
+		const parameterFileName = new Parameter('FILENAME', fileName)
+		const managedId = new Parameter('MANAGED-ID', `${sharedData.file_source}||${sharedData.has_preview}`)
```
st3iny, line 1361: Custom 
```suggestion
		const managedId = new Parameter('X-MANAGED-ID', `${sharedData.file_source}||${sharedData.has_preview}`)
```

Non-standard parameters should be prefixed with `X-`.

st3iny (changes requested):
I tested it locally and it works great so far. Huge thanks for implementing this feature!

However, there is a big issue regarding privacy: All attachments are shared publicly by default. Even when I don't share the event with anyone.

For private events, the attachments should not be shared at all (only visible if owner is logged in). When participants are invited we can create a public share + link **but** we need to communicate this to the user.

E.g. Google Calendar explicitly asks for my permission if I invite someone to an event with attachments:

ksnip_20220712-161609

## Translation:
A user requires access to your file

[REDACTED]

**Enable link share**
Everyone with the link will be able to **see** the file

- [ ] Deny access 

[**Cancel**] [**Invite**]

---

If I tick `Deny access` and send the invitation, the attachment link will be present in the ics file but I won't be able to open it in my browser (until I log in as the owner).

src/store/calendarObjectInstance.js
tcitworld: It's https://datatracker.ietf.org/doc/html/rfc8607#section-4.3

st3iny: Oh, right. I'm sorry for my stupidity.

But still, I don't like the custom serialization format. Ideally we should use base64 encoded JSON data and handle the encoding/serialization transparently through calendar-js.

src/components/Editor/Attachments/AttachmentsList.vue
```diff
@@ -0,0 +1,249 @@
...
+		},
+		getIcon(mime) {
+			return OC.MimeType.getIconUrl(mime)
+		},
+
+		getPreview(managedId, mime) {
+			if (managedId !== null) {
+				const _mId = managedId.split('||')
+				if ((_mId.length === 2) && (_mId[1] === 'true')) {
+					return generateUrl(`/core/preview?fileId=${_mId[0]}&x=100&y=100&a=0`)
```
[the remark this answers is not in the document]
st3iny: See comment from Thomas above.

src/models/attachment.js
```diff
@@ -0,0 +1,64 @@
...
+ * @return {object}
+ */
+const mapAttachmentPropertyToAttchmentObject = (attachmentProperty) => {
+	return getDefaultAttachmentObject({
+		attachmentProperty,
+		fileName: typeof attachmentProperty._parameters.get('FILENAME') !== 'undefined' ? attachmentProperty._parameters.get('FILENAME').value : null,
+		formatType: attachmentProperty.formatType,
+		uri: attachmentProperty.uri,
+		value: attachmentProperty.value,
+		managedId: typeof attachmentProperty._parameters.get('MANAGED-ID') !== 'undefined' ? attachmentProperty._parameters.get('MANAGED-ID').value : null,
```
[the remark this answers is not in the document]
st3iny: See the comment from Thomas above.

src/store/calendarObjectInstance.js
sazanof: @st3iny we can use parameters (X-NC-PREVIEW X-NC-FILE-ID), then this will simplify event processing and their correct selection in the future. now I have found a bug related to the slpit method. I'll redo it.

sazanof: https://github.com/nextcloud/calendar/pull/4251#issuecomment-1207938079

tcitworld (commented):
I did a small review of the code, still need to test myself.

tcitworld:
I'm wondering if https://github.com/nextcloud/server/pull/29886 should be used as an upload folder. :thinking:

szaimen:
> I'm wondering if [nextcloud/server#29886](https://github.com/nextcloud/server/pull/29886) could be used as an upload folder. 🤔

I am not sure as well. I mean it would be convenient to be able to work with these files like with any other file too. cc @jancborchardt @nimishavijay

jancborchardt:
@szaimen @sazanof @tcitworld yeah, in other apps like Talk, Notes and Collectives we put Files in the filesystem, so probably apt to do it here too.

@miaulalala @ChristophWurst "Calendar" folder, or do you prefer a hidden folder?

miaulalala:
> @szaimen @sazanof @tcitworld yeah, in other apps like Talk, Notes and Collectives we put Files in the filesystem, so probably apt to do it here too.
> 
> @miaulalala @ChristophWurst "Calendar" folder, or do you prefer a hidden folder?

I think we should do the same as other apps do.

src/components/Editor/Attachments/AttachmentsList.vue
```diff
@@ -0,0 +1,242 @@
...
+				</ActionButton>
+			</Actions>
+		</div>
+		<div v-if="attachments.length > 0">
+			<ul class="attachments-list-item">
+				<ListItem v-for="attachment in attachments"
+					:key="attachment.path"
+					:force-display-actions="true"
+					:title="getBaseName(attachment.fileName)"
+					@click="openFile(attachment.uri)">
```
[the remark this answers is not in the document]
sazanof: PR merged, but nc-vue version didnt' updated in package.json with this change

sazanof:
@tcitworld Tomas, I tried to fix everything.

> It's been published

don't understand what did you mean

sazanof:
**Oops!! 
I push it again!**

=)

st3iny:
@sazanof I rebased, fixed conflicts and am currently testing your PR. I had to force push in order to achieve that.

sazanof:
> I rebased, fixed conflicts and am currently testing your PR. I had to force push in order to achieve that. I hope that's not a problem.

Hi all!
@st3iny thanks! Let me know what do you think about implementation! 

I haven't tested it for a long time, and I hope that everything works as before. In any case, let me know how you test it.

st3iny:
> > @sazanof Tests have to be updated for the new `attachments` property.
> 
> Tell me, is it enough to add a parameter or do I need to write a full test? If it's a complete test, then I ask for help, since **I've never written tests and don't understand how to write them** (It's a matter of time and effort, I don't argue). =( I tried to write my own block for checking (by analogy with other tests) `it()` This error occurs if I add my parameters to
> 
> ```
> expect(getDefaultEvent Object()).to Equal({
> ....
> attachments: [
> 	'ATTACHMENT1',
> 	'ATTACHMENT2',
> 	'ATTACHMENT3',
> 	'ATTACHMENT4',
> 	]
> })
> ```
> 
> https://gist.github.com/sazanof/ecb2aaa8c760f93e74d2952294ba09ba
> 
> When checking, I get an error. I can't understand why it occurs...
> 
> ```
>  ● Test suite failed to run
> 
>     TypeError: Converting circular structure to JSON
>         --> --> --> --> --> --> --> --> starting at object with constructor 'CalendarComponent'
>         --- property '_root' closes the circle
>         at stringify (<anonymous>)
> ```

Maybe a bit late: Anything you put into `expect(...)` has to a plain object that can be converted to JSON. Most likely, you are trying to pass an object that has some form of reference to itself (e. g. via child -> parent -> child -> parent). It probably is caused by `expect(mapEventComponentToEventObject(eventComponent))`.

st3iny (commented):
Tested and works so far.

### I noticed multiple small issues:
1. ~Share links are not deleted once an attachment is removed. The file should be unshared in this case.~
2. ~Share links are not deleted once an event has no attendees anymore.~
3. Attachments are not saved into a sub folder (e. g. `/Calendar`). Instead, they are saved to the root directory which can get polluted quite quickly.
4. Uploaded attachments are silently overwritten if they have the same name. Example: Upload an attachment, edit another event and upload a similar file with the same name again -> The first attachment file will be overwritten without any warning.

I'm not sure what to do about 4. Maybe just append a random hash of 

**EDIT:** Regarding 1 and 2: This makes sense for recurring events because there might be recurrences that still have attendees/attachments. However, for regular events this might be worth to consider. I'm fine with ignoring this for now.

**EDIT 2:** Regarding 4: I just tested this in Talk (as they are essentially doing the same thing) and they just append a suffix (e. g. `some file (2).png`) to the file name.

sazanof:
@st3iny, thnaks! I have a question:

> Attachments are not saved into a sub folder (e. g. /Calendar). Instead, they are saved to the root directory which can get polluted quite quickly.

Did we decide that it would be a subfolder `/Calendar`?
And if so, how to create it? Probably this should happen at the time of activation or installation of the application itself?

And a completely noob question... how do I pull up the changes that you have made?
`git pull origin enh/attachements-to-events` -> Already updated =(

st3iny:
> @st3iny, thnaks! I have a question:
> 
> > Attachments are not saved into a sub folder (e. g. /Calendar). Instead, they are saved to the root directory which can get polluted quite quickly.
> 
> Did we decide that it would be a subfolder `/Calendar`? And if so, how to create it? Probably this should happen at the time of activation or installation of the application itself?

You can create folders using axios and the DAV API. You could create the folder just before uploading the attachment. If it already exist, the API will return status code 405.

```javascript
axios({
	method: 'MKCOL',
	url: `/remote.php/dav/files/${dav.userId}/Calendar`,
}).
```

> And a completely noob question... how do I pull up the changes that you have made? `git pull origin enh/attachements-to-events` -> Already updated =(

Hmmm ... git pull should just work. You could try `git pull --rebase origin enh/attachements-to-events`. If this doesn't work either and you have no uncommitted and/or unpushed local changes you could try `git fetch && git reset --hard origin/enh/attachements`. **Warning:** This will delete all local changes on this branch (if any).

sazanof:
> You can create folders using axios and the DAV API. You could create the folder just before uploading the attachment. If it already exist, the API will return status code 405.

OK, let`s check

BTW, There is an idea to ask the user whether he wants to create a duplicate file or overwrite it or not.

src/services/attachmentService.js
```diff
@@ -3,6 +3,8 @@ import { generateOcsUrl } from '@nextcloud/router'
 import { showError, showSuccess } from '@nextcloud/dialogs'
 import { translate as t } from '@nextcloud/l10n'
 
+const FOLDER = 'Calendar'
```
szaimen, line 6: can we somehow make the folder configurable? I know a few cases where this folder will not work in the user root. (e.g. when it is read only). 

E.g. this might be a good example: https://github.com/nextcloud/photos/blob/master/src/components/Settings/PhotosLocationSettings.vue

src/components/Editor/Attachments/AttachmentsList.vue
```diff
@@ -0,0 +1,241 @@
...
+						</ActionButton>
+					</template>
+				</ListItem>
+			</ul>
+		</div>
+	</div>
+</template>
+
+<script>
+import ListItem from '@nextcloud/vue/dist/Components/NcListItem'
```
szaimen, line 62: i think the imports are probably missing the file endings...

sazanof: Done via [e532920](https://github.com/nextcloud/calendar/pull/4251/commits/e532920b514c965edbd96e6138dafbaeba82eb52)

`150 problems (0 errors, 150 warnings)` =)

src/services/attachmentService.js
szaimen: Ping @sazanof ? :)

sazanof: > Ping @sazanof ? :)
```
PING @sazanof  (127.0.0.1) 56(84) bytes of data.
64 bytes from @sazanof  (127.0.0.1): icmp_seq=1 ttl=247 time=500000 ms
```
Yes, I looked at how it is implemented in the Photo application. In fact, we can add routes on the backend and expand the configuration on the frontend. I'll do it! :100: 

PS: Pings are unpredictable now.

sazanof: https://github.com/nextcloud/calendar/pull/4251/commits/ee408dc0a974625b8320203e1db4e16d21ccb1b8

Well, it was difficult, but that's in the past.

lib/Controller/SettingsController.php
```diff
@@ -172,6 +174,28 @@ private function setShowTasks(string $value):JSONResponse {
...
+	 *
+	 * @param string $value
+	 * @return JSONResponse
+	 */
+	private function setAttachmentsFolder(string $value):JSONResponse {
+
+		try {
+			$this->config->setUserValue(
+				$this->userId,
+				$this->appName,
```
tcitworld, line 188: Once the attachment part will be handled on the server by [Managed Attachments](https://datatracker.ietf.org/doc/html/rfc8607), it would be nice to keep this setting value, so we could set the preference in `dav` instead of `calendar` (`$this->appName`).

lib/Controller/ViewController.php
```diff
@@ -100,6 +100,7 @@ public function index():TemplateResponse {
 		$showWeekNumbers = $this->config->getUserValue($this->userId, $this->appName, 'showWeekNr', $defaultWeekNumbers) === 'yes';
 		$skipPopover = $this->config->getUserValue($this->userId, $this->appName, 'skipPopover', $defaultSkipPopover) === 'yes';
 		$timezone = $this->config->getUserValue($this->userId, $this->appName, 'timezone', $defaultTimezone);
+		$attachmentsFolder = $this->config->getUserValue($this->userId, $this->appName, 'attachmentsFolder', '/');
```
tcitworld, line 103: Save as above, use `dav` instead of `calendar`

tcitworld, line 103: Default to `/calendar`, not the root folder?

tcitworld:
Just noticed in https://datatracker.ietf.org/doc/html/rfc8607#section-4.3
> This property (`MANAGED-ID`) parameter MUST NOT be present in the case of unmanaged attachments.

So I'm no longer sure we should use it before implementing Managed Attachments in server. :thinking:

tcitworld:
Another few notes from reading [RFC 8607](https://datatracker.ietf.org/doc/html/rfc8607):
* Changing attachments should trigger scheduling messages to attendees https://datatracker.ietf.org/doc/html/rfc8607#section-3.12.6
* `MANAGED-ID` occurrences must be removed when exporting and importing data https://datatracker.ietf.org/doc/html/rfc8607#section-3.12.7

lib/Controller/ViewController.php
sazanof: I did so because the /Calendar folder is missing by default. And deleted the code that creates the /calendar folder before uploading local attachments. any ideas how to create folder once?

lib/Controller/SettingsController.php
sazanof: Simply change to 

```
$this->config->setUserValue(
				$this->userId,
				'dav',
				'attachmentsFolder',
				$value
			);
			```

?

File: lib/Controller/ViewController.php (hunk 11 above)
Comment (sazanof) [reply lib/Controller/ViewController.php]: https://github.com/nextcloud/calendar/pull/4251/commits/5a5638c6ff7a1b7d336b1fc8a90b913a8d32bb8b

File: lib/Controller/SettingsController.php (hunk 10 above)
Comment (sazanof) [reply lib/Controller/SettingsController.php]: https://github.com/nextcloud/calendar/pull/4251/commits/5a5638c6ff7a1b7d336b1fc8a90b913a8d32bb8b

**Comment by sazanof** [comment]
`Ping @all` =)

I'm thinking is there an option to create (check) a user folder without adding an additional queries of the form? (I hope I wrote it right. Translation difficulties, you know=))

```
axios({
	method: 'MKCOL',
	url: `/remote.php/dav/files/${dav.userId}/Calendar`,
}).

```
Who thinks what? 

see https://github.com/nextcloud/calendar/pull/4251#discussion_r1011385909

_There is very little left =)_

miaulalala:
> Another few notes from reading [RFC 8607](https://datatracker.ietf.org/doc/html/rfc8607):
> 
> * Changing attachments should trigger scheduling messages to attendees https://datatracker.ietf.org/doc/html/rfc8607#section-3.12.6

This should be doable by hooking into the sabre event lifecycle methinks.

> * `MANAGED-ID` occurrences must be removed when exporting and importing data https://datatracker.ietf.org/doc/html/rfc8607#section-3.12.7

I think the export should be ok, we sanitise the data a lot. Of course testing is needed. 
The import might work or it might not - `\Sabre\VObject\Component\VEvent::getValidationRules` is pretty strict, had this in another ticket where the (valid) ENCODING param threw.

lib/Controller/ViewController.php
st3iny: Checking for the folder each time involves a network request, too. I suggest checking the folder once on opening the calendar app and then store the boolean. The MKCOL operation can be done once an attachment file needs to be created if the folder does not exist yet.

sazanof: Hello! So, you can test this. I updated the code.

st3iny (commented):
I tested it again and it works really great now! I fixed the share creation as it wasn't using the attachment folder.

There is a bug with the attachment folder picker and I'm currently researching what is causing it.

sazanof:
Hello, @st3iny! Thank you for review =)
Hello everyone! So... drum roll...
> Tested and works!
> 
> Please squash **all** commits (including mine) into a single commit and then we are ready to merge.

szaimen:
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

sazanof:
*** Sounds of a champagne cork flying out ***

ChristophWurst:
@sazanof https://youtu.be/CeKNPSsK-kI?t=524 :pray:

PR #4268, opened 2022-06-10
Migrate timezone and calendar icon

GretaD:
ref https://github.com/nextcloud/groupware/issues/38

### DIFF
--- a/src/components/Shared/DatePicker.vue
+++ b/src/components/Shared/DatePicker.vue
@@ -36,16 +36,23 @@
 		:use12h="showAmPm"
 		:append-to-body="appendToBody"
 		v-bind="$attrs"
+		confirm
 		v-on="$listeners"
 		@close="close"
 		@change="change"
-		@pick="pickDate"
-		confirm>
+		@pick="pickDate">
 		<template #icon-calendar>
-			<button class="datetime-picker-inline-icon icon"
-				:class="{'icon-timezone': !isAllDay, 'icon-new-calendar': isAllDay, 'datetime-picker-inline-icon--highlighted': highlightTimezone}"
+			<Button type="tertiary-no-background"
 				@click.stop.prevent="toggleTimezonePopover"
-				@mousedown.stop.prevent="() => {}" />
+				@mousedown.stop.prevent="() => {}">
+				<template #icon>
+					<IconNewCalendar v-if="isAllDay"
+						:size="20" />
+					<IconTimezone v-else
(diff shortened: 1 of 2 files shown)


src/components/Shared/DatePicker.vue
```diff
@@ -46,10 +46,13 @@
 				@click.stop.prevent="toggleTimezonePopover"
 				@mousedown.stop.prevent="() => {}">
 				<template #icon>
-					<IconTimezone v-if="!isAllDay"
-						:size="20" />
 					<IconNewCalendar v-if="isAllDay"
 						:size="20" />
+					<IconTimezone v-if="!isAllDay && highlightTimezone"
```
tcitworld, line 51:
```suggestion
					<IconTimezone v-else-if="highlightTimezone"
```

src/components/Shared/DatePicker.vue
```diff
@@ -46,10 +46,13 @@
...
+						class="highlighted-timezone-icon"
+						:size="20" />
+					<IconTimezone v-if="!isAllDay && !highlightTimezone"
```
tcitworld, line 54:
```suggestion
					<IconTimezone v-else-if="!highlightTimezone"
```
Or just `v-else`

src/components/Shared/DatePicker.vue
```diff
@@ -36,16 +36,25 @@
...
-				@mousedown.stop.prevent="() => {}" />
+				@mousedown.stop.prevent="() => {}">
+				<template #icon>
+					<IconNewCalendar v-if="isAllDay"
+						:size="20" />
+					<IconTimezone v-else-if="highlightTimezone"
+						class="highlighted-timezone-icon"
+						:size="20" />
+					<IconTimezone v-else-if="!highlightTimezone"
+						:size="20" />
```
tcitworld, line 55:
```suggestion
					<IconTimezone v-else
						:class="{ 'highlighted-timezone-icon': highlightTimezone }"
						:size="20" />
```
Sorry, thought of it afterwards.

GretaD: ok, i will change it, but it does work like it is though

tcitworld: Yes, the class will be applied only if `highlightTimezone` is `true`.

GretaD: done

miaulalala (approved):

PR #4288, opened 2022-06-14
Add a config appsetting to allow to disable appointments

tcitworld:
Context: on a production instance I administer (~100k users, ~15M events), the search for available slots systematically timeouts. A smaller instance (~6k) works great.

As there's no loading or error feedback on the loading page, the experience isn't great. There's probably room for improvement in the booking filters implementation.

In any case, I'd like to be able to disable this feature for the meantime, and there's probably some admins that don't care about this feature anyway.

### DIFF
--- a/lib/Controller/ViewController.php
+++ b/lib/Controller/ViewController.php
@@ -95,6 +95,7 @@ class ViewController extends Controller {
 		$defaultReminder = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminder', $defaultDefaultReminder);
 		$showTasks = $this->config->getUserValue($this->userId, $this->appName, 'showTasks', $defaultShowTasks) === 'yes';
 		$hideEventExport = $this->config->getAppValue($this->appName, 'hideEventExport', 'no') === 'yes';
+		$disableAppointments = $this->config->getAppValue($this->appName, 'disableAppointments', 'no') === 'yes';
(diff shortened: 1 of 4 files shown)


tcitworld:
Note : this only hides the appointments panel in the sidebar, it doesn't prevent existing appointments to be filled by attendants.

miaulalala (approved):

pierreozoux:
@tcitworld if I may, I'm not a programmer, but as an admin, I'm always confused by negative variables.

I think it is better to avoid them:
https://davej.io/2018/05/negative-naming.html
(Not sure, and quiet interested by your programmer feedback :) )

Hope it helps! And thanks for this PR, it helps us a lot!

Mer0me:
Is this appsetting documented somewhere ? Is it available on V25.0.1 ?

tcitworld:
It's available in the calendar app >= 3.5.0.

See https://github.com/nextcloud/calendar/issues/4637#issuecomment-1282146040 for an example of usage.

Mer0me:
> It's available in the calendar app >= 3.5.0.
> 
> See [#4637 (comment)](https://github.com/nextcloud/calendar/issues/4637#issuecomment-1282146040) for an example of usage.

Hmm... doesn't work here...

`sudo -u www-data ./occ config:app:set calendar disableAppointments --value="yes"`
`Config value disableAppointments for app calendar set to yes`

But it's still here :

image

Am I missing something ?

PR #4329, opened 2022-06-29
l10n: Remove quotation marks

Valdnet:
_Reported in Transifex._

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListItem.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListItem.vue
@@ -327,11 +327,11 @@ export default {
 		 */
 		descriptionAppNavigationItem() {
 			if (this.calendar.enabled && this.calendar.displayName) {
-				return t('calendar', 'Disable calendar "{calendar}"', { calendar: this.calendar.displayName })
+				return t('calendar', 'Disable calendar {calendar}', { calendar: this.calendar.displayName })
 			} else if (this.calendar.enabled && !this.calendar.displayName) {
 				return t('calendar', 'Disable untitled calendar')
 			} else if (!this.calendar.enabled && this.calendar.displayName) {
-				return t('calendar', 'Enable calendar "{calendar}"', { calendar: this.calendar.displayName })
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Why? Appears like they were added intentionally.

raimund-schluessler:
> Why? Appears like they were added intentionally.

I would also keep them. They make it clear what the calendar name is. E.g. with a calendar named "today" it would be less obvious without the quotation marks: `Disable calendar today` vs `Disable calendar "today"`.

rakekniven:
I raised this issue by filing issue at Transifex.
I thought this lines are part of the activity list. There we do not use any quotation marks around variables. Mostly all variables are displayed bold in activity view.

Does this lines here appear at calender list inside calendar app?

rakekniven (commented):
Added question to conversation.

ChristophWurst:
It's the aria text on the color dot and the text that toggle a calendar visibility in the navigation

https://github.com/nextcloud/calendar/issues/3955#issuecomment-1064972946 for the design specs

Bildschirmfoto vom 2022-06-29 12-02-58

Valdnet:
Additionally, I would change all `» «` type quotes in text strings to be consistent and consistent with the spelling with other text strings.

https://github.com/nextcloud/calendar/blob/69c27f1c21032ae4c5ebbf3a280e99d7f6021d1e/lib/Controller/EmailController.php#L123
https://github.com/nextcloud/calendar/blob/69c27f1c21032ae4c5ebbf3a280e99d7f6021d1e/lib/Controller/EmailController.php#L194
https://github.com/nextcloud/calendar/blob/69c27f1c21032ae4c5ebbf3a280e99d7f6021d1e/lib/Controller/EmailController.php#L196
https://github.com/nextcloud/calendar/blob/69c27f1c21032ae4c5ebbf3a280e99d7f6021d1e/lib/Controller/EmailController.php#L197

rakekniven:
That means we keep the quotation marks here? (fine with me)

Valdnet:
So I'm closing the issue.

PR #4331, opened 2022-06-29
l10n: Change type of quotation marks

Valdnet:


Valdnet:
I am closing too many text strings to be changed in other applications.

PR #4385, opened 2022-07-19
[stable3.5] Fix contrast issue with share icon

jancborchardt:
Before | After
-|-
icon opacity before|icon opacity after

Please review @raimund-schluessler @JuliaKirschenheuter @nimishavijay :)

Resulting color & contrast:
image

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListItem.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListItem.vue
@@ -517,6 +517,6 @@ export default {
 
 <style lang="scss" scoped>
 	.app-navigation-entry__counter-wrapper .action-item.sharing .material-design-icon.share {
-		opacity: .3;
+		opacity: .7;
 	}
 </style>


nimishavijay (commented):
Looks good :) I'm also wondering why the colour of the link icon looks slightly darker 🤔

jancborchardt:
> Looks good :) I'm also wondering why the colour of the link icon looks slightly darker :thinking: 

Yeah it is darker – I didn’t want to mess too much with it cause there’s also a difference between having shared or not (maybe?) so I focused on just making it contrasting enough. Not an ideal fix – @JuliaKirschenheuter @miaulalala if you see no issues with it we could also remove the opacity value completely?

miaulalala:
> f you see no issues with it we could also remove the opacity value completely?

The icon is indeed different when a calendar has been shared:

image

"Work" was shared, "Personal" was not.

JuliaKirschenheuter:
@jancborchardt Thank you for taking over!
Suggestion: i would do such fixes in NC Vue lib to prevent possible conflicts. Share icon is everywhere (as all icon-buttons) and i think it would be better fix it in central place. What do you think about?

miaulalala:
Changed base to stable3.5 because the new design is already fixed.

PR #4391, opened 2022-07-21
Rename settings to Calendar Settings

PVince81:
For https://github.com/nextcloud/server/issues/32350

image

- [ ] Needs @jancborchardt's feedback, see https://github.com/nextcloud/server/issues/32350#issuecomment-1191434496

### DIFF
--- a/src/components/AppNavigation/Settings.vue
+++ b/src/components/AppNavigation/Settings.vue
@@ -218,7 +218,7 @@ export default {
 			return this.$store.state.importState.importState.stage === IMPORT_STAGE_IMPORTING
 		},
 		settingsTitle() {
-			return this.$t('calendar', 'Calendar Settings')
+			return this.$t('calendar', 'Calendar settings')
 		},
 		slotDurationOptions() {
 			return [{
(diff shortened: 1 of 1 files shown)


src/components/AppNavigation/Settings.vue
```diff
@@ -218,7 +218,7 @@ export default {
 			return this.$store.state.importState.importState.stage === IMPORT_STAGE_IMPORTING
 		},
 		settingsTitle() {
-			return this.$t('calendar', 'Settings & import').replace(/&amp;/g, '&')
+			return this.$t('calendar', 'Calendar Settings')
```
jancborchardt, line 221:
```suggestion
			return this.$t('calendar', 'Calendar settings')
```
We use "Sentence case" instead of "Capital Case" as per design guidelines. :)

jancborchardt (changes requested):
Probably ok to change the label (but note the capitalization issue).

To point to import we should probably also have something in the "New calendar" menu, but that's a separate enhancement.

src/components/AppNavigation/Settings.vue
PVince81: dammit, that reflex again :-(

miaulalala (approved):
:shipit:

PR #4450, opened 2022-08-31
Remove php test workaround for stable20

raimund-schluessler:
Nc server stable20 is not tested anymore, hence the workaround can go.


tcitworld (approved):
🧹

szaimen (approved):

PR #4452, opened 2022-08-31
Possibility to choose adressbook(s) for birthdays

omocinteractive:
It would be nice, to choose the adressbook(s) for birthdays or to disable adressbooks for the generated birthday calendar. Reason: I don't want to see every birthday from e.q. ex girlfriends and other contacts. But often, it's important, to have the adress.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 
 # Changelog
 
-## 3.5.0 - 2022-08-25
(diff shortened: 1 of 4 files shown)


tcitworld:
Did you mean to create an issue instead of a PR? Please use https://github.com/nextcloud/calendar/issues/new/choose instead.

omocinteractive:
Oh, Sorry!

PR #4459, opened 2022-09-05
Upgrade to Nextcloud 25 design

ChristophWurst:
* [x] Drop incompatible Nextcloud 22-24 support
* [x] Bump the version to reflect the breaking change(s)
* [ ] Update ´@nextcloud/vue` to v7
  * [x] Fix imports
  * [x] Fix vue/vue-template-compiler version mismatches
* [ ] Migrate `icons.scss` to a `icons.css` or process through webpack
* [x] Requires https://github.com/nextcloud/groupware/issues/38

PR #4489, opened 2022-09-08
Bump @nextcloud/vue to v6beta7

ChristophWurst:
To break https://github.com/nextcloud/calendar/pull/4459 down a bit.

Gave it a quick smoke tests. Some styling isn't ideal, but I'd suggest we do a good design review at the end of the migration and fix things in one go.

Bildschirmfoto vom 2022-09-08 19-15-45


szaimen (approved):
🐘

PR #4498, opened 2022-09-10
Move resources into Details tab and add new Media tab with related resources panel

Pytal:
### Requires

- [ ] https://github.com/nextcloud/calendar/pull/4491


tcitworld:
So the resources are related to the calendar, not the event?

Also, if the tab is named « Media » people might expect upcoming event attachments to be here too.

Pytal:
> So the resources are related to the calendar, not the event?

Yes that is how it is implemented, please confirm @ArtificialOwl

> Also, if the tab is named « Media » people might expect upcoming event attachments to be here too.

Will move the attachments as well

src/views/EditSidebar.vue
```diff
@@ -201,17 +206,24 @@
...
-				<MapMarker :size="20" decorative />
+				<FolderMultipleImage :size="20" decorative />
 			</template>
 			<div class="app-sidebar-tab__content">
-				<ResourceList v-if="!isLoading"
-					:calendar-object-instance="calendarObjectInstance"
-					:is-read-only="isReadOnly" />
+				<!-- FIXME provide integer calendar id -->
+				<RelatedResourcesPanel provider-id="calendar"
+					:item-id="calendarId"
```
Pytal, line 219: @st3iny @ChristophWurst related resources is implemented in a way that uses the integer calendar id https://github.com/nextcloud/related_resources/blob/31f3b2864d2a176686a25bda85d43e6c77b2c082/lib/RelatedResourceProviders/CalendarRelatedResourceProvider.php#L137, from @ArtificialOwl

The `calendarId` in this context is an alphanumeric string. How do we get the integer id here?

Pytal: Any insights @st3iny?

PVince81: @ChristophWurst can you help ?

tcitworld: (you need both the calendar URI and the principal URI to identify the calendar)

Pytal: In that case it sounds like the backend needs to be updated to use URI's instead of the integer id

Is this possible @ArtificialOwl?

ArtificialOwl: sure, send me an example of your data, i'll accommodate the backend

Pytal: Commits added to use the item id format: `{principalUri}:{calendarUri}`

Pytal:
- [ ] https://github.com/nextcloud/calendar/pull/4776

PVince81 (approved):
:+1:

PVince81:
@ChristophWurst can you help move this forward ?

PVince81:
@ChristophWurst @miaulalala @GretaD needs second review

miaulalala:
Sorry I really dislike the look of this feature.

Users are used to the separation of a tab for resources, and having resources in the Details view doesn't make sense.

We added the resources as a separate tab for a reason.

Why not add a 4th tab such as in the linked PR?

PVince81:
@jancborchardt @nimishavijay ^

tcitworld:
I also have some concerns with the whole thing:
* From what I can see in the code, related resources are opened in the same tab by default, so people will lose their event changes if they click on them before saving the event.
* There's already little share of users using the current resources tab, due to the lack of a GUI to manage resources and the fact only admins can manage them. Moving those into the Details first tab which is already really quite full since reminders don't have their tab anymore will only confuse users more.
* The fact that we show media in the event editor, whereas relations are made through calendars, makes little sense. Calendar shares don't change often, so my understanding is that users might always see pretty much the same items in this section
* From other apps, since there's no way to "view" a calendar in particular, clicking on a related resource with is a calendar will pretty much only load the calendar app, the benefit is not comparable to files, talk and deck where you access the related entity directly
* These two remarks make me think relations should really be made with calendar events (while considering things like calendar sharees, event attendees, title and date). Links from outside can then load the event directly.

Sorry for the late review, and please forgive me for being so negative on the topic, but I hope perhaps you'll find some constructive stuff in what I've said.

PVince81:
> Why not add a 4th tab such as in the linked PR?

from what I heard it was an explicit decision of @jancborchardt to put it there

@miaulalala @ChristophWurst if a different location needs to be used, please take over and make the change

an alternative would be to first merge this as is and move it in a separate iteration

miaulalala:
@nextcloud/designers pls add your input

SebastianKrupinski:
Hi,

This is now obsolete, the sidebar editor will be gone in the next release.

Issue #4399, opened 2022-07-27
Disabling sharing via link not honored


pboguslawski:
### Steps to reproduce

1. Disable "Settings > Sharing > Allow users to share via link and emails"
2. Calendars still allow to share it via links.

### Expected behavior

Sharing calendar via link should be disabled.

### Actual behaviour

Sharing calendar via link is still enabled.

### Calendar app version

3.4.2

pboguslawski:
Was fixed in https://github.com/nextcloud/calendar/pull/5679 but still does not work because [required fix was not applied to server](https://github.com/nextcloud/server/issues/50358).

tcitworld:
Fix at https://github.com/nextcloud/server/pull/34873

Issue #525, opened 2017-07-02
Option to disable public sharing


baltaner:
### Steps to reproduce
0. users are authenticated by ldap
1. Click on sharing-icon
2. Checkbox to public sharing appears
3.

### Expected behaviour
I disabled "Allow users to share by link" in nextcloud configuration
I enabled "Allow apps to use share-API"

I expect that there is no public sharing possible in calendar-app

### Actual behaviour
Public sharing is possible!

### Server configuration

**Operating system**:
ubuntu 16.04
**Web server:**
apache 2.4
**Database:**
mysql 5.7.18
**PHP version:**
7.0.18
**Server version:** (see your admin page)
12.0.0
**Calendar version:** (see the apps page)
1.5.3

georgehrke:
Please report in the server repo as well
@tcitworld

tcitworld:
Fix is coming, writing test...

PR #4572, opened 2022-10-01
Hide sharing calendar via link in UI when sharing via link is disabled

pboguslawski:
This mod hides sharing calendar via link in UI when `shareapi_allow_links` is disabled. Should be applied together with https://github.com/nextcloud/server/pull/34372

Related: https://github.com/nextcloud/server/pull/34372
Related: https://github.com/nextcloud/calendar/issues/525
Related: https://github.com/nextcloud/calendar/issues/4399
Author-Change-Id: IB#1126264
Signed-off-by: Pawel Boguslawski <pawel.boguslawski@ib.pl>

### DIFF
--- a/lib/Controller/PublicViewController.php
+++ b/lib/Controller/PublicViewController.php
@@ -6,6 +6,7 @@ declare(strict_types=1);
  *
  * @author Georg Ehrke
  * @copyright 2019 Georg Ehrke <oc.list@georgehrke.com>
+ * @copyright Copyright (c) 2022 Informatyka Boguslawski sp. z o.o. sp.k., http://www.ib.pl/
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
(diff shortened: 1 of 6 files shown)


tcitworld (changes requested):
You don't need this at all. :)

`calendar.canBePublished` will be `false` since it depends on the DAV property 
`cs:allowed-sharing-modes`, which is set by
https://github.com/nextcloud/server/blob/905753be42af7e8c5af304b0d3c93f5d9bcea789/apps/dav/lib/CalDAV/Publishing/PublishPlugin.php#L138

As long as the plugin is unregistered, the value won't be set anymore.
(ref https://github.com/nextcloud/server/pull/34372#discussion_r985205500)

pboguslawski:
This mod is not required when https://github.com/nextcloud/server/pull/34372 is merged. Thanks for pointing this optimization!

pboguslawski:
> As long as the plugin is unregistered, the value won't be set anymore.

Seems this solution is wrong because unregistering plugin like in https://github.com/nextcloud/server/pull/34372/commits/bfd8c622d99f24d7769db39b095ce00176e6da8a disables also possibility to share calendar with other nc users (and this PR is about disabling sharing via link only).

Let us know if you see better solution or we'll return to the original proposition.

PR #4600, opened 2022-10-12
Fix node ci

skjnldsv:
Signed-off-by: John Molakvoæ <skjnldsv@users.noreply.github.com>


skjnldsv:
Jest should also receive the same treatment imho

.github/workflows/node.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: npm i -g npm@"${{ steps.versions.outputs.npmVersion }}"
+
+      - name: Install dependencies & build
+        run: |
+          npm ci
+          npm run build --if-present
+
+      - name: Check webpack build changes
+        run: |
+          bash -c "[[ ! \"`git status --porcelain `\" ]] || (echo 'Please recompile and commit the assets, see the section \"Show changes on failure\" for details' && exit 1)"
```
tcitworld, line 50: There's no build files in this repo, this isn't needed

skjnldsv: This is a template and therefore this step can safely be ignored
No need to alter the template imho

ChristophWurst (approved):

PR #4601, opened 2022-10-12
Use npm 8, node 14, 16 and 18 on CI

miaulalala:


ChristophWurst (approved):

.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 15, 16]
+        node-version: [14, 16]
```
tcitworld, line 15: Title says 18?

ChristophWurst: My bad

PR #4644, opened 2022-10-18
Fix overlapint of timezones with time suggestions

### LINKED ISSUE #4615: Timezone element overlays time button on mid size screen resolutions on Appointments booking page
### Steps to reproduce

On 1920x1080
image

### Expected behavior

All buttons are visible

### Actual behaviour

One button is overlaid by the datetime div

### Calendar app version

v4.0.0-rc.1

### CalDAV-clients used

Web

### Browser

Chrome

### Client operating system

Ubuntu

### Server operating system

c.nc

### Web server

_No response_

### Database engine version

_No response_

### PHP engine version

_No response_

### Nextcloud version

_No response_

_No response_

### List of activated apps

_No response_

### Nextcloud configuration

_No response_

### Web server error log

_No response_

### Log file

_No response_

### Browser log

_No response_

### Additional info

_No response_

GretaD:
fixes #4615

after
Screenshot from 2022-10-18 15-13-55


GretaD:
> I'm sorry about being a show stopper today see_no_evil But there is a regression.
> 
> The time picker wraps onto the next line on normal resolutions. How about bumping `max-width` to a higher value, e. g. `250px`.
> 
hmmm, it all shows fine for me, it wraps only on small screen starting from 512px, but increasing the max-width gives me the same result as not having one. So lets have a bigger one.

miaulalala:
Maybe not for this PR but there is no indication that this is a datetime picker. Can we do something about that?

kesselb (approved):

PR #4689, opened 2022-10-27
Fix mobile widget showing past events and limiting results too much

miaulalala:
for calendar widget


ChristophWurst (approved):
Did the clients send an empty string?

miaulalala:
> Did the clients send an empty string?

looks like, at least iOS. I asked Marino to test it, so I've left it as a draft for now.

miaulalala:
Maybe this is a clue:
```php
	 * @param array $sinceIds Array indexed by widget Ids, contains date/id from which we want the new items
```

Could be that the value is just... wrong in general. I have disabled the option and will only show the next 14 days now.

ChristophWurst (approved):

Issue #4688, opened 2022-10-27
Repair step to update stored view for all users is extremely slow on large instances


akhil1508:
### Steps to reproduce

1. On a large nextcloud instance(10k+ users), run `occ app:disable calendar` 
2. Enable calendar again with `occ app:enable calendar`

### Expected behavior

Calendar app should be enabled instantly. 

### Actual behaviour

Calendar app takes a long time to be enabled. This is especially bad when updating a nextcloud instance as the amount of downtime is much higher. For an instance with ~10000 "seen" users, the update process is slowed down by around 5 minutes.

### Calendar app version

3.4.3

### Server operating system

Ubuntu

### Web server

Nginx

### Database engine version

MariaDB

### PHP engine version

PHP 8.0

### Nextcloud version

23.0.9

### Additional info

- [This repair step](https://github.com/nextcloud/calendar/blob/main/lib/RepairSteps/CurrentViewNameRepairStep.php#L69) goes through [all the seen users](https://github.com/nextcloud/calendar/blob/23a5826e122a6b7d9d54eb699f1f066ffd490985/lib/RepairSteps/CurrentViewNameRepairStep.php#L69) and attempts to update a config value if it is faulty.
- The faulty values "agendaDay", "agendaWeek" and "month" are already handled correctly as seen [here](https://github.com/nextcloud/calendar/blob/23a5826e122a6b7d9d54eb699f1f066ffd490985/lib/Controller/ViewController.php#L144) 
- It doesn't seem necessary to run this repair step every time the app is enabled(on updates or installs) as running it once should fix all values correctly
- Further, since the "faulty" values are indeed handled correctly in the controller code as seen above, it doesn't seem necessary to run this repair step at all
- I did run quick greps to find out if the old "faulty" values are actually set through some method in the code
  - `grep -Rnw 'agendaDay' ./apps/calendar`
  - `grep -Rnw 'agendaWeek' ./apps/calendar`
  - `grep -Rnw  'month' ./apps/calendar`
  - I couldn't find any place where they are set(so fixing them once should be enough, we don't need a repair step running each time imo)

ChristophWurst:
We could check if that *repair step* should rather be a migration. The difference is that repair steps run at every mgiration. A migration runs only once.

tcitworld:
I'm not even sure we need this anymore at all, this was for pre-2.0 stuff, long passed. We can keep the fallback in the controller for a while longer anyway.

miaulalala:
[...]

PR #4690, opened 2022-10-27
Remove current view repair step

avinashdchase:
This merge request fix the issue marked in https://github.com/nextcloud/calendar/issues/4688

Calendar app takes a long time to be enabled. This is especially bad when updating a nextcloud instance as the amount of downtime is much higher. For an instance with ~10000 "seen" users, the update process is slowed down by around 5 minutes.


miaulalala (approved):
Tested with:

```
occ app:disable calendar
occ app:enable calendar
```
Loaded calendar, works fine.

Thanks for the PR!

PR #4717, opened 2022-11-04
Change Appointment Config 'name' label

miaulalala:
to 'Appointment Name' to avoid translation errors to 'Surname' in UK English
| Before: | After |
| --- | --- |
| image | image|


ChristophWurst (approved):
Knickers in a twist

ChristophWurst:
That's one for the "why translations should't be identified by the source language".

PR #4718, opened 2022-11-04
REQ for Stable4.1: remove double user identification on Cals

punkyard:
hi
not long ago, the shared calendars have seen their display evolved.

Now, when a Calendar is shared, the creator is identified twice:
in between the brackets and in the user icon. why twice?
it's not elegant and hardens the reading - could you please remove the text in between the brackets please?

calendar user x2

plus, into brackets is given the user name, not full name, which is security issue as it can be use to try to login with it. When full user name can't permit to log in.

and please keep the Talk link through the user icon ;)
Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,15 @@
 
 # Changelog
 
-## 4.0.0 - 2022-10-11
+## 4.0.1 - 2022-10-18
+### Changed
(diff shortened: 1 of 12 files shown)


tcitworld:
Please create an issue instead of sending a Pull Request.

Issue #3853, opened 2021-12-28
Email address not shown for person who booked an appointment


Bubu:
### Steps to reproduce
1. Offer bookable appointments
2. As a non-logged in user book an appointment
3. Click on the email confirmation link
4. As the person offering the appointments, look at the created calendar entry and try and find a contact address for the person who booked the appointment

### Expected behaviour
Name and mail address are shown, perhaps with a mailto: link/button

### Actual behaviour
No address is shown. You can downlaod the event and try and import it into thunderbird to see the address. (So it's in the event, just not shown).

### Calendar app
**Calendar app version:** 3.0.3

**CalDAV-clients used:** Thunderbird... but this is solely done via the web-interface here

### Client configuration
**Browser:** FF 95.0.2

**Operating system:** Arch Linux

### Server configuration

**Operating system**: Arch Linux

**Web server:** Nginx

**Database:** PostgreSQL

**PHP version:** 8.0.14

**Nextcloud Version:** 23.0.0

**Updated from an older installed version or fresh install:** Upgraded from Owncloud 8 or 9 or so... :D

**List of activated apps:** 

```
  - accessibility: 1.9.0
  - activity: 2.15.0
  - audioplayer: 3.2.3
  - bookmarks: 10.0.3
  - calendar: 3.0.3
  - camerarawpreviews: 0.7.15
  - circles: 23.0.0
  - cloud_federation_api: 1.6.0
  - comments: 1.13.0
  - contacts: 4.0.7
  - contactsinteraction: 1.4.0
  - dashboard: 7.3.0
  - dav: 1.21.0
  - deck: 1.6.0
  - federatedfilesharing: 1.13.0
  - federation: 1.13.0
  - files: 1.18.0
  - files_external: 1.15.0
  - files_markdown: 2.3.5
  - files_pdfviewer: 2.4.0
  - files_rightclick: 1.2.0
  - files_sharing: 1.15.0
  - files_trashbin: 1.13.0
  - files_versions: 1.16.0
  - files_videoplayer: 1.12.0
  - firstrunwizard: 2.12.0
  - gpxpod: 4.3.0
  - logreader: 2.8.0
  - lookup_server_connector: 1.11.0
  - mail: 1.11.4
  - news: 17.0.1
  - nextcloud_announcements: 1.12.0
  - notes: 4.2.0
  - notifications: 2.11.1
  - oauth2: 1.11.0
  - ocsms: 2.2.0
  - passman: 2.4.0
[...]

PR #4742, opened 2022-11-17
Allow to select contact groups as attendees

### LINKED ISSUE #167: Add contact groups as attendees for an event [$50]
The participant event module should suggest contacts groups **from the contacts app**.

# Related issues
* To add Nextcloud groups (user management, including LDAP) as attendees for an event, see #2955
* To add a Nextcloud circle as attendees for an event, see #2954
* To add a Nextcloud talk group as attendee for an event, see #2956

tcitworld:
[demo.webm](https://user-images.githubusercontent.com/2197836/202509226-f6f5eaad-dfa9-4005-a0b8-15d7eae00f95.webm)

A **significant** limitation is that if a contact has multiple addresses, all of them will be added (as shown in the above video). Unfortunately once attendees are added, their email addresses are currently not shown, so you can't know which attendee item to remove.

I'm not sure what's the best way to proceed here and I'd like your opinion on that (as well as @nextcloud/designers ):
* only add the first email
* show the email in the attendee list (always or when a contact is added with several emails), so that the user can remove the unwanted emails
* add a special handling modal form to decide which email to pick (please don't say that :sweat_smile: ) ?
* change the way the multiselect works to select actual contacts, not contacts emails, and then pick the email used in the attendee list. This means we also need to allow contacts being added twice if the user wants to sent invites to both emails.

(Also the message « Group contains X contacts » kinda lies, it's about the attendees emails, not the attendees contacts.

Closes #167

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -39,8 +39,6 @@ return [
 		['name' => 'contact#searchPhoto', 'url' => '/v1/autocompletion/photo', 'verb' => 'POST'],
 		// Circles
 		['name' => 'contact#getCircleMembers', 'url' => '/v1/circles/getmembers', 'verb' => 'GET'],
-		// Contact Groups
-		['name' => 'contact#getContactGroupMembers', 'url' => '/v1/autocompletion/groupmembers', 'verb' => 'POST'],
 		// Settings
 		['name' => 'settings#setConfig', 'url' => '/v1/config/{key}', 'verb' => 'POST'],
 		// Tools
--- a/lib/Controller/ContactController.php
+++ b/lib/Controller/ContactController.php
@@ -103,10 +103,10 @@ class ContactController extends Controller {
 		}
 
 		$externalAttendeesDisabled = $this->appConfig->getValueBool('dav', 'caldav_external_attendees_disabled', false);
-		$result = $this->contactsManager->search($search, ['FN', 'EMAIL'], ['enumeration' => true]);
+		$contactsResult = $this->contactsManager->search($search, ['FN', 'EMAIL'], ['enumeration' => true]);
 
 		$contacts = [];
-		foreach ($result as $r) {
+		foreach ($contactsResult as $r) {
 			if (!$this->contactsService->hasEmail($r)) {
 				continue;
 			}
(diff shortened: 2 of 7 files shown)


ChristophWurst:
> * show the email in the attendee list (always or when a contact is added with several emails), so that the user can remove the unwanted emails

This was also discussed in https://github.com/nextcloud/calendar/issues/3853#issuecomment-1308523996. I think it would make sense and add value in general.

nimishavijay:
> show the email in the attendee list (always or when a contact is added with several emails), so that the user can remove the unwanted emails

This sounds good to me as well.

There should be something to indicate that there are multiple emails associated with one contact, especially when the group has many contacts. So we could possibly:
1. show the emails only for the contacts with multiple emails (and not for every contact)
2.  show all the emails and sort the contacts with multiple emails up top
3. show all emails and add some visual indication to show that the contact has multiple emails, like combining the 2 list items

I would go with 2) because it's simple and doesn't require any scrolling to see that there are multiple email contacts 
cc @jancborchardt

tcitworld:
For now I have added the suggestion from @ChristophWurst as it's the easy part. However I'm not sure it's needed if we show the emails in the attendee item subline as per 2. .

image

In any case we really need to remove the « invitation sent » label until the event is saved and the invitations are really sent, as adding the email next to it will even more increase confusion.

miaulalala:
I fundamentally disagree with hiding the email address in the three dot menu. I think it should be visible immediately without user interaction as it is relevant information and belongs to the same sort of information as a Display Name is.

ekrahmer:
from the user's perspective it is very simple: use the preferred email address
EMAIL;PREF=1;TYPE=OTHER:john@nowhere.com
EMAIL;TYPE=HOME:john@nowhere.net
This (RFC6350) PREF parameter is part of the NC contact database and can be handled by CardDAV apps like Cardbook. Ask the contact app developers to make it accessible in the NC web interface.

(I'm waiting since long to be able to invite the Tenor voice section of our choit to extra rehersals including theit accept/decline replies.....)

jancborchardt:
Talked a bit about this with @nimishavijay, we were going through the comments here and concluded:
- It would be annoying to receive mails to all your email addresses. Already now we have the situation that when receiving an invite you just accept it, and it would just be more annoying notifications.
- It would also be strange to clutter up the attendee list with all the mails of one person. You want to invite people, not emails. If you want to invite a person with a specific mail, you can still check through the list and invite the person via that mail


What we can do is:
- Using the "preferred" mail address as suggested by @ekrahmer is a good option
   - If that doesn’t exist, simply only pick the first one. In a server like e.g. a work Nextcloud, it is likely the first (or only) mail of a person which is their relevant one for a work event.
- Showing the emails directly in the list as mentioned by @miaulalala is also a good point. We can show them in the subline like so: "Invitation sent · mail@example.com"

miaulalala:
@tcitworld are you happy to implement the changes Jan suggested? I can take over otherwise, excellent work btw.

VLGM:
Hi,

I'm a newbie but when this will be released ?
How can I use this, because I absolutely need this function form teams planning.

Regards

Ekinoxies:
Hello all, great work ! Do you know when this feature will be implemented ? :D

JMoVS:
has there been any movement on this?

Durkar:
It would be great, if someone will contiue the work. :)

Upperholme:
The lack of any way to easily add multiple email addresses as invitees to an event is crippling. I've tried pasting in a comma separated list of addresses (doesn't work), pasting in from Excel or similar (doesn't work), adding contacts to an address book and/or group in the Contacts app and trying to name the book or group in the invitees field. Any or all of these approaches would allow me to use Nextcloud calendar. This particular thread is 18 months old. I've come across others that stretch back to 2015 or beyond.

mlennert:
Hoping that by posting here, this pull request will move higher up the chain of priorities of the reviewers that need to approve it. Really is a necessary feature. At this stage, not having it leaves to people creating circles that are redundant with groups as circles can be added, but groups not.

JustTigerDE:
Any progress on this since user groups are recently also shown in the contacts app as teams? Is there any chance this feature will be added?

jancborchardt (commented):
@tcitworld do you need any input from our side (@nimishavijay or design in general) to proceed? :)

tcitworld:
I don't have the time to look into this further unfortunately. Feel free to have someone take over.

ChristophWurst:
Superseded by https://github.com/nextcloud/calendar/pull/6285 and https://github.com/nextcloud/calendar/pull/8020

PR #4768, opened 2022-11-24
Fix dashboard calendar icon in dark mode

cappuMUC:
The fix enables dynamic switching to a white icon in dark mode. It is stuck to the black icon at the moment.

Signed-off-by: Bernhard Rohloff <rohloff.bernhard@gmail.com>

### DIFF
--- a/css/dashboard.css
+++ b/css/dashboard.css
@@ -1,3 +1,3 @@
 .app-icon-calendar {
-	background-image: url('../img/calendar-dark.svg');
+	background-image: var(--icon-calendar-dark);
 }


ChristophWurst:
Did it not work? :)

cappuMUC:
Well actually, yes and no! It fixes the issue in the dashboard widget but it also breaks the icon color in the widget select modal. To me it looks like there's no consistent way of handling icon colors in dark mode, is it?

This hole Github fork/pull request workflow was a bit confusing, too. It seems completely different since the last time I created a pull request. xD

I think I can come up with another fix, later today.

PR #4771, opened 2022-11-24
Fix black calendar icon on dashboard widget in dark mode

cappuMUC:
Second attempt. ;-)
This should fix the black calendar icon and leave the icon in the customization modal untouched.

### DIFF
--- a/css/dashboard.css
+++ b/css/dashboard.css
@@ -1,3 +1,4 @@
 .app-icon-calendar {
 	background-image: url('../img/calendar-dark.svg');
+  filter: var(--background-invert-if-dark);
 }


css/dashboard.css
```diff
@@ -1,3 +1,4 @@
 .app-icon-calendar {
-	background-image: url('../img/calendar-dark.svg');
+	background-image: var(--icon-calendar-white);
```
GretaD, line 2: this will not work because this variable doesnt exist anymore. All you need to do is to add the filter `filter: var(--background-invert-if-dark);` and leave the background image like it was :+1:

cappuMUC: Thank you for reviewing my patch. It looked like the right approach because of the other widgets on my dashboard had it all done with either inline SVG or CSS variables. However, I've made the changes according to your advice. :+1:

GretaD:
perfect
Screenshot from 2022-11-25 10-53-46

GretaD (approved):

PR #4803, opened 2022-12-13
Compare userId instead of email to exclude current user

artonge:
Right now, if Alice and Bob share the same address email, Alice won't be able to find Bob in the invitee list.

This PR improves the cleaning of the list by comparing userIds instead of email address.

### DIFF
--- a/src/components/Editor/Invitees/InviteesList.vue
+++ b/src/components/Editor/Invitees/InviteesList.vue
@@ -132,14 +132,7 @@ export default {
 			return !this.calendarObjectInstance.organizer && this.invitees.length === 0
 		},
 		alreadyInvitedEmails() {
-			const emails = this.invitees.map(attendee => removeMailtoPrefix(attendee.uri))
-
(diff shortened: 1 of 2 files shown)


ChristophWurst:
The fix probably makes sense. But I wonder if one doesn't encounter much bigger issues if two or more users share the email address. Doesn't sabre also look up principals via the email address? If you invite the email, will it invite both alice and bob? Or just whoever it finds first? Just thinking aloud …

miaulalala:
> The fix probably makes sense. But I wonder if one doesn't encounter much bigger issues if two or more users share the email address. Doesn't sabre also look up principals via the email address? If you invite the email, will it invite both alice and bob? Or just whoever it finds first? Just thinking aloud …

It will break. The invitation scheduling I think will work, but responding to an invitation is impossible as it can't be reliably determined who replied. Sabre Internal btw, so we have no option to modify this, nor should we as we would break compatibility with other clients.

PR #4851, opened 2022-12-30
chore!: drop support for PHP 7.4

kesselb:
Ref https://github.com/nextcloud/groupware/issues/41

Needs:
- [x] https://github.com/nextcloud/calendar/pull/4850

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up php$
         uses: shivammathur/setup-php@master
         with:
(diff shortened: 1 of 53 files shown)


miaulalala:
What's with all the newline removals? Are they needed for php8?

kesselb:
> What's with all the newline removals? Are they needed for php8?

It looks like a rule from PHP-CS that behaves different with PHP 7.4 (and do not complain about the newlines there).

st3iny (approved):
LGTM!

PR #4916, opened 2023-01-23
fix(appointments): Fix disabling appointments feature

tcitworld:
Was removed as part of a87dfb4bc0c00d0e68fee9d06c664f645faa46a3


ChristophWurst (approved):
Whoooooops

PR #4935, opened 2023-01-30
feat(caldav): Add calendar timezone picker

ChristophWurst:
https://www.rfc-editor.org/rfc/rfc4791#section-5.2.2

>    Description:  The CALDAV:calendar-timezone property is used to
>       specify the time zone the server should rely on to resolve "date"
>       values and "date with local time" values (i.e., floating time) to
>       "date with UTC time" values.  The server will require this
>       information to determine if a calendar component scheduled with
>       "date" values or "date with local time" values overlaps a CALDAV:
>       time-range specified in a CALDAV:calendar-query REPORT.  The
>       server will also require this information to compute the proper
>       FREEBUSY time period as "date with UTC time" in the VFREEBUSY
>       component returned in a response to a CALDAV:free-busy-query
>       REPORT request that takes into account calendar components
>       scheduled with "date" values or "date with local time" values.  In
>       the absence of this property, the server MAY rely on the time zone
>       of their choice.

Ref https://github.com/nextcloud/calendar/issues/4778#issuecomment-1385407475

## Todo

- [x] Make it work
- [ ] Make it pretty

Bildschirmfoto vom 2023-01-30 20-13-27

### DIFF
--- a/src/components/AppNavigation/EditCalendarModal.vue
+++ b/src/components/AppNavigation/EditCalendarModal.vue
@@ -43,6 +43,12 @@
 					:placeholder="$t('calendar', 'Calendar name …')">
 			</div>
 
+			<h2>{{ $t('calendar', 'Timezone') }}</h2>
+			<div class="edit-calendar-modal__timezone">
+				<NcTimezonePicker v-model="calendarTimezone"
+					@input="saveTimezone" />
+			</div>
+
 			<template v-if="canBeShared">
 				<h2 class="edit-calendar-modal__sharing-header">
 					{{ $t('calendar', 'Share calendar') }}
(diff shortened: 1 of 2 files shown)


src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -172,6 +184,23 @@ export default {
 
 			this.calendarName = this.calendar.displayName
 			this.calendarColor = this.calendar.color
+			if (this.calendar.timezone) {
+				try {
+					const jCal = ICAL.parse(this.calendar.timezone)
+					const iCalComp = new ICAL.Component(jCal)
+					const tzComp = TimezoneComponent.fromICALJs(iCalComp)
+					this.calendarTimezone = tzComp.getFirstComponent('VTIMEZONE').getFirstProperty('TZID').value
```
tcitworld, line 192: **question (non-blocking):** Do we need to validate the value? Maybe use calendar-js' `TimezoneManager.hasTimezoneForId()` before assigning it?

SebastianKrupinski:
Closing as this is no longer relavent

Issue #812, opened 2018-05-02
APIs for other apps


georgehrke:
The calendar app should provide APIs to allow other apps to extend the calendar's functionality.
Things that should be extendable:
- [ ] custom views
- [ ] advanced events editor

anything else?

Peterede:
We are looking at extending the calendar UI ourselves and are happy to help implement this.  Do you have a preferred hooking framework for other apps to modify the calendar?

georgehrke:
@Peterede One of the biggest issues that's keeping us from implementing this is the fact that we are still running on Angular 1, which is not being developed any further and is soon to be end of live.
We are looking forward to move to Vue: #540 

Once we moved to Vue, we will most likely provide a js API to hook in your own vue modules and a hock in our view controller to allow you to load your own javascript.

ImaCrea:
hello there, hope all is fine. I was wondering about the status of this issue please :)
we'd love to have this kind of features https://github.com/nextcloud/calendar/issues/2136 implemented after 🤗

all the good and happy new year

towo:
Hey @ChristophWurst, is this on the roadmap? Would incentives help? ;)

ChristophWurst:
https://docs.nextcloud.com/server/stable/developer_manual/digging_deeper/groupware/calendar.html

tcitworld:
The issue is more about extending the front-end views though. That link is not really relevant…yet.

towo:
Exactly, I was essentially riding on the back of #4948, as that's something we'd very much like to deploy.

luflow:
I am also currently looking for an easy way in the frontend to add a new entry to this context menu of an event to link to my own nextcloud app:

<img width="778" height="366" alt="Image" src="https://github.com/user-attachments/assets/a12d22ff-67ab-47ca-955c-c2b56549abcc" />

PR #4948, opened 2023-02-02
Calendar  BigBlueButton integration

SergeyMosin:
### Calendar Big Blue Button App Integration

If the BBB app ( https://apps.nextcloud.com/apps/bbb ) is installed and enabled users can replace the `Create Talk room for this event` button with an `Add video link` button which would add a link to a BBB room to the even description.

For now the following `occ` command enables/disables this functionality:
```
php occ config:app:set bbb calendar_integration --value="yes"
```

### DIFF
--- a/src/components/Editor/Invitees/BbbAddRoomActions.vue
+++ b/src/components/Editor/Invitees/BbbAddRoomActions.vue
@@ -1,0 +1,133 @@
+<template>
+	<NcActions v-if="isCreateBbbRoomButtonVisible"
+		:menu-title="$t('calendar', 'Add video link')"
+		force-title
+		force-menu
+		:disabled="isCreateBbbRoomButtonDisabled"
+		@open="getBbbRooms">
+		<template #icon>
+			<Video :size="20" />
+		</template>
+		<NcActionButton close-after-click @click="addBbbRoom(null)">
+			<template #icon>
+				<Plus :size="20" />
+			</template>
+			{{ $t('calendar', 'Create new room') }}
+		</NcActionButton>
+		<NcActionSeparator v-if="bbbRoomsList.length > 0" />
+		<NcActionButton v-for="room in bbbRoomsList"
+			:key="room.uid"
+			close-after-click
+			@click="addBbbRoom(room.uid)">
+			<template #icon>
+				<AccountBoxOutline :size="20" />
+			</template>
+			{{ room.name || $t('calendar', '(unnamed room)') }}
+		</NcActionButton>
+	</NcActions>
+</template>
+<script>
+
+import { mapState } from 'vuex'
+import NcActions from '@nextcloud/vue/dist/Components/NcActions.js'
+import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton.js'
+import NcActionSeparator from '@nextcloud/vue/dist/Components/NcActionSeparator.js'
+import Plus from 'vue-material-design-icons/Plus.vue'
+import Video from 'vue-material-design-icons/Video.vue'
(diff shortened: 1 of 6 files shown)


src/components/Editor/Invitees/BbbAddRoomActions.vue
```diff
@@ -0,0 +1,133 @@
...
+			<Video :size="20" />
+		</template>
+		<NcActionButton close-after-click @click="addBbbRoom(null)">
+			<template #icon>
+				<Plus :size="20" />
+			</template>
+			{{ $t('calendar', 'Create new room') }}
+		</NcActionButton>
+		<NcActionSeparator v-if="bbbRoomsList.length > 0" />
+		<NcActionButton v-for="room in bbbRoomsList"
```
tcitworld, line 18: **issue:** This won't be very nice when you have like 10+ rooms. There could be some kind of limitation of the list, then some kind of « List more rooms » option that opens maybe a modal to help filtering and finding the correct room.

Otherwise, just use a multi-select?

src/components/Editor/Invitees/BbbAddRoomActions.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+	<NcActions v-if="isCreateBbbRoomButtonVisible"
+		:menu-title="$t('calendar', 'Add video link')"
```
tcitworld, line 3: **suggestion:** I have the feeling it would be best to mention BBB in the label.

src/components/Editor/Invitees/BbbAddRoomActions.vue
```diff
@@ -0,0 +1,133 @@
...
+import { listBbbRooms, createBbbRoom, makeRoomUrlFromToken, doesDescriptionContainBbbLink } from '../../../services/bbbService.js'
+
+export default {
+	name: 'BbbAddRoomActions',
+	components: {
+		NcActions,
+		NcActionButton,
+		NcActionSeparator,
+		Plus,
+		// eslint-disable-next-line vue/no-reserved-component-names
```
tcitworld, line 48: **suggestion:** Just import the component as `VideoIcon` to avoid this.

src/services/bbbService.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * @param {string} [token=''] token
+ * @return {string}
+ */
+export function makeRoomUrlFromToken(token = '') {
+	// https://github.com/sualko/cloud_bbb/blob/master/ts/Common/Api.ts#L148
+	// TODO: url shortener
+	return window.location.protocol + '//' + window.location.host + getUrl('b/' + token)
```
tcitworld, line 86: **issue:** `getUrl` should already provide a full URL, not sure why you're concatenating `protocol` and `host` to it (also this breaks NC installed in subpaths).

tcitworld (commented):
This looks acceptable to me, but I'm guessing Nextcloud would prefer an integration through https://github.com/nextcloud/calendar/issues/812, and not have code for 3rd-party services integrated into this app.

ChristophWurst:
Thanks a lot for the integration. This is great!

> This looks acceptable to me, but I'm guessing Nextcloud would prefer an integration through https://github.com/nextcloud/calendar/issues/812, and not have code for 3rd-party services integrated into this app.

Right, I would say so indeed. The code for the integration should probably live in the BBB app, and the Calendar app gives BBB and any other app the possibility to hook into the UI.
This would lower the impact of this inter-app dependency and leave the code in the respective code owner's repo.

shurakenas:
What needs to be done to make it work? After entering the command, the button has not changed.

PR #4969, opened 2023-02-08
Better Calendar Widget Data

miaulalala:
Needs https://github.com/nextcloud/server/pull/36620

Resolved part of https://github.com/nextcloud/calendar/issues/4813

Signed-off-by: Anna Larch <anna@nextcloud.com>


lib/Dashboard/CalendarWidget.php
```diff
@@ -143,31 +154,65 @@ public function load(): void {
...
+				if(!empty($dtEnd) && $dtStart->diff($dtEnd)->days >= 1) {
+					// All day (and longer)
+					$timeString = $this->dateTimeFormatter->formatDateSpan($dtStart);
+				} elseif(!empty($dtEnd)) {
+					$timeString = $this->dateTimeFormatter->formatDateTime($dtStart, 'short', 'short') . ' - ' . $this->dateTimeFormatter->formatTime($dtEnd, 'short');
+				} else {
+					$timeString = $this->dateTimeFormatter->formatTimeSpan($dtStart, $startTimeWithTimezone);
+				}
 				$widget = new WidgetItem(
 					$calendarEvent['objects'][0]['SUMMARY'][0] ?? 'New Event',
```
tcitworld, line 211: **thought (non-blocking):** Unrelated, but the fallback should rather be named `Untitled event`.

lib/Dashboard/CalendarWidget.php
```diff
@@ -143,31 +154,65 @@ public function load(): void {
...
+		//          const start = dateFactory()
+		//			const end = dateFactory()
+		//			end.setDate(end.getDate() + 14)
+		//          const startOfToday = moment(start).startOf('day').toDate()
+		// get all vevents in this time range
+		// if "show tasks" is enabled, get all todos in the time range
+		// sort events by time
+		// filter events by:
+		// .filter(event => !event.classNames.includes('fc-event-nc-task-completed'))
+		// .filter(event => !event.classNames.includes('fc-event-nc-cancelled'))
```
tcitworld, line 167: **issue:** This part is no longer taken into account.

tcitworld (commented):
Since the time formatting is now done on the backend's side, we should format datetime using the user's own timezone setting, which might not be the same as the calendar's timezone or event the calendar object start or end time timezone.

lib/Dashboard/CalendarWidget.php
miaulalala: right, then it's in line with the frontend. nice catch

lib/Dashboard/CalendarWidget.php
miaulalala: yoda-patience

lib/Dashboard/CalendarWidget.php
```diff
@@ -30,6 +30,7 @@
 use DateTimeImmutable;
 use OCA\Calendar\AppInfo\Application;
 use OCA\Calendar\Service\JSDataService;
+use OCA\DAV\CalDAV\CalendarImpl;
```
ChristophWurst, line 33: this is a private API

ChristophWurst: yeah, still means this will introduce an inter-app dependency. we should avoid that at all cost.

tcitworld: `IGetTimezone` is just fine, right?

miaulalala:
Superceded by https://github.com/nextcloud/calendar/pull/5417

PR #4998, opened 2023-02-20
l10n: Fixed grammar

rakekniven:
Reported at Transifex.


ChristophWurst (approved):
Nit: if you want to use conventional commits as spec'ed at https://www.conventionalcommits.org/en/v1.0.0/#summary the commit message should be `fix(l10n): ......` or just `fix: ....`, **not** `10n: ...` because l10n is not a valid commit type.

PR #5022, opened 2023-02-25
Respect locale when displaying date and time in calendar

### LINKED ISSUE #2663: Start of the week day not correct, locale ignored
### Steps to reproduce
1. Open settings and change locale from (e.g.) en_US to:
Locale Settings
2. Open calendar.

### Expected behaviour
Week should start with monday as indicated in the settings. Name of days should be respecting the language (they are english instead of german in this case). Popup in "Create events" should also start on monday instead of sunday.

### Actual behaviour
Week starts on sunday. Weeknames are english.

### Calendar app
2.0.4

### Client configuration
Chrome Version 84.0.4147.105 (Official Build) (64-Bit)
(same for all browsers, browsers configured to german)

### Server configuration
Ubuntu Server 20.04 LTS

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.0.4

**Nextcloud Version:** 19.0.4

**Updated from an older installed version or fresh install:** Updated

**Nextcloud configuration:**

{
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
 "***REMOVED SENSITIVE VALUE***"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "dbtype": "mysql",
        "version": "19.0.4.2",
        "overwrite.cli.url":  "***REMOVED SENSITIVE VALUE***",
        "dbname": "***REMOVED SENSITIVE VALUE***",
        "dbhost": "***REMOVED SENSITIVE VALUE***",
        "dbport": "",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "installed": true,
        "htaccess.RewriteBase": "\/",
        "maintenance": false,
        "theme": "",
        "logfile": "***REMOVED SENSITIVE VALUE***",
        "loglevel": 2,
        "app_install_overwrite": [
            "calendar",
            "files_external_dropbox"
        ],
        "mail_smtpmode": "smtp",
        "mail_smtpsecure": "ssl",
        "mail_sendmailmode": "smtp",
        "mail_from_address": "***REMOVED SENSITIVE VALUE***",
        "mail_smtpauthtype": "LOGIN",
        "mail_smtpauth": 1,
        "mail_domain": "***REMOVED SENSITIVE VALUE***",
        "mail_smtphost": "***REMOVED SENSITIVE VALUE***",
        "mail_smtpport": "465",
        "mail_smtpname": "***REMOVED SENSITIVE VALUE***",
        "mail_smtppassword": "***REMOVED SENSITIVE VALUE***",
        "updater.release.channel": "stable",
        "mysql.utf8mb4": true,
        "memcache.local": "\\OC\\Memcache\\APCu",
        "ldapIgnoreNamingRules": false,
        "ldapProviderFactory": "OCA\\User_LDAP\\LDAPProviderFactory"
    }
}

#### Browser log
JS Log
JS Info Log
NC Settins

TeaDrinkingProgrammer:
Fixes #2663

This is my first commit, so please bear with me if I have made some mistakes 😄 .

Before: locale was not respected, fullCalendar displayed dates and time according to language
Now: Dates and times are now rendered by the two new functions in the options object.

# The new situation on the en_DE locale:
image
image

I might have missed some things and this pull request does not currently have tests, as I do not know what is normally tested and how it is tested.

### DIFF
--- a/src/components/CalendarGrid.vue
+++ b/src/components/CalendarGrid.vue
@@ -63,9 +63,7 @@ import VTimezoneNamedTimezone from '../fullcalendar/timezones/vtimezoneNamedTime
 // Import other dependencies
 import { mapGetters, mapState } from 'vuex'
 import debounce from 'debounce'
-import { getLocale } from '@nextcloud/l10n'
 import { getYYYYMMDDFromFirstdayParam } from '../utils/date.js'
-import { getFirstDayOfWeekFromMomentLocale } from '../utils/moment.js'
 
 export default {
 	name: 'CalendarGrid',
@@ -124,8 +122,7 @@ export default {
 				navLinks: true,
 				// Localization
 				...getDateFormattingConfig(),
-				locale: getFullCalendarLocale(getLocale(), this.locale),
-				firstDay: getFirstDayOfWeekFromMomentLocale(this.locale),
+				...getFullCalendarLocale(),
 				// Rendering
 				dayHeaderDidMount,
 				eventDidMount,
--- a/src/utils/moment.js
+++ b/src/utils/moment.js
@@ -53,6 +53,10 @@ export default async function loadMomentLocalization() {
 			LL: moment.localeData(realLocale).longDateFormat('LL'),
 			LLL: moment.localeData(realLocale).longDateFormat('LLL'),
 			LLLL: moment.localeData(realLocale).longDateFormat('LLLL'),
+			l: moment.localeData(realLocale).longDateFormat('l'),
+			ll: moment.localeData(realLocale).longDateFormat('ll'),
+			lll: moment.localeData(realLocale).longDateFormat('lll'),
+			llll: moment.localeData(realLocale).longDateFormat('llll'),
 		},
 		week: {
 			dow: moment.localeData(realLocale).firstDayOfWeek(),
@@ -70,19 +74,31 @@ export default async function loadMomentLocalization() {
  * @return {Promise<string>}
  */
 async function getLocaleFor(locale) {
+	// IMPORTANT: Keep each '/moment/local/...' string as is. Otherwise, webpack might not bundle
+	//            locale data because the contentRegExp fails to detect any files.
 	try {
-		// default load e.g. fr-fr
-		await import('moment/locale/' + locale)
+		// default load e.g. en-de
+		await import(`moment/locale/${locale}.js`)
 		return locale
 	} catch (error) {
+		const splitLocale = locale.split('-')
 		try {
-			// failure: fallback to fr
-			locale = locale.split('-')[0]
-			await import('moment/locale/' + locale)
+			// failure: fallback to de
+			locale = splitLocale[1]
+			await import(`moment/locale/${locale}.js`)
 			return locale
 		} catch (e) {
-			// failure, fallback to english
-			console.debug('Fallback to locale', 'en')
+			try {
+				// failure: fallback to en
+				locale = splitLocale[0]
+				await import(`moment/locale/${locale}.js`)
+				return locale
+			} catch (e) {
+				// failure, fallback to english
+				console.debug('Fallback to locale', 'en')
+				// English is the default locale and doesn't need to imported.
+				// It is already included in moment.js.
+			}
 		}
 	}
 
@@ -92,19 +108,8 @@ async function getLocaleFor(locale) {
 /**
  * Get's the first day of a week based on a moment locale
  *
- * @param {string} momentLocale Id of moment locale
  * @return {number}
  */
-export function getFirstDayOfWeekFromMomentLocale(momentLocale) {
-	return moment.localeData(momentLocale).firstDayOfWeek()
-}
-
-/**
- * Get's the first day of a year based on a moment locale
- *
- * @param {string} momentLocale Id of moment locale
- * @return {number}
- */
-export function getFirstDayOfYearFromMomentLocale(momentLocale) {
-	return moment.localeData(momentLocale).firstDayOfYear()
+export function getFirstDayOfWeekFromMomentLocale() {
+	return moment.localeData().firstDayOfWeek()
 }
--- a/src/components/Editor/FreeBusy/FreeBusy.vue
+++ b/src/components/Editor/FreeBusy/FreeBusy.vue
@@ -74,8 +74,6 @@ import {
 import Modal from '@nextcloud/vue/dist/Components/NcModal.js'
 import DatePicker from '../../Shared/DatePicker.vue'
 import { getColorForFBType } from '../../../utils/freebusy.js'
-import { getLocale } from '@nextcloud/l10n'
-import { getFirstDayOfWeekFromMomentLocale } from '../../../utils/moment.js'
 
 export default {
 	name: 'FreeBusy',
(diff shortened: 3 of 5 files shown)


TeaDrinkingProgrammer:
Commits should now be signed off

TeaDrinkingProgrammer:
And now with my actual email address 😅

src/components/CalendarGrid.vue
```diff
@@ -124,8 +123,7 @@ export default {
 				navLinks: true,
 				// Localization
 				...getDateFormattingConfig(),
-				locale: getFullCalendarLocale(getLocale(), this.locale),
-				firstDay: getFirstDayOfWeekFromMomentLocale(this.locale),
+				locale: getFullCalendarLocale(getLanguage(), this.locale),
```
ChristophWurst, line 126: We need some clarity on locale vs language for both FullCalendar and Nextcloud and if the two definitions match. With this change the FC locale is set to the NC language. This feels wrong, but maybe it is correct.

src/components/CalendarGrid.vue
```diff
@@ -124,8 +123,7 @@ export default {
 				navLinks: true,
 				// Localization
 				...getDateFormattingConfig(),
-				locale: getFullCalendarLocale(getLocale(), this.locale),
-				firstDay: getFirstDayOfWeekFromMomentLocale(this.locale),
```
ChristophWurst, line 128: is the first day ever set now?!

src/components/CalendarGrid.vue
TeaDrinkingProgrammer: Good point, it is the core of the current issue. NC seperates the two, but in FC locale = language. FC doesn't have a seperate way to input a locale. That is why, in the GetFullCalendarLocale, I have replaced the content of the date and time fields with NC moment calls. Should I rename locale to localeAndLang maybe?

src/components/CalendarGrid.vue
TeaDrinkingProgrammer: The first day was set in both the getFullCalendarLocale function (dow field) and firstDay before. Just the dow field works, but is not documented. I have now looked more into it, and I think that the locale should actually be set correctly from the moment factory plugin. I will look more into this.

TeaDrinkingProgrammer:
I had a look now, and changed the following:
The actual issue turns out to be that the custom Momentjs locale didn't override the formats for l, ll, lll and llll. Only L, LL, LLL and LLLL. When that was fixed, a lot of the custom code could be removed. The 'code' and 'week' options are now irrelevant because they are already taken care of by the Momentjs plugin. The buttonText options also are irrelevant now, as they are used [in a place](https://fullcalendar.io/docs/toolbar) which isn't used (anymore?) in Calendar. To make the sidebar time work, a new setting had to be added to the formatter which made the formatting of time explicit.

TeaDrinkingProgrammer:
No problem 😄. Do I need to add tests before it gets merged? I did look into it before, but as I di not know the codebase, it was a bit overwhelming making the tests

src/utils/moment.js
```diff
@@ -71,18 +75,25 @@ export default async function loadMomentLocalization() {
...
-			console.debug('Fallback to locale', 'en')
+			try {
+				// failure: fallback to en
+				locale = locale.split('-')[0]
+				await import('moment/locale/' + locale)
+				return locale
+			} catch (e) {
+				// failure, fallback to english
+				console.debug('Fallback to locale', 'en')
+			}
```
st3iny, line 96: I don't quite get this change. Could you please elaborate on it?

Why change the order of parsing locales (e.g. en-de: de -> en instead of en -> de)?

st3iny (commented):
Looks good so far. I left some questions regarding the code.

Testing this PR: I'm assuming that your changes should have no effect on the UI. It is just a code cleanup, right?

I did some testing using different locale and language settings and couldn't sport any changes compared to main.

src/utils/moment.js
TeaDrinkingProgrammer: This is the core of the problem. The locale here is loaded. If it doesn't exist, it will fallback. Right now, with en-de, English will be the locale and thus, the date formatting will be the moment 'en' locale. This is why right now, setting the locale to en-de, the starting day will be on Sunday. Most users see this as a bug.

This is why I think that the ultimate problem is the settings. I will illustrate. If you are German and want the language be English but date formatting in German, you can do all these combinations:

| Language       | Locale     | Expected | Current actual result |
|--------------|-----------|------------| ------ |
| English | German (de)    | German dates, English text  | German dates, English text
| English      | German English (en-de)  | German dates, English text       | English dates, English text

This system means you can also do weird combinations:

| Language       | Locale     | Expected | Current actual result |
|--------------|-----------|------------|------ |
| English | Monocan French  (fr-mc) | ??  | French dates, English text |

As you can see, it currently ignores the locale part of the language-locale string. This is why a temporary solution is to ignore the language part, as that is already encoded as the language setting, but ultimately I think that you should only be able to select countries in the locale menu, unless I am missing something.

st3iny: Ahh, now I understand the issue. Thanks for elaborating :)

I found a tiny bug in your code and pushed a fixup. Please take a look. The already split locale was split another time in the fallback logic and thus the second fallback was never working as intended.

TeaDrinkingProgrammer: I tested the original implementation in JSfiddle and it should also work: this is because it splits the locale parameter multiple times instead of the splitLocale. It is confusing though, so thank you for the refactor

src/utils/moment.js
```diff
@@ -71,18 +75,26 @@ export default async function loadMomentLocalization() {
...
+		// default load e.g.en-de
 		await import('moment/locale/' + locale)
 		return locale
 	} catch (error) {
+		const splitLocale = locale.split('-')
 		try {
-			// failure: fallback to fr
-			locale = locale.split('-')[0]
+			// failure: fallback to de
+			locale = splitLocale[1]
```
tcitworld, line 85: This will probably not go well with special cases of locales such as `ja-Latn-hepburn`.

TeaDrinkingProgrammer: I don´t see how it would, let me illustrate:

First it tries to load 'ja-Latn-hepburn´ and fails.

Then it splits the string into ['ja', 'Latn', 'hepburn']

The second try, it tries to load 'Latn', and fails.

The third try, it tries to load 'ja' and succeeds.

This is why this code should always succeed, as all locales should start with a language tag according to the (the W3 spec)[https://www.w3.org/International/articles/language-tags/#rfc]. The only theoretical situation where it fails I can think of, is a language which has not corresponding Momentjs locale, at which point it should fall back to english. I do now notice that the english fallback does not actually load a moment locale, I will fix that later today

st3iny (approved):
Tested and works. Thanks a lot!

I fixed a small issue with dynamically importing locales. The strings inside `import(...)` calls have to be literals. Otherwise, webpack won't be able to detect which modules to bundle.

tcitworld (approved):

foss-:
Should this change be in Calendar 4.3.2 update? This PR has no milestone so it is hard to tell. With 4.3.2 installed in 25.0.3:



Calendar shows as:

tcitworld:
It's not, this was not backported.

miaulalala:
> It's not, this was not backported.

we could?

PR #5024, opened 2023-02-27
ci(package): Use Composer v2

ChristophWurst:
:crossed_fingers: it helps with the CI failure in https://github.com/nextcloud/calendar/pull/4971


tcitworld (approved):
🧹

Issue #3820, opened 2021-12-16
Calendar creating unwanted empty events


Helium314:
### Steps to reproduce
1. click on empty time in day, week or month view
2. click on an existing event
3. see a new "untitled event" at the date / time you clicked before

### Expected behaviour
No event should be created. _Especially_ if there was no interaction with the event editor.

### Actual behaviour
An event is created

### Calendar app
**Calendar app version:** 3.0.2

**CalDAV-clients used:** none

### Client configuration
**Browser:** Vivaldi 5.0.2497.28, Firefox 95.0

**Operating system:** Kubuntu 20.04

### Server configuration

**Operating system**: Ubuntu 20.04

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.2.?

**Nextcloud Version:** 22.2.3

**Updated from an older installed version or fresh install:** updated

**List of activated apps:**

Accessibility	1.8.0	Featured	
Activity	2.15.0	Featured	
AppOrder	0.13.0		
Auditing / Logging	1.12.0	Featured	
Bookmarks	10.0.3		
Brute-force settings	2.3.0	Featured	
Calendar	3.0.2	Featured	
Circles	22.1.1	Featured	
Collaborative tags	1.12.0	Featured	
Comments	1.12.0	Featured	
Contacts	4.0.7	Featured	
Contacts Interaction	1.3.0	Featured	
Cookbook	0.9.8		
Dashboard	7.2.0	Featured	
Deleted files	1.12.0	Featured	
External storage support	1.13.0	Featured	
Federation	1.12.0	Featured	
File sharing	1.14.0	Featured	
Keeweb	0.6.8		
Log Reader	2.7.0	Featured	
Mail	1.11.4	Featured	
Maps	0.1.9	Featured	
Markdown Editor	2.3.5		
Monitoring	1.12.0	Featured	
Nextcloud announcements	1.11.0	Featured	
Notes	4.2.0	Featured	
Notifications	2.10.1	Featured	
Password policy	1.12.0	Featured	
PDF viewer	2.3.1	Featured	
PhoneTrack	0.6.9		
Photos	1.4.0	Featured	
Privacy	1.6.0	Featured	
Recommendations	1.1.0	Featured	
Right click	1.1.0	Featured	
Share by mail	1.12.0	Featured	
Support	1.5.0	Featured	
Talk	12.2.2	Featured	
Tasks	0.14.2	Featured	
Text	3.3.0	Featured	
[...]

PR #5117, opened 2023-03-30
Fix creating empty events

### LINKED ISSUE #4092: Unable to replace deleted appointment target calendar if there is only one
### Steps to reproduce

1. Make sure you have exactly one calendar
2. Create an appointment configuration
3. Delete the calendar
4. Create a new calendar
5. Open, edit and save the appointment configuration

### Expected behavior

As of https://github.com/nextcloud/calendar/pull/3967 we already select the new calendar. So after I click *Update*, I assume the new calendar to be saved for the appointment config.

### Actual behaviour

Visually it shows the new calendar in the multi select, but the old calendar URI is still saved. As there is only one option, the multi select is even disabled.

### Calendar app version

v3.2.2

### CalDAV-clients used

_No response_

### Browser

FF

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

_No response_

### Database engine version

_No response_

### PHP engine version

_No response_

### Nextcloud version

_No response_

_No response_

### List of activated apps

_No response_

### Nextcloud configuration

_No response_

### Web server error log

_No response_

### Log file

_No response_

### Browser log

_No response_

### Additional info

_No response_

rita-gama:
The problem was in the "beforeRouteLeave" because it didn't contemplated the case when the from route was the Popover to create the event and the to route was the popover to edit an existing event. 

Fixes https://github.com/nextcloud/calendar/issues/3820

### DIFF
--- a/src/mixins/EditorMixin.js
+++ b/src/mixins/EditorMixin.js
@@ -811,7 +811,10 @@ export default {
 		}
 
 		try {
-			await this.save()
+			if ((from.name !== 'NewPopoverView' || to.name !== 'EditPopoverView')
+			&& (from.name !== 'NewPopoverView' || to.name !== 'EditSideBarView')) {
+				await this.save()
+			}
 			next()
 		} catch (error) {
 			console.debug(error)
--- a/tests/javascript/unit/fullcalendar/interaction/eventClick.test.js
+++ b/tests/javascript/unit/fullcalendar/interaction/eventClick.test.js
@@ -20,6 +20,7 @@
  *
  */
 import eventClick from "../../../../../src/fullcalendar/interaction/eventClick.js";
+import EditorMixin from "../../../../../src/mixins/EditorMixin.js";
 import {
 	getPrefixedRoute,
 	isPublicOrEmbeddedRoute,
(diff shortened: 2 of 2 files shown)


ChristophWurst:
Did you copy the wrong ticket number? #4092 is about the appointments feature and selecting a conflict calendar. Your patch targets saving regular calendar events.

szaimen:
I think this probably fixes https://github.com/nextcloud/calendar/issues/3820?

rita-gama:
Sorry, I copied the worng ticket number. The ticket was indeed #3820.

st3iny (commented):
I noticed a regression while testing this: 

1. Edit an existing event.
2. Click on a new slot to create a new event.

Then: Edited event is saved.  
Now: Edited event is not saved.

Maybe both ifs could be replaced with: `if (from.name !== 'NewPopoverView' && from.name !== 'NewSideBarView')`. This way changes of existing events are always saved while new events are not automatically saved.

ChristophWurst:
Tests restarted because I don't think the failure is related to the changes

st3iny (commented):
I tested again and it does not solve the original issue.

https://github.com/nextcloud/calendar/assets/1479486/ad5983c5-8ce4-4005-9515-836f5040093a

rita-gama:
> I tested again and it does not solve the original issue.
> 2023-06-01.12-36-13.mp4

https://github.com/nextcloud/calendar/assets/101993105/6dea0dd4-2df7-4682-95cf-3071ff6131a3


I return to the initial conditional, which seems to resolve it.

rita-gama:
> I tested again and it does not solve the original issue.
> 2023-06-01.12-36-13.mp4

@st3iny can you try again? I made a small fix in the original condition that takes into account this https://github.com/nextcloud/calendar/pull/5117#pullrequestreview-1378641544

st3iny (approved):
Tested it again and now it works as expected. Thanks!

Please squash all your commits into a single one before the merge.

rita-gama:
> Tested it again and now it works as expected. Thanks!
> 
> Please squash all your commits into a single one before the merge.

Done! :+1:

Issue #4864, opened 2022-12-31
color picker not working correctly


JojoBr0:
### Steps to reproduce

1. open new settings modal
2. open extended color picker
3. change color

see video:
https://user-images.githubusercontent.com/92635650/210141307-57110511-6b8a-4475-9ba6-9e69ce8ec8a7.mp4

### Expected behavior

color menu should update color preview and wait until user presses "select" button

### Actual behaviour

color picker always closes immediately.

### Calendar app version

4.2.0

### Browser

Brave 1.46.144

### Client operating system

Windows 11

### Server operating system

Ubuntu 22.04

### Web server

Nginx

### Database engine version

PostgreSQL

### PHP engine version

PHP 8.1

### Nextcloud version

25.0.2

### Updated from an older installed version or fresh install

Updated from an older version

miaulalala:
Can reproduce, agreed that this is an issue. The close shouldn't be triggered when in the colour picker.

joaquimlbacao-ist:
Hi there! I'd like to know if I can start working on this bug, since it hasn't been fixed yet.

miaulalala:
> Hi there! I'd like to know if I can start working on this bug, since it hasn't been fixed yet.
of course! :pray:

PR #5137, opened 2023-04-08
fix: calendar color picker

joaquimlbacao-ist:
This pull request fixes issue #4864. By removing the use of debounce, the user can preview different colors for the calendar through the color picker, closing it when ready to save the edits.

Possible regression of https://github.com/nextcloud/calendar/pull/4515


st3iny (approved):
Tested and works! Thank you for the fix.

Issue #3735, opened 2021-11-28
Calendar Categories: Propose Categories already used


mklemme1:
**Is your feature request related to a problem? Please describe.**

The current calendar supports categories. 
The list of categories is hard-coded and cannot be modified (see #1644) . That's not helpful for many use cases.
We, for instance, have a calendar of events that many people maintain. 
We use categories to distinguish different types of events, like internal and external events. 

**Describe the solution you'd like**
In the categories selection list, list all the categories that were used in entries of the last two years and the future. 

**Describe alternatives you've considered**
Configuring a list (see #1644) would also work.
This solution is probably easier to implement (no extra GUI)

tcitworld:
Requires : https://github.com/nextcloud/server/issues/29950

Right now we can only suggest categories from the events currently loaded in the calendar, we don't have information for all categories.

rotdrop:
> Requires : [nextcloud/server#29950](https://github.com/nextcloud/server/issues/29950)
> 
> Right now we can only suggest categories from the events currently loaded in the calendar, we don't have information for all categories.

But the table `oc_calendarobjects`_props also already stores categories, so `vcategory_to_object` would be redundant ... and somehow appears to be completely unused.

SebastianKrupinski:
Closing in favour of https://github.com/nextcloud/calendar/issues/1644

Issue #1644, opened 2019-11-09
[Feature Request] Add own categories, delete default ones


Hundhausen:
**Describe the solution you'd like**
It would be nice if you could add own categories and delete default ones.

**Additional context**
With v2.0.0-beta2 and Issue #107 categories got added and so far it works. Categories from Thunderbird get shown in the web calendar and vice versa but it would be nice to delete unwanted default categories and add own categories for direct selection for the web calendar. I've looked into the files if there is an easy way to change this but as it looks, it isn't.

Malspherus:
You can add own categories in 2.0.1 (and probably also in some earlier versions). Deleting the defaults seems to not be possible at the moment, self-created categories don't show up in the list. Also see #71

Hundhausen:
You miss understood me or I wasn't specific enough. Adding a own Category to one event works but also assigning this category to an other event needs to be done manually (I talk about the Web Interface). When multiply users have access to this Calendar, this would be an issue. 

My idea is that an admin can add own categories that should be used as default and others not. When everyone needs to retype a category, small differences can occur and would mess things up. 
You can color code every category to show how important something is. For example blue for a in house event, purple for external events, green for holidays, red for very important event and so on. It would be a mess if someone would messes up one category by slipping on the keyboard and adding a character and so messing up the color and people not noticing it.

For a single user, this won't be an issue, because in Thunderbird I can add my own categories and just using them but this isn't possible for the web client.

Malspherus:
So in addition to what is available now:
- [ ] Shareable list of categories for every calendar, prefilled with current defaults
- [ ] Ability to add, remove and edit all categories in this list (which would probably mean updates to all events using them the way they are implemented at the moment)
- [ ] Some kind of permission system. In your case not everyone who can add events should be able to add new categories but they should still be able to use existing ones
- [ ] Sync of the whole list of categories to Thunderbird and other clients. I don't think that's possible right now, especially considering permissions

I could also imagine the Flow-App to integrate with this somehow, also kind of related to [#1540 (Comment)](https://github.com/nextcloud/calendar/pull/1540#issuecomment-544563314) for viewing categories easily I guess

Hundhausen:
Sync/Download with other clients would be a nice to have but not really mandatory. I don't know if the Clients or the Protocol would support this. Also it's not so hard to add the Categories once in the clients.

[...]

PR #5161, opened 2023-04-24
Augment the category menu by system tags and already used categories

rotdrop:
This commit add all available "collaborative tags" and all already used categories into option groups of the tags-menu of the side-bar editor.

Determining the set of already used categories is a little bit ugly: it used the oc_calendarobject_props table which might be considered "internal". However, this is the Nextcloud calendar app which only talks to the Nextcloud calendar server. So using this "internal ingredient" might be acceptable.

This commit addresses and is a related to a couple of open issues:

nextcloud/calendar#3735 Calendar Categories: Propose Categories already used

- this should be fixed by this commit

nextcloud/calendar#1644 Add own categories, delete default ones

- this is partly fixed in the sense that collaboritive tags are now also proposed as calendar categories.
- still default categories cannot be deleted
- however, using option groups one at least has some sort of overview about the origin of the proposed category

nextcloud/server#29950 Save VEVENT CATEGORIES as vcategory

- this issue is totally "ignored" by this commit as the proposed solution there is not needed (the categories are already there in the oc_calendarobject_props table)
- that would have to be discussed there: but my impression that the tables and classed mentioned there are obsolete and no longer used.

### DIFF
--- a/lib/Service/CategoriesService.php
+++ b/lib/Service/CategoriesService.php
@@ -1,0 +1,144 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Calendar App
+ *
+ * @copyright 2023 Claus-Justus Heine <himself@claus-justus-heine.de>
+ *
+ * @author Claus-Justus Heine <himself@claus-justus-heine.de>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Calendar\Service;
+
+use OCP\Calendar\ICalendarQuery;
+use OCP\Calendar\IManager as ICalendarManager;
+use OCP\IL10N;
+use OCP\SystemTag\ISystemTag;
+use OCP\SystemTag\ISystemTagManager;
+
+/**
+ * @psalm-type Category = array{label: string, value: string}
+ * @psalm-type CategoryGroup = array{group: string, options: array<int, Category>}
+ */
+class CategoriesService {
+	/** @var ICalendarManager */
+	private $calendarManager;
+
+	/** @var ISystemTagManager */
+	private $systemTagManager;
+
+	/** @var IL10N */
+	private $l;
+
+	public function __construct(ICalendarManager $calendarManager,
+		ISystemTagManager $systemTagManager,
+		IL10N $l10n) {
+		$this->calendarManager = $calendarManager;
+		$this->systemTagManager = $systemTagManager;
+		$this->l = $l10n;
+	}
+
+	/**
+	 * This is a simplistic brute-force extraction of all already used
+	 * categories from all events accessible to the given user.
+	 *
+	 * @return array
+	 */
+	private function getUsedCategories(string $userId): array {
+		$categories = [];
+		$principalUri = 'principals/users/' . $userId;
+		$query = $this->calendarManager->newQuery($principalUri);
+		$query->addSearchProperty(ICalendarQuery::SEARCH_PROPERTY_CATEGORIES);
+		$calendarObjects = $this->calendarManager->searchForPrincipal($query);
+		foreach ($calendarObjects as $objectInfo) {
+			foreach ($objectInfo['objects'] as $calendarObject) {
+				if (isset($calendarObject['CATEGORIES'])) {
+					$categories[] = explode(',', $calendarObject['CATEGORIES'][0][0]);
+				}
+			}
+		}
+
+		// Avoid injecting "broken" categories into the UI (avoid empty
+		// categories and categories surrounded by spaces)
+		$categories = array_filter(array_map(fn ($label) => trim($label), array_unique(array_merge(...$categories))));
+
+		return $categories;
+	}
+
+	/**
+	 * Return a grouped array with all previously used categories, all system
+	 * tags and all categories found in the iCalendar RFC.
+	 *
+	 * @return CategoryGroup[]
+	 */
+	public function getCategories(string $userId): array {
+		$systemTags = $this->systemTagManager->getAllTags(true);
+
+		$systemTagCategoryLabels = [];
+		/** @var ISystemTag $systemTag */
+		foreach ($systemTags as $systemTag) {
+			if (!$systemTag->isUserAssignable() || !$systemTag->isUserVisible()) {
+				continue;
+			}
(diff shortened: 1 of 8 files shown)


miaulalala:
Hi, thanks so much for your PR!

The system tags look great, that can stay as it is, but unfortunately, querying the oc_calendar_properties table from the calendar is a big no no. The table, despite it's misleading name, doesn't belong to the calendar app but the dav app in server. If the dav app changes anything about the table, it could break the calendar app. So the custom tags need to be exposed via an API from the dav app to get them in the calendar app.

If you need help with that, let me know! You can find any calendar related code in apps/dav/lib/CalDAV.

rotdrop:
> Hi, thanks so much for your PR!
> 
> The system tags look great, that can stay as it is, but unfortunately, querying the oc_calendar_properties table from the calendar app is a big no no. The table, despite it's misleading name, doesn't belong to the calendar app but the dav app in server. If the dav app changes anything about the table, it could break the calendar app

1. There is of course the possibility to just use the calendar PHP API from `lib/public/Calendar/`, fetch all user calendars, fetch all events, extract from each event the categories etc. This is not difficult but maybe somewhat inefficient.
2. And then there are those `vcategory` and `vcategory_to_object` tables mentioned in nextcloud/server#29950. These would be accessible through the `ITagManager` and `ITag` interfaces. However, those two tables appear to be unused (I think really just unused, not only unused by the dav app).

So a clean which lives without further modifications of the nextcloud/server and/or the dav-app would have to use possibility 1).

rotdrop:
>     1. There is of course the possibility to just use the calendar PHP API from `lib/public/Calendar/`, fetch all user calendars, fetch all events, extract from each event the categories etc. This is not difficult but maybe somewhat inefficient.
> 
>     2. And then there are those `vcategory` and `vcategory_to_object` tables mentioned in [Save VEVENT CATEGORIES as vcategory server#29950](https://github.com/nextcloud/server/issues/29950). These would be accessible through the `ITagManager` and `ITag` interfaces. However, those two tables appear to be unused (I think really just unused, not only unused by the dav app).
> 
> 
> So a clean implementation which lives without further modifications of the nextcloud/server and/or the dav-app would have to use possibility 1).

The current state of the code uses the public API from lib/public/Calendar/IManager. It simply slurps in all events and extract the categories. The result array returned by IManager->search('') is really quite complicated. I hope that the array layout is stable and will not change.

lib/Service/CategoriesService.php
```diff
@@ -0,0 +1,159 @@
...
+			return [];
+		}
+		$calendars = $this->calendarManager->getCalendarsForPrincipal('principals/users/' . $this->userId);
+		$count = count($calendars);
+		if ($count === 0) {
+			return [];
+		}
+
+		$categories = [];
+		$calendarObjects = $this->calendarManager->search('');
```
miaulalala, line 82: You should be able to limit the search to events that have set a category by using
```suggestion
		$calendarObjects = $this->calendarManager->search('', ['CATEGORIES']);
```

See https://github.com/nextcloud/server/blob/be59c4efd6caf4b1e80902441287a570718b5383/apps/dav/lib/CalDAV/CalendarImpl.php#L99

rotdrop: Your are right, that should work, just examined CalDavBackend.php. Late at night now, this is left for tomorrow.

Many thanks for the  hint!

lib/Service/CategoriesService.php
```diff
@@ -0,0 +1,159 @@
...
+			return [];
+		}
+		$calendars = $this->calendarManager->getCalendarsForPrincipal('principals/users/' . $this->userId);
+		$count = count($calendars);
+		if ($count === 0) {
+			return [];
+		}
+
+		$categories = [];
+		$calendarObjects = $this->calendarManager->search('', ['CATEGORIES']);
```
ChristophWurst, line 82: ```\OCP\Calendar\IManager::search``` is deprecated. Could you try to use ```\OCP\Calendar\IManager::searchForPrincipal``` instead? https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/groupware/calendar.html?highlight=searchforprincipal#calendar-objects

rotdrop: Overlooked that ... in progress ...

lib/Service/CategoriesService.php
```diff
@@ -0,0 +1,158 @@
...
+		$categories = [];
+		$principalUri = 'principals/users/' . $this->userId;
+		$query = $this->calendarManager->newQuery($principalUri);
+		$query->addSearchProperty(ICalendarQuery::SEARCH_PROPERTY_CATEGORIES);
+		$calendarObjects = $this->calendarManager->searchForPrincipal($query);
+		foreach ($calendarObjects as $objectInfo) {
+			foreach ($objectInfo['objects'] as $calendarObject) {
+				if (isset($calendarObject['CATEGORIES'])) {
+					$eventCategories = explode(',', $calendarObject['CATEGORIES'][0][0]);
+					$categories = array_merge(
```
ChristophWurst, line 86: tip: build a multi-dimensional categories array first and flatten that array after the loop

https://chemaclass.medium.com/never-use-array-merge-in-a-loop-74917b056ff9

lib/Service/CategoriesService.php
```diff
@@ -0,0 +1,158 @@
...
+						$categories,
+						array_fill_keys($eventCategories, true)
+					);
+				}
+			}
+		}
+		$categories = array_filter(array_map(fn ($label) => trim($label), array_keys($categories)));
```
ChristophWurst, line 93: could you add a comment explaining what the filter does here?

ChristophWurst (commented):
Code looks good otherwise

lib/Service/CategoriesService.php
rotdrop: I am not sure that this applies here. The point is also that we want to have a unique list of categories. Of course, if this was only about merging the individual category array then the repeated array_merge() inside the  loop would not be a good idea.

Of course, I can rephrase that piece, but I am not fully convinced that this will make things faster. A straight forward approach would be to generate a blown up array inside the loop, and then merge it and call array_unique.

Anyhow, see 055cbc7f

lib/Service/CategoriesService.php
rotdrop: I experience some "broken" categories in some instances which are surrounded by spaces or even completely "empty". I wanted to avoid to inject those into the category multi-select in the UI. I have added a comment.

ChristophWurst (approved):
Works

ChristophWurst:
If it's in your skill set I would appreciate unit tests for the newly added classes

JohannesGGE (approved):
Tested a bit, seems to work :+1:

lib/Service/CategoriesService.php
```diff
@@ -0,0 +1,144 @@
...
+		return $categories;
+	}
+
+	/**
+	 * Return a grouped array with all previously used categories, all system
+	 * tags and all categories found in the iCalendar RFC.
+	 *
+	 * @return CategoryGroup[]
+	 */
+	public function getCategories(string $userId): array {
```
ChristophWurst, line 90: FYI I have moved the userId from an injectable to a parameter because that makes the service usable without a user context, e.g. if we make use of it in a background job or CLI task one day

ChristophWurst:
@JohannesGGE @miaulalala please have a final test after my modifications

JohannesGGE (approved):
Tested and works.

---

The tag multiselect design issue is already known, right? (couldn't find a ticket):

Right after adding the tags:
image

After reopening the event sidebar:
Screenshot from 2023-06-28 18-25-46

PR #5170, opened 2023-04-28
Fix issue with Attendee not being string crashing tooltip

masonbrothers:
There is an issue when you upload a new ics file to nextcloud calendar that does not contain attendees.

Since there is an assumption that the attendee is going to be a string (and not undefined), the tooltip crashes when you try to edit an event.

```
TypeError: Cannot read properties of undefined (reading 'startsWith')
    at b9 (attendee.js:30:10)
    at o.userAsAttendee (EditorMixin.js:234:22)
    at e.get (vue.runtime.esm.js:3446:33)
    at e.evaluate (vue.runtime.esm.js:3547:27)
    at o.userAsAttendee (vue.runtime.esm.js:5537:25)
    at o.isViewedByAttendee (EditorMixin.js:222:16)
    at e.get (vue.runtime.esm.js:3446:33)
    at e.evaluate (vue.runtime.esm.js:3547:27)
    at o.isViewedByAttendee (vue.runtime.esm.js:5537:25)
    at o.<anonymous> (EditSimple.vue:1:4915)
```

This PR should fix this by coercing non strings to become strings.

### DIFF
--- a/src/utils/attendee.js
+++ b/src/utils/attendee.js
@@ -27,6 +27,8 @@
  * @return {string} URI without a mailto prefix
  */
 export function removeMailtoPrefix(uri) {
+	if (typeof uri !== 'string') return ''
+
 	if (uri.startsWith('mailto:')) {
 		return uri.slice(7)
 	}
@@ -41,6 +43,8 @@ export function removeMailtoPrefix(uri) {
  * @return {string} URI with a mailto prefix
  */
 export function addMailtoPrefix(uri) {
+	if (typeof uri !== 'string') return 'mailto:'
+
 	if (uri.startsWith('mailto:')) {
 		return uri
 	}
--- a/src/utils/attendee.js
+++ b/src/utils/attendee.js
@@ -45,7 +45,9 @@ export function removeMailtoPrefix(uri) {
  * @return {string} URI with a mailto prefix
  */
 export function addMailtoPrefix(uri) {
-	if (typeof uri !== 'string') return 'mailto:'
+	if (typeof uri !== 'string') {
+		return 'mailto:'
+	}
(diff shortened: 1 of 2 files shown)


src/utils/attendee.js
```diff
@@ -27,6 +27,8 @@
  * @return {string} URI without a mailto prefix
  */
 export function removeMailtoPrefix(uri) {
+	if (typeof uri !== 'string') return ''
```
ChristophWurst, line 30:
```suggestion
	if (typeof uri !== 'string') {
		return ''
	}
```

miaulalala, line 30: Please wrap your ifs
```suggestion
	if (typeof uri !== 'string')  {
	     return ''
	}
	     
```

miaulalala (changes requested):

src/utils/attendee.js
masonbrothers: Thanks! Will do next time I submit!

src/utils/attendee.js
```diff
@@ -41,6 +45,8 @@ export function removeMailtoPrefix(uri) {
  * @return {string} URI with a mailto prefix
  */
 export function addMailtoPrefix(uri) {
+	if (typeof uri !== 'string') return 'mailto:'
```
masonbrothers, line 48:
```suggestion
	if (typeof uri !== 'string') {
		return 'mailto:'
	}
```

miaulalala (approved):

miaulalala:
Thanks for the PR!

pierreozoux:
Thanks a lot \o/ Have a wonderful weekend!

PR #5184, opened 2023-05-02
fix(Public-Sharing): Show footer with server details and eventual ToS through PublicTemplateResponse

tcitworld:
The public calendar view used 'base' `Template` responses, preventing the footer that's being used on other public sharing pages from showing. This makes use of `PublicTemplateResponse` to show it automatically. It's however hidden (unchanged) on embed view.

Reference: https://help.nextcloud.com/t/feature-request-impressum-and-privacy-declaration-links/161162

localhost_8080_index php_apps_calendar_p_z9gd4FP8sgB8k8Cs_dayGridMonth_now

### DIFF
--- a/css/public.scss
+++ b/css/public.scss
@@ -60,12 +60,23 @@
 	}
 
 	.app-content {
-		margin-top: 44px
+		margin-top: 44px;
 		//position: absolute !important;
 		//top: 44px;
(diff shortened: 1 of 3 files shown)


st3iny (approved):
Tested and works.

Issue #4263, opened 2022-06-09
Many appointments are not distinguishable due to identical title


ChristophWurst:
### Steps to reproduce

1. Create an appointment "1:1 meeting"
2. Send the link to people and ask them to book a meeting freely

### Expected behavior

As the organizer of the meetings, I'm able to look at my daily/weekly schedule and see who I will have meeting with.

### Actual behaviour

It's all the same. There are many "1:1 meetings" in my calendar. I have to open the individually and check the attendee list to eventually find out which even belongs to which booking party.

### Calendar app version

3.x

### Additional info

@jancborchardt @nimishavijay any ideas? I thought about including the booking person name in the even title by default, but it's hard to place it correctly.

E.g. the format could be *{meeting title} with {name}*. In my example that would translate to *1:1 meeting with Jan*.

For the organizer that is a huge improvement, but then the booking person has a "meeting with themselves" in their calendar. Like for them it should probably say *1:1 meeting with Christoph*.

Getting the title right this way is not trivial.

A more technical solution would be to allow the user to place the name. Then I could say the appointment title is *1:1 meeting Christoph/{name}" and it turns into *1:1 meeting Christoph/Jan*. That title make sense for the organizer and the booking person.

The other idea would be to show attendee avatars on the events. We discussed that in https://github.com/nextcloud/calendar/issues/3162.

Do you have any other ideas?

JuliaKirschenheuter:
I agree with your suggestion about title like "1:1 meeting Christoph/Jan". It is a bit easier and also understandable for  both participants =)

leArthurDent:
Please consider that appointments are also used for external people booking slots. Focusing on usernames or avatars might not suffice to cover all applications.
Since you can provide a name when booking a slot, using that custom name might be more reasonable to use.

nimishavijay:
It's definitely a good idea to show the name of the booking person. 

I'm thinking about the cases when the appointment name itself is something like "Coffee with Nimisha" where it includes the organizer's name, then the calendar event would show "Coffee with Nimisha Christoph/Nimisha" which seems odd. Perhaps we could use a separator symbol, for eg. a middle dot (·)
Then the events would look like
 "1:1 meeting · Christoph and Jan" or "Coffee with Nimisha · Christoph and Nimisha"
The format would be `{title} · {booking person} and {organizer}`.

What do you think? cc @jancborchardt for any other suggestions

schiessle:
[...]

PR #5229, opened 2023-05-16
Include name of booking creator in event name in calendar

pjrobertson:
This is a really simple fix for #4263 -  I won't argue that this is the best or most 'complete' fix, but it's already a big improvement on having *all* appointments use the same name.

The format is now: `[DisplayName] - [AppointmentName]`

Throwing this out there in the hopes it can get merged quickly. If people still think having a field to enter 'topic' or 'meeting title' is worthwhile, then #4263 can be kept open for that. :)

### DIFF
--- a/lib/Service/Appointments/BookingCalendarWriter.php
+++ b/lib/Service/Appointments/BookingCalendarWriter.php
@@ -113,7 +113,8 @@ class BookingCalendarWriter {
 			'CALSCALE' => 'GREGORIAN',
 			'VERSION' => '2.0',
 			'VEVENT' => [
-				'SUMMARY' => $config->getName(),
+				// TRANSLATORS Title for event appoinment, first the attendee name, then the appointment name
+				'SUMMARY' => $this->l10n->t('%1$s - %2$s', [$displayName, $config->getName()]),
 				'STATUS' => 'CONFIRMED',
 				'DTSTART' => $start,
 				'DTEND' => $start->setTimestamp($start->getTimestamp() + ($config->getLength()))


lib/Service/Appointments/BookingCalendarWriter.php
```diff
@@ -113,7 +113,7 @@ public function write(AppointmentConfig $config,
 			'CALSCALE' => 'GREGORIAN',
 			'VERSION' => '2.0',
 			'VEVENT' => [
-				'SUMMARY' => $config->getName(),
+				'SUMMARY' => $displayName . ' - ' . $config->getName(),
```
tcitworld, line 116: **suggestion:** While the order is debatable, in any case this should be localisable (for RTL languages and such).
```suggestion
                                // TRANSLATORS Title for event appoinment, first the attendee name, then the appointment name
				'SUMMARY' => $this->l10n->t('%1$s - %2$s', [$displayName, $config->getName()]),
```

Otherwise it could be something like « $appointmentName with $attendeeName »

tcitworld:
cc @nextcloud/designers

lib/Service/Appointments/BookingCalendarWriter.php
pjrobertson: Much better @tcitworld  :) I went for name first because otherwise in my calendar I would still only see "Really long appointment name... " and wouldn't get to see the actual name, so they would all still appear the same.

Name first means I can at least get: "JOHN SMITH - Really long app...."

jancborchardt (approved):
Yep, nice and simple improvement! :)

lib/Service/Appointments/BookingCalendarWriter.php
```diff
@@ -113,7 +113,8 @@ public function write(AppointmentConfig $config,
 			'CALSCALE' => 'GREGORIAN',
 			'VERSION' => '2.0',
 			'VEVENT' => [
-				'SUMMARY' => $config->getName(),
+                                // TRANSLATORS Title for event appoinment, first the attendee name, then the appointment name
+				'SUMMARY' => $this->l10n->t('%1$s - %2$s', [$displayName, $config->getName()]),
```
szaimen, line 117: I doubt that this needs to be translated. cc @nickvergessen @Valdnet

szaimen: ah right, that would be a reason then

nimishavijay:
Nice! :) This is an improvement already! What do you think about the organizer's name also being included in order for it to make sense to the person to booked it as well? Something like `[Appointment name] · [name of person who booked] and [name of organizer]`? Also cc @jancborchardt and @szaimen

tcitworld:
Ideally we could have a different title for the event that's sent to the attendee and the one send and added to the organizer's calendar, but that would take special handing.

pjrobertson:
> Linter isn't happy (you can run composer run cs:fix for that), but otherwise LGTM, thanks!

Code updated to keep linter happy + squash commits.

> Yep, nice and simple improvement! :)

Yep, simple and does the trick :D

miaulalala (approved):
Works!

PR #5249, opened 2023-05-22
Fix 'after event' slot not being displayed on modal

### LINKED ISSUE #4743: Option adding time after appointment is saved but not displayed in th UI
### Steps to reproduce

1. Create an appointement schedule
2. Tick the box 'Time after an appointment'
3. Set a time
4. Click "Update"
5. Reopen the appointment schedule with the 3 dot menu

### Expected behavior

Time after an appointment should be set to the value defined previsoulsy.

### Actual behaviour

Time after an appointment is displayed as if nothing was set.
image

However, the feature is working correctly, meaning that if I've set once time after an appointement, that time is actually booked when somedy request a meeting with me.

NB: the behavior of time BEFORE the appointment is normal.

### Calendar app version

c.nc.c

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

_No response_

### Database engine version

_No response_

### PHP engine version

_No response_

### Nextcloud version

_No response_

_No response_

### List of activated apps

_No response_

### Nextcloud configuration

_No response_

### Web server error log

_No response_

### Log file

_No response_

### Browser log

_No response_

### Additional info

_No response_

GretaD:
If the code is identical to another method that works, then its a typo :) it took me 20 min.

fixes #4743

### DIFF
--- a/lib/Db/AppointmentConfig.php
+++ b/lib/Db/AppointmentConfig.php
@@ -207,7 +207,7 @@ class AppointmentConfig extends Entity implements JsonSerializable {
 			'length' => $this->getLength(),
 			'increment' => $this->getIncrement(),
 			'preparationDuration' => $this->getPreparationDuration(),
-			'followUpDuration' => $this->getFollowupDuration(),
+			'followupDuration' => $this->getFollowupDuration(),
 			'totalLength' => $this->getTotalLength(),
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
:see_no_evil: 

Backports?

PR #5251, opened 2023-05-22
Fix header showing on embed view

tcitworld:
Regression from https://github.com/nextcloud/calendar/pull/5184

| Before | After |
|--------|--------|
| Screenshot 2023-05-22 at 17-12-58 May 2023 - Nextcloud | Screenshot 2023-05-22 at 17-11-22 May 2023 - Nextcloud |

### DIFF
--- a/lib/Controller/PublicViewController.php
+++ b/lib/Controller/PublicViewController.php
@@ -104,6 +104,8 @@ class PublicViewController extends Controller {
 		$csp->addAllowedFrameAncestorDomain('*');
 		$response->setContentSecurityPolicy($csp);
 
+		$this->initialStateService->provideInitialState($this->appName, 'is_embed', true);
+
 		return $response;
 	}
 
--- a/css/public.scss
+++ b/css/public.scss
@@ -70,13 +70,18 @@
 	}
 }
 
-#body-public .app-calendar-public {
-	& + footer {
-		// Only show bottom rounded corners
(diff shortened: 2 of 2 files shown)


tcitworld:
Maybe `PublicTemplateResponse` could also have a `setHeaderVisible` method for this kind of case.

lib/Controller/PublicViewController.php
```diff
@@ -104,6 +104,8 @@ public function publicIndexForEmbedding(string $token):PublicTemplateResponse {
 		$csp->addAllowedFrameAncestorDomain('*');
 		$response->setContentSecurityPolicy($csp);
 
+		$this->initialStateService->provideInitialState($this->appName, 'is_embed', true);
```
ChristophWurst, line 107: is this used?

tcitworld: Yes, needed to target only the embed page so that we don't hide the header on the public page as well. Otherwise the only other solution I found was to use the `has()` CSS selector (to select based on child presence criteria), but it's [not fully available yet](https://caniuse.com/css-has).

ChristophWurst: Found it: ``input#initial-state-calendar-is_embed``. I looked for js code that fetches the state.

I guess that works until we change the internals of initial state

tcitworld: I'll propose https://github.com/nextcloud/calendar/pull/5251#issuecomment-1557406604 to avoid this

PR #5311, opened 2023-06-12
chore: add integration tests to CI and composer.json command

miaulalala:


.github/workflows/php-test.yml
```diff
@@ -50,10 +50,59 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          path: nextcloud/apps/calendar
+      - name: Install dependencies
+        working-directory: nextcloud/apps/calendar
+        run: composer install
+      - name: Run tests
+        working-directory: nextcloud/apps/calendar
+        run: composer run test:integration
+      - name: Upload coverage to Codecov
```
ChristophWurst, line 92: I think we can skip this. Coverage is better suited for unit tests.

.github/workflows/php-test.yml
```diff
@@ -50,10 +50,59 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: [ '8.0', '8.1', '8.2' ]
+        nextcloud-versions: [ 'master', 'stable27', 'stable26' ]
+        include:
+          - php-versions: '7.4'
+            nextcloud-versions: 'stable25'
+          - php-versions: '8.1'
+            nextcloud-versions: 'stable25'
```
ChristophWurst, line 63: simplify so it only tests each php version and nextcloud version at least once but not all possible combinations

this will reduce CI time

ChristophWurst (approved):
:rocket:

st3iny (approved):

.github/workflows/php-test.yml
```diff
@@ -50,10 +50,56 @@ jobs:
...
+          path: nextcloud/apps/calendar
+      - name: Install dependencies
+        working-directory: nextcloud/apps/calendar
+        run: composer install
+      - name: Run tests
+        working-directory: nextcloud/apps/calendar
+        run: composer run test:integration
+        with:
+          fail_ci_if_error: true
+          verbose: true
```
ChristophWurst, line 96:
```suggestion
```

ChristophWurst (dismissed):
clover options moved into previous step. drop them

ChristophWurst:
if's sufficient to re-request a review. you don't have to cancel :)

ChristophWurst (approved):

PR #5345, opened 2023-06-28
chore: add codecov token

miaulalala:
Add the codedov token so coverage reports are reliable again.


ChristophWurst (approved):

ChristophWurst:
well alright

Issue #4598, opened 2022-10-12
Sidebar editor updates original even instead of duplicate


ChristophWurst:
### Steps to reproduce

0. Enable the simple editor
1. Create an event
2. Click on the same event
3. Click on *More* to open the sidebar
4. Make modifications
5. Save

### Expected behavior

A modified duplicate of the original event is created.

### Actual behaviour

The original event is updated.

### Calendar app version

v4 alpha 2

PR #5348, opened 2023-06-30
fix(duplicateEvent): Open duplicate in sidebar not original

JohannesGGE:
Fix: #4598

#### How to test:
1. Create an event
2. Open the event again in simplified editor
3. Duplicate event
4. Click "More" to show sidebar editor
5. Change something
6. Save Event

On main this leads to updating the original event.
In this PR a new event is created (with the copied and edited data)


hamza221 (approved):
Tested

PR #5391, opened 2023-08-02
Fix navigation toggle overlap

jancborchardt:
This fixes the leftovers of https://github.com/nextcloud/groupware/issues/2
The only part where this is still a problem is in month view in mobile mode, but all the rest is significantly better so I would consider it done for now.

https://github.com/nextcloud/calendar/assets/925062/a04a018a-f41d-4b8e-9540-c1a3af597f43

https://github.com/nextcloud/calendar/assets/925062/1a7ceb80-8fd4-4c3b-a967-40561fc8a26b

### DIFF
--- a/css/app-navigation.scss
+++ b/css/app-navigation.scss
@@ -108,6 +108,11 @@
 	}
 }
 
+// Add background to navigation toggle to fix overlap with calendar elements
+.app-navigation-toggle {
+	background-color: var(--color-main-background) !important;
+}
+
 .app-navigation {
 	button.icon-share {
 		opacity: 0.3 !important;
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -116,6 +116,16 @@
 	word-break: break-word;
 }
 
+// Prevent events overlapping over day header
+.fc .fc-list-sticky .fc-list-day > * {
+	z-index: 1;
+}
+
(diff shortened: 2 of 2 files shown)


st3iny (approved):
Tested and works. Thank you very much for fixing this :heart:

raimund-schluessler:
Sorry to ruin the party here, but if one enables `Show week numbers`, the navigation toggle now overlaps the week number completely. This is an issue in the day, week and month view.

Day view:
Screenshot 2023-08-04 at 14-37-28 4  Aug  2023 - Kalender - Nextcloud


Week view:
Screenshot 2023-08-04 at 14-35-09 Woche 31 aus 2023 - Kalender - Nextcloud


Month view:
Screenshot 2023-08-04 at 14-35-30 August 2023 - Kalender - Nextcloud

jancborchardt:
@raimund-schluessler right, good point.
- in Day and Week view, we could widen the first column a little bit
- in Month view we could make the weekday row higher

Do you think this would be a proper fix or do you see any further issues with that? @ChristophWurst @st3iny @siccovansas

szaimen:
I found a working solution with 
```
.fc-scrollgrid-sync-inner {
margin-top:8px
}

.fc-timegrid-axis-cushion  {
margin-top:44px;
}
```

Issue #5389, opened 2023-08-02
Improve handling of links for online events


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a user of online meetings I want a smooth experience for links I have to open to join a call.

### Describe the solution you'd like

1. Write the conference URL into the location field when creating Talk rooms for an event. https://github.com/nextcloud/calendar/pull/5395
2. Implement a clickable link for the location field. https://github.com/nextcloud/calendar/pull/5355
3. Render the location field as a clickable link in the appointment notification mail (if it contains a link). https://github.com/nextcloud/server/pull/39753

### Additional context

This is an alternative approach to https://github.com/nextcloud/calendar/issues/1758 while it's poorly supported by other clients.

tcitworld:
I don't think 3. can be made with the current email template system. Maybe with the help of https://github.com/nextcloud/server/pull/38093 ? :smirk_cat:

ChristophWurst:
``\OCA\DAV\CalDAV\Schedule\IMipService::addMoreOptionsButton`` renders the "More options…" in emails. Can we use that?

PR #5395, opened 2023-08-04
feat: Write Talk URL into location, not description

ChristophWurst:
Fall back to description if location is filled, e.g. for a hybrid event.

This contributes part 2 of https://github.com/nextcloud/calendar/issues/5389.

## How to test

1. Set up Calendar and Talk
2. Create an event
3. Click the *Create Talk room* button

On `main` there will be an URL inserted into the description.
On this branch the URL will be written to the location if it's empty or not set. If it is, the URL is appended to the description.

### DIFF
--- a/src/components/Editor/Invitees/InviteesList.vue
+++ b/src/components/Editor/Invitees/InviteesList.vue
@@ -73,7 +73,7 @@ import InviteesListItem from './InviteesListItem.vue'
 import OrganizerListItem from './OrganizerListItem.vue'
 import NoAttendeesView from '../NoAttendeesView.vue'
 import OrganizerNoEmailError from '../OrganizerNoEmailError.vue'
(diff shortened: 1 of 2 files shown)


kesselb:
Very nice :+1: 

I would like to suggest adding a bit of explanation before the link in the description. 

Example Teams:

Screenshot from 2023-08-04 18-47-06

(Screenshot is from the imip email but the same text is used for the description ics field)

Example Google:

image

(Screenshot is also from the imip email but the same text is used for the description ics field)

As backup, the link should always be in the description. 
If the location field is empty, write the link inside. But always append to the description.

ChristophWurst:
Generating text is fun and games until you have more than one language involved 😃

kesselb:
Klicken Sie auf den nachfolgenden Link, um an der Online-Video-Besprechung mit nächster Wolkenverbreitung teilzunehmen ;)

Issue #2706, opened 2020-11-17
Secretary Permission for Calendar Sharing


ksteinb:
Hi,

it would be nice to have an additional permission for calendar sharing:  "Secretary mode" (or how you want to call it)

This permission should allow sharees in addition to create/edit/delete calendar entries also to send out invitations in behalf of the owner of the calender.

this would be a nice feature both for a secretary or also for some of our experimental groups to organize a common calender.

tcitworld:
The Nextcloud CalDAV backend [already has support for calendar delegation](https://github.com/nextcloud/server/pull/16666) and you can test it with some clients like Apple Calendar.

Renaming the issue title to match the feature request.

42ske:
any news on this?

ChristophWurst:
I've modified title and description again to have a user case and work packages.

kra-mo:
The main UI for adding a delegate could look like so:

<img width="500" height="296" alt="Image" src="https://github.com/user-attachments/assets/a2d8811f-f301-482c-866a-cf187b795c00" />

- [ ] There is no placeholder if no accounts are delegated. The context is enough.
- [ ] The tooltip for the `X` button is "Revoke access".
- [ ] When the `X` button is clicked, a confirmation dialog appears: "**Revoke access?** {email address} will no longer be able to act on your behalf" with the options "Cancel" and "Revoke".

When clicking "Add delegate", the following dialog appears:

<img width="449" height="202" alt="Image" src="https://github.com/user-attachments/assets/7fed368c-c933-4b37-8376-99e9a056c959" />

Then, after selecting someone, the user receives an email informing them about the delegation.

Delegated calendars would show up under their own section in the sidebar, "Delegated".

If there are more requirements, like custom permissions, etc., please tell me :)

ChristophWurst:
> The main UI for adding a delegate

Since we delegate specific calendars, should this be part of the existing calendar editing modal or a separate one?

kra-mo:
> > The main UI for adding a delegate
> 
> Since we delegate specific calendars, should this be part of the existing calendar editing modal or a separate one?

I guess you're right. It could be its own dialog to be more discoverable.

In that case, the main "Delegation" title could be dropped and it could be accessed from a "Delegation" menu item or similar from the … menu.

ChristophWurst:
> Since we delegate specific calendars, should this be part of the existing calendar editing modal or a separate one?

[...]

PR #5457, opened 2023-09-06
Shared calendars shouldn't allow adding attendees if the calendar is not yours

### LINKED ISSUE #5450: Shared calendars shouldn't allow adding attendees if the calendar is not yours
### Steps to reproduce

1. Open calendar app
2. Create and event and select a shared calendar
3. Add an attendee

### Expected behavior

I shouldnt be able to add an attendee in a calendar that is not mine because they will receive no invitation

### Actual behaviour

I can add an attendee, and thats confusing people the other person will not know that i sent them a request for a meeting

### Calendar app version

_No response_

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

_No response_

None

### List of activated apps

_No response_

### Nextcloud configuration

_No response_

### Web server error log

_No response_

### Log file

_No response_

### Browser log

_No response_

### Additional info

_No response_

GretaD:
fixes #5450

Screenshot from 2023-09-07 09-16-12

### DIFF
--- a/src/components/Editor/Invitees/InviteesList.vue
+++ b/src/components/Editor/Invitees/InviteesList.vue
@@ -23,22 +23,24 @@
 
 <template>
 	<div>
-		<InviteesListSearch v-if="!isReadOnly && hasUserEmailAddress"
+		<InviteesListSearch v-if="!isReadOnly && !isSharedWithMe && hasUserEmailAddress"
 			:already-invited-emails="alreadyInvitedEmails"
 			@add-attendee="addAttendee" />
 		<OrganizerListItem v-if="hasOrganizer"
-			:is-read-only="isReadOnly"
+			:is-read-only="isReadOnly || isSharedWithMe"
 			:organizer="calendarObjectInstance.organizer" />
 		<InviteesListItem v-for="invitee in inviteesWithoutOrganizer"
 			:key="invitee.email"
 			:attendee="invitee"
-			:is-read-only="isReadOnly"
+			:is-read-only="isReadOnly || isSharedWithMe"
 			:organizer-display-name="organizerDisplayName"
 			@remove-attendee="removeAttendee" />
 		<NoAttendeesView v-if="isReadOnly && isListEmpty"
 			:message="noInviteesMessage" />
 		<NoAttendeesView v-if="!isReadOnly && isListEmpty && hasUserEmailAddress"
 			:message="noInviteesMessage" />
+		<NoAttendeesView v-if="isSharedWithMe"
+			:message="noOwnerMessage" />
 		<OrganizerNoEmailError v-if="!isReadOnly && isListEmpty && !hasUserEmailAddress" />
 
 		<div class="invitees-list-button-group">
@@ -101,6 +103,10 @@ export default {
 			type: Object,
 			required: true,
 		},
+		isSharedWithMe: {
+			type: Boolean,
+			required: true,
+		},
 	},
 	data() {
 		return {
@@ -114,6 +120,9 @@ export default {
 		}),
 		noInviteesMessage() {
 			return this.$t('calendar', 'No attendees yet')
+		},
+		noOwnerMessage() {
+			return this.$t('calendar', 'You don\'t own this calendar, so you cannot add attendees to this event')
 		},
 		invitees() {
 			return this.calendarObjectInstance.attendees.filter(attendee => {
@@ -136,6 +145,7 @@ export default {
 		hasOrganizer() {
 			return this.calendarObjectInstance.organizer !== null
 		},
+
 		organizerDisplayName() {
 			return organizerDisplayName(this.calendarObjectInstance.organizer)
 		},
--- a/src/mixins/EditorMixin.js
+++ b/src/mixins/EditorMixin.js
@@ -214,6 +214,18 @@ export default {
 
 			return calendar.readOnly
 		},
+		isSharedWithMe() {
+			if (!this.calendarObject) {
+				return true
+			}
+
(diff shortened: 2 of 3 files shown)


ChristophWurst (approved):

src/components/Editor/Invitees/InviteesList.vue
```diff
@@ -115,6 +119,9 @@ export default {
 		noInviteesMessage() {
 			return this.$t('calendar', 'No attendees yet')
 		},
+		noOwnerMessage() {
+			return this.$t('calendar', 'You dont own this calendar, so you cannot invite other people')
```
tcitworld, line 123: **suggestion:** 
```suggestion
			return this.$t('calendar', "You don't own this calendar, so you cannot add attendees to this event")
```
As attendees may not just be people in the future.

ChristophWurst: We should probably also disable rooms and resources

nunoleite:
I'm sorry, but i disagree.

On our Nextcloud implementation we have a user that's the Admin, that manages some global things. Like folders and calendars. Because Nextcloud in centered on the User and not on Projects we have a user that created some resources to centralize Projects and that is shared to everyone.

So, what we have is a calendar that was created in the Admin user (no one uses this Admin).
We have a "Generic" calendar that is shared to everyone. Than we use this calendar to put events so everyone can know whats happening but these events are only for some people to attend.

We need to have a calendar to put events about a Project so everyone knows whats happening, but these events are related to only one project so only 2 or 3 persons will be attended.

And by the logic these 2 or 3 attendees should receive a notification by email, they only receive by Nextcloud App notification.
And if i have Edit permissions on this calendar why shouldn't i have permissions to edit the Attendees?
They should receive the email notification has it is on normal calendars.

Thanks

ChristophWurst:
@nunoleite the problem is that this setup is broken. Nextcloud doesn't send out invites if you aren't the owner of the calendar. See https://github.com/nextcloud/server/issues/26668.

outofcontrol:
@ChristophWurst  we have the patch in place for invites to go out which works very quite well, and our users have always managed our shared calendar. We just upgraded from v21 to v27 and find users can no longer add attenedees. I would ask that this change be reconsidered.

Ideally, and what IMO would make the calendar better, is to fix the invitation issue and not hobble the ability to maintain calendars. 

I believe this change to be taking NextCloud calendar in the wrong direction.

kesselb:
> we have the patch in place for invites to go out which works very quite well

I advise against using https://github.com/nextcloud/server/pull/36756 or https://github.com/nextcloud/server/pull/39702. While invitations work, it still leaves many rough edges. For example, events are shown twice sometime. The behavior for create and update events is not aligned, and weird things happen when deleting such events.

outofcontrol:
We've been using it for I think close to 2 years, seems fine. But that is neither here nor there in regards to this change, which is removing a feature that was relied on.

tahonermann:
> I advise against using https://github.com/nextcloud/server/pull/36756 or https://github.com/nextcloud/server/pull/39702.

@kesselb, unfortunately, I believe Nextcloud offers no alternative at present. Per #2706, there is no ability to delegate calendar management to another user or group. Per [#26668](https://github.com/nextcloud/server/issues/26668), invitations aren't sent if a non-owning user edits a shared calendar.

The removal of notifications for edits to shared calendars effectively removed the only working method for group management of a calendar. Or do you perhaps know of some other way to configure a calendar so that it can 1) be edited by multiple users, and 2) have such edits be reflected in notifications? I can't find any other documented way to accomplish this. Use of a shared account is not an attractive option for several reasons.

nunoleite:
The lack of functionalities centered on Projects and not only on Users is a problem for project managers. And we, as a user, have to invent some workarounds. If these workarounds are being limited then the all use of Nextcloud for these things is being limited.

Note to Nextcloud development... please start thinking in implementing something for Project Management... Centered in Projects... If we (Nextcloud community) want to be a ms teams and others alternative, we have to move that way. That's what i think. We have all the projects things spread across multiple users and multiple owners. This is no good. Stop implementing new features like crazy and work on workflows... These week another extension died "Approval".
Sorry this is off-topic, but it's a problem we have when using Nextcloud since the early days.

ypatios:
We have also been negatively affected by this change. As a relatively large org with about 60 people, we regularly use several shared NC calendars and we are used to being able to edit the attendee list.
Please consider fixing the invitation issue instead of removing this sensible feature (sensible because if you have edit rights in a calendar you should be able to add/invite attendees).
Thanks for reading

artforlife:
> The lack of functionalities centered on Projects and not only on Users is a problem for project managers. And we, as a user, have to invent some workarounds. If these workarounds are being limited then the all use of Nextcloud for these things is being limited.
> 

I too find the absence of this functionality highly inconvenient. I am managing a team and was sure that my assistant would be able to add people to events on a shared (mine) calendar. I was quite surprised to find out that this functionality was removed. It is quite inconvenient. We use Nextcloud because we care for our data and like the all-in-on-place approach. However, it is changes or discoveries like these that make life with Nextcloud quite frustrating. I truly hope that this gets reconsidered or fixed some other way.

kesselb:
https://github.com/nextcloud/calendar/issues/5674#issuecomment-2271896247

artforlife:
> [#5674 (comment)](https://github.com/nextcloud/calendar/issues/5674#issuecomment-2271896247)

Great news, thanks!

PR #5475, opened 2023-09-14
fix overlapping toggle

szaimen:
Close https://github.com/nextcloud/calendar/pull/5391

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -279,3 +279,11 @@
 .fc-daygrid-day-events {
 	position:relative !important;
 }
+
+// Fix week button overlapping with the toggle
+.fc-scrollgrid-sync-inner {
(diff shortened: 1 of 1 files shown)


nimishavijay (approved):
Nice! :)

st3iny (changes requested):
I like the approach but it causes the week numbers to be misaligned in all but the first rows.

grafik

st3iny (approved):
Tested and works.

raimund-schluessler:
That's how it looks for me now with Calendar 4.5.1 on Firefox 117 in Week view with German locale:

grafik

szaimen:
This seems to fix it in firefox:
```
.fc-timegrid-axis.fc-scrollgrid-shrink {
  height:65px;
}
```

Issue #5476, opened 2023-09-15
With v4.5.0 web ui loads infinitely


p5n:
### Steps to reproduce

1. nextcloud-app-calendar (1:4.4.5-1 -> 1:4.5.0-1)
2. try open calendar app

### Expected behavior

Calendar opened

### Actual behaviour

It never loads, database executes selects from oc_calendarobjects joining on oc_calendarobjects_props for 3 minutes

I have ~15 000 000 rows in calendarobjects-props and 1 500 000 in oc_schedulingobjects

~300 users has ~5-15 shared calendars

But for some users it loads successfully, cannot understand the difference

It surely regression because of rolling back to 4.4.5 fixes the problem

### Calendar app version

4.5.0

### CalDAV-clients used

firefox or any web browser

### Browser

firefox 117.0.1

### Client operating system

Arch Linux

### Server operating system

Arch Linux

### Web server

Apache

### Database engine version

PostgreSQL

### PHP engine version

PHP 8.1

### Nextcloud version

27.0.2-1

### Updated from an older installed version or fresh install

Updated from an older version

### List of activated apps

```Text
Enabled:
  - activity: 2.19.0
  - admin_audit: 1.17.0
  - bookmarks: 13.1.1
  - calendar: 4.4.5
  - calendar_resource_management: 0.5.0
  - circles: 27.0.1
  - cloud_federation_api: 1.10.0
  - comments: 1.17.0
  - contacts: 5.4.0
  - contactsinteraction: 1.8.0
  - dashboard: 7.7.0
  - dav: 1.27.0
  - deck: 1.10.0
  - end_to_end_encryption: 1.13.1
  - federatedfilesharing: 1.17.0
  - federation: 1.17.0
  - files: 1.22.0
  - files_external: 1.19.0
  - files_markdown: 2.3.6
  - files_pdfviewer: 2.8.0
  - files_rightclick: 1.6.0
  - files_sharing: 1.19.0
  - files_trashbin: 1.17.0
  - files_versions: 1.20.0
  - forms: 3.3.0
  - impersonate: 1.14.0
  - keeweb: 0.6.13
  - logreader: 2.12.0
  - lookup_server_connector: 1.15.0
  - notes: 4.8.1
  - notifications: 2.15.0
  - notify_push: 0.6.3
  - oauth2: 1.15.1
  - onlyoffice: 8.1.0
  - photos: 2.3.0
  - polls: 3.4.1
  - previewgenerator: 5.1.0
[...]

PR #5498, opened 2023-09-28
revert: Augment the category menu by system tags and already used categories

miaulalala:
Ref b3af4c3


ChristophWurst:
This should also go into main (or even main first, then backport) so we don't have a blocked main branch for the next minor release.

Critical paths to test and offer a migration will be events that now have a custom category set. They should at least render and be editable. Ideally the category should be removable.

miaulalala:
Codecov upload is failing

miaulalala:
Tested the categories - I can manually add a separate category 'testing' and remove it again.

ChristophWurst:
Cc @rotdrop 

This reverts https://github.com/nextcloud/calendar/pull/5161 to remedy https://github.com/nextcloud/calendar/issues/5476

ChristophWurst (commented):
Change looks good

rotdrop:
> Cc @rotdrop
> 
> This reverts #5161 to remedy #5476

Sad story, but I understand that this is a problem. I'll have a look at the SQL things. Perhaps it would be possible to have some lazy autocompletion setup on-demand instead of just loading everything from the DB at page load.

ChristophWurst:
The change can be brought back once the performance issues are gone.

tcitworld:
> Sad story, but I understand that this is a problem. I'll have a look at the SQL things. Perhaps it would be possible to have some lazy autocompletion setup on-demand instead of just loading everything from the DB at page load.

Yes, that is a good lead. Add with that some cache and it might be acceptable.

kesselb (approved):

PR #5532, opened 2023-10-15
Reset attendees participation on eventDrop

onny:
Following change resets participation state of all attendees (except organizer) on eventDrop to NEEDS-ACTION. Considering that eventDrop would be always a significant change. This will trigger resending of an invitation.

This is a workaround for sabre-dav doesn't respecting [RFC 6638 3.2.8](https://datatracker.ietf.org/doc/html/rfc6638#section-3.2.8): https://github.com/sabre-io/dav/issues/1282

Addresses: https://github.com/nextcloud/server/issues/13862

### DIFF
--- a/src/fullcalendar/interaction/eventDrop.js
+++ b/src/fullcalendar/interaction/eventDrop.js
@@ -68,6 +68,19 @@ export default function(store, fcAPI) {
 			return
 		}
 
+		// Reset attendees participation state to NEEDS-ACTION, since eventDrop
+		// is always a signification change
+		// Partly a workaround for Sabre-DAV not respecting RFC 6638 3.2.8, see
+		// https://github.com/sabre-io/dav/issues/1282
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> Fixes: [nextcloud/server#13862](https://github.com/nextcloud/server/issues/13862)

I'd argue that it doesn't fully but only for events modified through our client and when they are dragged and dropped. Changes through other channels will still have the bug.

:+1: for the workaround though

tcitworld:
Let's not close the server issue, if we can't merge https://github.com/sabre-io/dav/pull/1443 in sabre we can always patch it in 3rdparty.

miaulalala (approved):

onny:
@miaulalala Signed my commit, tests should now hopefully pass :)

st3iny (approved):
Tested and works.

miaulalala:
Unfortunately tests are failing

onny:
thank you for noticing! fixed the tests :) should still work as intended

Issue #5507, opened 2023-10-02
Add PHP8.3 support


ChristophWurst:
### Is your feature request related to a problem? Please describe.

PHP8.3 is coming. As an admin of Nextcloud Hub 7 I want to be able to run Nextcloud and its apps with PHP8.3.

### Describe the solution you'd like

Add PHP8.3 support.

### Describe alternatives you've considered

N/a

### Additional context

N/a

PR #5542, opened 2023-10-20
Add PHP8.3 support

JohannesGGE:
Fix: #5507


ChristophWurst (approved):
PHPUnit output is clean

PR #5587, opened 2023-11-22
fix(appointments): switch to/from order for the booking information email (to the organizer)

pjrobertson:
I noticed that the confirmation email I receive when somebody else books an appointment has the from/to names mixed up.

Right now it looks like this:

Screen Region 2023-11-22 at 17 34 44

To clarify, it's set as:

```
From: MY NAME <no-reply@nextcloud.address>
To: PERSON WHO BOOKED <my-email@domain.com>
```

but it should be:
```
From: PERSON WHO BOOKED <no-reply@nextcloud.address>
To: MY NAME <my-email@domain.com>
```

This is really confusing, the names are mixed so it looks like the email has been send *from* me but *to* the person who booked. That's the case for all the other emails (to the person who booked) but this one should be the other way around :)

### DIFF
--- a/lib/Service/Appointments/MailService.php
+++ b/lib/Service/Appointments/MailService.php
@@ -284,12 +284,12 @@ class MailService {
 		if ($toEmail === null) {
 			throw new ServiceException('Organizer has no email set');
 		}
-		$fromName = $user->getDisplayName();
+		$toName = $user->getDisplayName();
 
 		$sys = $this->getSysEmail();
 		$message = $this->mailer->createMessage()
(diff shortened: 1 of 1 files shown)


pjrobertson:
Hey @ChristophWurst - are there any other edits/checks I need to complete for this to get merged? I see two tests are failing , but that seems to just be the "upload coverage" step.

Thanks as always :)

miaulalala (approved):
Heh that looks fine. Thanks for taking the time!

pjrobertson:
Thanks Anna! :)

PR #5167, opened 2023-04-27
feat(webcal): Thunderbird holiday subscription picker

ChristophWurst:
For https://github.com/nextcloud/calendar/issues/962

Bildschirmfoto vom 2023-04-28 10-36-37

## Todo
- [x] Design work
- [x] Subscribe handling

Issue #962, opened 2018-11-27
Easily add holidays


ryanprior:
There ought to be an easy, obvious interface to add a standard holiday calendar to your NextCloud.

In https://github.com/nextcloud/server/issues/1363 the request was made to "make national holidays available in calendar" and this was implemented by allowing webcal subscriptions. That technically works but it has these drawbacks:

* NextCloud cedes control over the user experience to arbitrary third party holiday webcal providers, whose UX is atrocious
* somebody who isn't familiar with webcal will not think to solve the problem this way
* for people who know about webcal, it's still not trivial to find the right calendar out there on the internet
* a webcal can send spam to users or otherwise abuse the relationship, and has little incentive not to do so since NextCloud offers no alternative

A way to resolve this is to provide a convenient interface for users to find a holiday calendar that's relevant to them and add it in a few clicks, without having to know about the implementation. Advantages of this approach:

* NextCloud can ensure a good user experience
* holiday calendars are discoverable and anyone can use them easily
* basic calendar information like holidays are kept private, with no unnecessary reliance on third party data providers

<bountysource-plugin>

## Acceptance criteria

* Use upstream data from https://www.thunderbird.net/en-US/calendar/holidays/
* New entry in *+ New calendar* -> modal with list of countries/regions -> add as new subscription
* UI is in Calendar app. If Calendar is not installed/enabled the feature is not visible
* Optional: Subscriptions should be shareable (example: admin adds company holidays and shares with everyone)
  * Requires https://github.com/nextcloud/server/issues/1600

## Work packages

- [x] MVP: Add a link to the subscription menu that sends people to the Thunderbird subscription site https://github.com/nextcloud/calendar/pull/5166
- [x] Main implementation that fulfills the acceptance criteria https://github.com/nextcloud/calendar/pull/5167
- [x] https://github.com/nextcloud/documentation/issues/10018

## Follow-ups

* https://github.com/nextcloud/calendar/issues/5168

georgehrke:
I would really like to ship holiday calendars, but the biggest problem is: Where to get the data from?

It should be obvious that we can’t copy and paste the Google Holiday calendars for copyright reasons.

@ryanprior Are you aware of any database that offers holiday data under a free (foss) license?

tcitworld:
So it's basically unusable until the data is filled into Wikidata or someone finds an existing tool to scrape it from Wikipedia. :/

ryanprior:
[...]

PR #5614, opened 2023-12-08
Allow to define and subscribe to custom public calendars

escoand:
This PR enables the admin to define custom public calendars to be subscribed.
image
image

linked to #962, #5167

### DIFF
--- a/src/components/Subscription/PublicCalendarSubscriptionPicker.vue
+++ b/src/components/Subscription/PublicCalendarSubscriptionPicker.vue
@@ -21,22 +21,29 @@
 
 <template>
 	<NcModal @close="$emit('close', $event)">
-		<div class="holiday-subscription-picker">
-			<h2>{{ t('calendar', 'Public holiday calendars') }}</h2>
-			<p class="holiday-subscription-picker__attribution">
-				{{ t('calendar', 'Public holiday calendars are provided by Thunderbird. Calendar data will be downloaded from {website}', { website: 'thunderbird.net' }) }}
+		<div class="public-calendar-subscription-picker">
+			<h2 v-if="showHolidays">
+				{{ t('calendar', 'Public holiday calendars') }}
+			</h2>
+			<h2 v-else>
+				{{ t('calendar', 'Public calendars') }}
+			</h2>
+			<p v-if="showHolidays" class="holiday-subscription-picker__attribution">
+				{{ t('calendar',
+					'Public holiday calendars are provided by Thunderbird. Calendar data will be downloaded from {website}',
+					{ website: 'thunderbird.net' }) }}
 			</p>
-			<div v-for="calendar in calendars" :key="calendar.source" class="holiday-subscription-picker__region">
-				<div class="holiday-subscription-picker__region__name">
-					<h3>{{ calendar.country }}</h3>
-					<div class="holiday-subscription-picker__region__name__subline">
-						{{ calendar.datespan }}
+			<div v-for="calendar in calendars" :key="calendar.source" class="public-calendar-subscription-picker__region">
+				<div class="public-calendar-subscription-picker__region__name">
+					<h3>{{ calendar.name }}</h3>
+					<div v-if="calendar.description" class="public-calendar-subscription-picker__region__name__subline">
+						{{ calendar.description }}
 					</div>
-					<div class="holiday-subscription-picker__region__name__subline">
+					<div v-if="calendar.authors" class="public-calendar-subscription-picker__region__name__subline">
 						{{ t('calendar', 'By {authors}', { authors: calendar.authors }) }}
 					</div>
 				</div>
-				<div class="holiday-subscription-picker__region__subcribe">
+				<div class="public-calendar-subscription-picker__region__subcribe">
 					<NcButton :disabled="loading || subscribing[calendar.source] || subscribed[calendar.source]"
 						@click="subscribe(calendar)">
 						{{ subscribed[calendar.source] ? t('calendar', 'Subscribed') : t('calendar', 'Subscribe') }}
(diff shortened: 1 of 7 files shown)


escoand:
Don't know how to fix the php tests...

src/components/Subscription/CalendarOfInterestSubscriptionPicker.vue
```diff
@@ -92,21 +97,19 @@ export default {
...
+		async subscribe(calendar) {
 			try {
-			  this.subscribing[calendar.source] = true
+				this.subscribing[calendar.source] = true
 
 				await this.$store.dispatch('appendSubscription', {
-					displayName: t('calendar', 'Holidays in {region}', {
-					  region: calendar.country,
-					}),
+					displayName: t('calendar', calendar.name),
```
miaulalala, line 105: Since the calendar name is variable, this will not need a translation unless you add a string such as "Calendar of interest {calendar.name}" or some such

tests/php/unit/Controller/ViewControllerTest.php
```diff
@@ -112,6 +112,7 @@ public function testIndex(): void {
 				['calendar', 'forceEventAlarmType', '', 'forceEventAlarmType'],
 				['dav', 'allow_calendar_link_subscriptions', 'yes', 'no'],
 				['calendar', 'showResources', 'yes', 'yes'],
+				['calendar', 'calendarsOfInterest', ''],
```
miaulalala, line 115: tests can be fixed by increasing the counter:

`$this->config->expects(self::exactly(16))`

here and in the next method too.

miaulalala (changes requested):
I do like the idea but I think it would be better if the calendars from the config were appended instead of overwriting the complete file.

@jancborchardt your opinion on the name change?

src/components/Subscription/CalendarOfInterestSubscriptionPicker.vue
escoand: I'm not quiet sure but it seems to me you could create a translation like `"Public holidays in %s": "Feiertage in %s"` to translate every name with that prefix.

escoand:
> I do like the idea but I think it would be better if the calendars from the config were appended instead of overwriting the complete file.

Prepending was also my first implementation but in my case me as admin don't care of all the holiday calendars and there is no other way to get rid of them. If I would be interested I could just add the right ones to the settings.

Appending would be the worst solution IMHO because I want to add something specific but the user always has to scroll down.
Some kind of sorting could maybe be helpful.

miaulalala:
Another option would be to add a new button under the holiday calendar like this:

image

The initial state and everything would stay the same, and the button would only show when any custom calendars are set in the config.

Would leave the holiday feature as is and make the custom calendars easily discoverable. What do you think @escoand ?

jancborchardt:
@miaulalala I like your suggestion and wording a lot! :)

escoand:
Should I add an other picker dialog or make a generic one out of the holiday picker?

miaulalala:
> Should I add an other picker dialog or make a generic one out of the holiday picker?

Generic is fine, whatever you're most comfortable with.

ChristophWurst:
Please document the feature in https://github.com/nextcloud/documentation/ for https://docs.nextcloud.com/server/stable/admin_manual/groupware/calendar.html#subscriptions :pray:

src/components/Subscription/PublicCalendarSubscriptionPicker.vue
```diff
@@ -55,18 +62,38 @@ import { mapGetters } from 'vuex'
...
+					region: calendar.country,
+				}),
+				name: calendar.country,
+				description: calendar.datespan,
+				source: 'https://www.thunderbird.net/media/caldata/' + calendar.filename,
+			}))
+		} else {
+			try {
+				const state = loadState('calendar', 'publicCalendars')
+				calendars = JSON.parse(state)
```
tcitworld, line 91: This should also validate the parsed content to make sure each element has valid `displayName`, `name`, `description` and more importantly `source` attributes.

src/components/Subscription/PublicCalendarSubscriptionPicker.vue
```diff
@@ -21,22 +21,29 @@
...
-			<p class="holiday-subscription-picker__attribution">
-				{{ t('calendar', 'Public holiday calendars are provided by Thunderbird. Calendar data will be downloaded from {website}', { website: 'thunderbird.net' }) }}
+		<div class="public-calendar-subscription-picker">
+			<h2 v-if="showHolidays">
+				{{ t('calendar', 'Public holiday calendars') }}
+			</h2>
+			<h2 v-else>
+				{{ t('calendar', 'Public calendars') }}
+			</h2>
+			<p v-if="showHolidays" class="holiday-subscription-picker__attribution">
```
tcitworld, line 31: I'd appreciate if the "custom" public calendar section also had a paragraph explaining the below calendars have been suggested by the sever administrator.

src/components/Subscription/PublicCalendarSubscriptionPicker.vue
escoand: With what result if one or more are invalid?

tcitworld: You could simply filter the invalid elements out of the list.

src/components/Subscription/PublicCalendarSubscriptionPicker.vue
```diff
@@ -21,22 +21,32 @@
...
 			</p>
-			<div v-for="calendar in calendars" :key="calendar.source" class="holiday-subscription-picker__region">
-				<div class="holiday-subscription-picker__region__name">
-					<h3>{{ calendar.country }}</h3>
-					<div class="holiday-subscription-picker__region__name__subline">
-						{{ calendar.datespan }}
+			<p v-else class="holiday-subscription-picker__attribution">
+				{{ t('calendar', 'These public calendars are suggested by the sever administrator. Calendar data will be downloaded from the respective website.') }}
+			</p>
+			<div v-for="calendar in calendars" :key="calendar.source" class="public-calendar-subscription-picker__region">
```
tcitworld, line 39: Can you handle `calendars` being empty? Using [`NcEmptyContent`](https://nextcloud-vue-components.netlify.app/#/Components/NcEmptyContent).

src/components/Subscription/PublicCalendarSubscriptionPicker.vue
```diff
@@ -21,22 +21,32 @@
...
+				<div class="public-calendar-subscription-picker__region__name">
+					<h3>{{ calendar.name }}</h3>
```
tcitworld, line 41: I'd use 
```suggestion
					<h3>{{ calendar.displayName || calendar.name }}</h3>
```

as it will be used for the actual calendar display name when it's being added. It's likely an UI for the admins could make use of that (a source name vs a display name for the users).

escoand: But this mean a change to the existing public holiday calendars too:
image

tcitworld: You're right, please disregard.

tcitworld (approved):

miaulalala (approved):
nicely done! Thanks for the PR!

PR #5323, opened 2023-06-16
Fix: replace NcMultiselect with NcSelect

hamza221:
fix #5320

PR #5645, opened 2023-12-26
don't filter results while sharing

akhil1508:
- Share only works by display name right now
- Search by email to share calendar is broken completely as there is a `filterBy` option in [vue-select](https://vue-select.org/api/props.html#filterby) whose return value defaults to `return (label || '').toLocaleLowerCase().indexOf(search.toLocaleLowerCase()) > -1`
  - `label` in  our code is set to `displayName` so only search by `displayName` works

- Since we already get our results by searching(DAV & OCS), I don't see why we need the `filterBy` option in the first place
  - So I just make the return value `true`
  - In case we do need the option, we should make it check for both `displayName` and `email` attributes

### DIFF
--- a/src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
+++ b/src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
@@ -33,6 +33,7 @@
 			class="sharing-search__select"
 			:class="{ 'showContent': inputGiven, 'icon-loading': isLoading }"
 			:user-select="true"
+			:filter-by="filterResults"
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Not verified but if the limitation comes from with NcSelect we introduced this limitation with https://github.com/nextcloud/calendar/pull/5323. cc @hamza221

hamza221 (approved):

DnlKI:
Relates to #5691

Right now, one can only search for mail addresses and not for names nor groups.

PR #5648, opened 2023-12-28
Migrate/to nextcloud vue 8

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/5619

Progress:

- [x] Change title to name
- [x] Change MultiSelect to Select props
- [x] Work out errors with Select when `multiple = true`

PR #5649, opened 2023-12-29
fix Circular structure in calendar object

hamza221:
Dav Object contains circular reference which breaks NcSelect  
image
part of #5648

### DIFF
--- a/src/components/Shared/CalendarPicker.vue
+++ b/src/components/Shared/CalendarPicker.vue
@@ -2,11 +2,11 @@
 	<Select label="displayName"
 		input-id="url"
 		:disabled="isDisabled"
-		:options="calendars"
-		:value="value"
+		:options="options"
+		:value="removeCircularStructure(value)"
 		:multiple="multiple"
 		@input="change"
-		@remove="remove">
+		@option:deselected="remove">
 		<template #option="option">
 			<CalendarPickerOption :color="option.color"
 				:display-name="option.displayName"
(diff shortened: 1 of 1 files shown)


src/components/Shared/CalendarPicker.vue
```diff
@@ -55,6 +55,12 @@ export default {
 			// for calendars where only one calendar can be selected, disable if there are < 2
 			return this.multiple ? this.calendars.length < 1 : this.calendars.length < 2
 		},
+		options() {
+			return this.calendars.map(calendar => {
```
tcitworld, line 59: You could just reuse your new method in the map, right ?

hamza221: True 😂

hamza221:
Eslint unrelated to this Pr

src/components/Shared/CalendarPicker.vue
```diff
@@ -55,6 +55,9 @@ export default {
 			// for calendars where only one calendar can be selected, disable if there are < 2
 			return this.multiple ? this.calendars.length < 1 : this.calendars.length < 2
 		},
+		options() {
+			return this.calendars.map(calendar => this.removeCircularStructure(calendar))
```
tcitworld, line 59:
```suggestion
			return this.calendars.map(this.removeCircularStructure)
```
Does this work ?

tcitworld (approved):

GVodyanov (approved):
Looks good

Issue #5655, opened 2024-01-02
`files.forEach is not a function` when attaching local files to events


ChristophWurst:
### Steps to reproduce

1. Create an event
2. Open the sidebar
3. Pick a local attachment

### Expected behavior

Attachment uploads

### Actual behaviour

JS error with no visual feedback

### Calendar app version

4.7

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

ChristophWurst:
https://stackoverflow.com/questions/40902437/cant-use-foreach-with-filelist

PR #5656, opened 2024-01-02
fix(attachments): Convert FileList to array for FireFox

ChristophWurst:
Firefox does not have FileList.forEach.

Fixes https://github.com/nextcloud/calendar/issues/5655

Additionally, I've added try-catch with an error toast for when upload still fails.

## How to test

1) Start Firefox
2) Create an event
3) Open the sidebar
4) Upload a local attachment


st3iny (approved):
Tested and wörks.

Issue #5664, opened 2024-01-04
Appoinment config confirmation modal uses two button sizes and styles


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Create an appointment config in the left sidebar
3. Save it

### Expected behavior

Consistent button design

### Actual behaviour

![image](https://github.com/nextcloud/calendar/assets/1374172/30a86a26-5768-4c41-882f-5ee166f489ce)

### Calendar app version

4.7

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

ChristophWurst:
One is an `<a>`, the other a `NcButton`: https://github.com/nextcloud/calendar/blob/v4.6.2/src/components/AppointmentConfigModal/Confirmation.vue#L31-L40

PR #5672, opened 2024-01-08
fix(appointments): Fix button style

JohannesGGE:
Fix: #5664 

|before|after|
|---|---|
|image|image|


ChristophWurst (approved):
:+1:  :balance_scale:

PR #5680, opened 2024-01-10
fix(appointments): Rate limit config creation and booking

ChristophWurst:
Limits the number of bookings a single user can do.


ChristophWurst:
Psalm fails

PR #5683, opened 2024-01-10
[stable3.5] fix(appointments): Rate limit config creation and booking

ChristophWurst:
Manual backport of https://github.com/nextcloud/calendar/pull/5680.

NcNotecard doesn't exist for nc/vue 5 so I simplified the HTML.


st3iny (approved):
I tested creating a new config and booked it and it still works. I also tested the rate limit for the booking endpoint via curl and it also blocked me with 429 after some tries.

hamza221 (approved):

Issue #5687, opened 2024-01-11
Talk pariticipant automation implicitly enables guests if emails are invited to a room


ChristophWurst:
### Steps to reproduce

1. Create an event
2. Create a private Talk room
3. Paste the Talk room URL in the LOCATION field of the event
4. Add a Nextcloud user as event attendee
5. Add an email address (not mappable to a Nextcloud user) as event attendee

### Expected behavior

The Talk room only shows me and the other Nextcloud user

### Actual behaviour

The Talk room show me, the other Nextcloud user and an email address. On top of that, guests will be allowed

> You allowed guests

### Calendar app version

4.6

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

ChristophWurst:
The problem lies in https://github.com/nextcloud/calendar/blob/961ae74be001077be9023006d98e61d57be4f6f9/src/services/talkService.js#L94-L117. We should ignore email participants.

PR #5688, opened 2024-01-11
fix(talk): Only invite guest accounts to open Talk conversations

ChristophWurst:
Inviting guests to a private group chat has the implication that the Talk room has to be opened up for guests. This is unexpected.
This patch adjusts the logic to only invite guest users if the Talk room is already open for guests. Otherwise they are ignored.

Fixes https://github.com/nextcloud/calendar/issues/5687

## How to test

Repro steps from https://github.com/nextcloud/calendar/issues/5687

In addition, test it with an open conversation. Guests will be added.

### DIFF
--- a/src/services/talkService.js
+++ b/src/services/talkService.js
@@ -78,12 +78,14 @@ export async function updateTalkParticipants(eventComponent) {
 		return
 	}
 	try {
+		const { data: { ocs: { data: room } } } = await HTTPClient.get(generateOcsUrl('apps/spreed/api/' + apiVersion + '/', 2) + 'room/' + token)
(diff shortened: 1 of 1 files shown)


ChristophWurst:
I could first check the room status and if guests are allowed I still send these participants. Only for private rooms I ignore them.

AndyScherzinger (approved):

ChristophWurst:
> I could first check the room status and if guests are allowed I still send these participants. Only for private rooms I ignore them.

Done. Please test/review another time

nickvergessen (approved):

miaulalala:
Works, the guests are not invited to the room if it isn't open.

miaulalala (approved):

PR #5723, opened 2024-01-29
enh(appointments): add moar debug logging and overwrite app id

miaulalala:
Add moar logging because something is suspicious suspicious.

### DIFF
--- a/lib/Service/Appointments/AvailabilityGenerator.php
+++ b/lib/Service/Appointments/AvailabilityGenerator.php
@@ -31,6 +31,7 @@ use DateTimeImmutable;
 use DateTimeZone;
 use OCA\Calendar\Db\AppointmentConfig;
 use OCP\AppFramework\Utility\ITimeFactory;
+use Psr\Log\LoggerInterface;
 use function ceil;
 use function max;
 use function min;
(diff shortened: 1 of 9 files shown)


lib/Service/Appointments/AvailabilityGenerator.php
```diff
@@ -79,11 +80,17 @@ public function generate(AppointmentConfig $config,
 
 		// If we reach this state then there are no available dates anymore
 		if ($latestEnd <= $earliestStart) {
+			$this->logger->debug('Appointment config ' . $config->getToken() . ' has {latestEnd} as latest end but {earliestStart} as earliest start. No slots available.', [
+				'latestEnd' =>  $latestEnd,
+				'earliestStart' => $earliestStart,
+				'app' => 'appointments'
```
tcitworld, line 86: Would still prefix it like the following to avoid confusion with https://apps.nextcloud.com/apps/appointments

```suggestion
				'app' => 'calendar-appointments'
```

miaulalala: Good point.

tcitworld (approved):

PR #5747, opened 2024-02-08
fix(appointments): round up timeslots to next pretty time

miaulalala:


lib/Service/Appointments/AvailabilityGenerator.php
```diff
@@ -67,7 +67,8 @@ public function generate(AppointmentConfig $config,
 		// E.g. 5m slots should only be available at 10:20 and 10:25, not at 10:17
 		//      when the user opens the page at 10:17.
 		// But only do this when the time isn't already a "pretty" time
-		if ($earliestStart % $config->getIncrement() !== 0) {
+		$prettyTimes = range(0,2700, 300);
```
ChristophWurst, line 70:
```suggestion
		$prettyTimes = range(0, 2700, 300);
```

lib/Service/Appointments/AvailabilityGenerator.php
```diff
@@ -67,7 +67,8 @@ public function generate(AppointmentConfig $config,
...
+		if (!in_array($earliestStart % $config->getIncrement(), $prettyTimes)) {
```
ChristophWurst, line 71:
```suggestion
		if (!in_array($earliestStart % $config->getIncrement(), $prettyTimes, true)) {
```

lib/Service/Appointments/AvailabilityGenerator.php
```diff
@@ -67,7 +67,8 @@ public function generate(AppointmentConfig $config,
 		// E.g. 5m slots should only be available at 10:20 and 10:25, not at 10:17
 		//      when the user opens the page at 10:17.
 		// But only do this when the time isn't already a "pretty" time
-		if ($earliestStart % $config->getIncrement() !== 0) {
+		$prettyTimes = range(0, 2700, 300);
+		if (!in_array($earliestStart % $config->getIncrement(), $prettyTimes, true)) {
```
ChristophWurst, line 71: I'm not sure if there are side effects. The old logic was always specific to the length of the increment with modulo being 0. Example: half hour increments are "beautiful" at :00 and :30.
Now they are "beautiful" at :00, :05, :10, ... :45. Why is it always those 5minute steps regardless of the increment? And what about :50 and :55?

ChristophWurst: https://3v4l.org/0MXWN

I think only the :00 and :30 should pass the beauty test for a 30m increment

miaulalala: well the earliest start time for hourly slots with an availability of WED: 10:10 is 10:10. We can either adjust the slot generation to start at 11:00 earliest for full hour slots, 10:30 for half hour slots etc - which is nice - or we can allow a user to start a slot at any 5 minute increments independent of start time - which is also nice.

Forgot to add 50 and 55 to the range :sweat_smile:

ChristophWurst: That is not intended. If you book 30min slots at 10:11 the first slot should be 10:30, not 10:15.

miaulalala: Ok then we need to modify the slot generation some more to adjust the earlies start time.

miaulalala: Slot generation is using the first available timestamp from the JSON to extrapolate all slots. So if you enter 10:10 as your first available timestamp, the date filtering will extrapolate within a whole free day - and do 10:10 to 10:40 etc.

st3iny (approved):
I tested it and it works as I would expect it to work.

Unit test failures seem to be related though.

PR #5748, opened 2024-02-08
fix(appointments): add date to booking detail view

miaulalala:
Before:

image

After:

image


ChristophWurst (approved):

miaulalala:
Codecov is having a wobble

ChristophWurst:
The warnings are legit. The new code is not tested so the coverage drops

Issue #5788, opened 2024-02-21
Public holiday calendars menu broken


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Click *New calendar*
3. Click *Add public holiday calendar*

### Expected behavior

Public holiday calendar modal opens

### Actual behaviour

Nothing happens

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

ChristophWurst:
This could be a regression of the nextcloud/vue upgrade. @GVodyanov could you have a look please?

ChristophWurst:
Not reproducible anymore. Putting on hold until this pops up again and we know what causes it.

PR #5789, opened 2024-02-21
Fix PublicCalendarSubscriptionPicker import

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/5788

I'm honestly not sure why it was previously imported in such a weird way:
`PublicCalendarSubscriptionPicker: () => import(/* webpackChunkName: "public-calendar-subscription-picker" */ '../../Subscription/PublicCalendarSubscriptionPicker.vue'),`

But importing it normally seems to work without any issues, so I guess it can go

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListNew.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListNew.vue
@@ -128,7 +128,6 @@ import {
 } from '@nextcloud/dialogs'
 
 import { uidToHexColor } from '../../../utils/color.js'
-import PublicCalendarSubscriptionPicker from '../../Subscription/PublicCalendarSubscriptionPicker.vue'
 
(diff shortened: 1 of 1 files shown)


st3iny (changes requested):
This is a performance optimization. By doing a dynamic import we allow lazily loading the whole component instead of including it in the main bundle. The chunk will be loaded when the component is rendered for the first time.

Have a look at [the vue docs](https://v2.vuejs.org/v2/guide/components-dynamic-async#Async-Components) for more information.

GVodyanov:
> This is a performance optimization. By doing a dynamic import we allow lazily loading the whole component instead of including it in the main bundle. The chunk will be loaded when the component is rendered for the first time.
> 
> Have a look at [the vue docs](https://v2.vuejs.org/v2/guide/components-dynamic-async#Async-Components) for more information.

Ah that makes more sense 😅! I took a look at it and left the same import, there was just a small issue with the address

Issue #5791, opened 2024-02-22
When duration greater than incremtent: "Could not book the appointment. Please try again later or contact the organizer."


bentuna:
### Steps to reproduce

1. create appointment where the duration is greater than the increment, i.e. duration: 90 min, increment: 60 min
2. try to book slots
3. see how it works for some slots, and others show the error

### Expected behavior

All available slots are bookable.

### Actual behaviour

Some available slots are not bookable, instead the error from the title of the issue is displayed:
- in every second slot from the available slot list, when duration is double of increment
- in all other cases where duration > increment
So the error is present with every duration above 60 minutes (because currently incremtent cannot be set greater than 60 min)

### Calendar app version

4.6.5

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

28.0.2

### Updated from an older installed version or fresh install

None

small1:
This might happen in some other cases as well. Trying to figure it out on one instance.

small1:
From the log file when trying to book an appointment. It is available to book and nothing else is on that slot.

```
{"reqId":"4aCaCYEFfOeLtuLPPink","level":0,"time":"2024-02-24T09:17:26+00:00","remoteAddr":"*.*.*.*","user":"--","app":"calendar","method":"POST","url":"/apps/calendar/appointment/1/book","message":"Appointment config JmzM97mzCmxd has 1 intervals that result in 0 possible slots. 0 slots remain after the daily limit. 0 available slots remain after conflict checking.","userAgent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36","version":"28.0.2.5","data":{"app":"calendar"}}
[...]

PR #5820, opened 2024-02-29
fix(appointments): Round by increment

miaulalala:
Fixes https://github.com/nextcloud/calendar/issues/5791

The rounding by length might lead to more efficient slot generation but it breaks appts.


ChristophWurst (approved):

pjrobertson:
Thanks @miaulalala - this has plagued me for a while 🎉

PR #5834, opened 2024-03-07
fix(appointments): slot generation deviation for increments that don't fit modulo

miaulalala:
Thanks @nickvergessen who figured out that the `ciel()` function wasn't working properly because `(int)` is stronger than `x*y`, so the results were [not correct](https://www.youtube.com/watch?v=Jk6jVl1fAn0)


ChristophWurst (approved):
:seal:()

nickvergessen (approved):

Issue #2617, opened 2020-09-28
Allow configuration of the email address to send invitations from


Peque:
**Is your feature request related to a problem? Please describe.**

When creating an event and inviting people, an invitation is sent to them. The invitation is always sent from nextcloud@domain.tld and the logo of the Nextcloud instance is included as well.

I may not want to tell them that I am using Nextcloud, nor any other service, so:

- I would rather not share with them the details of the Nextcloud instance I am using
- I would rather not share with them the instance's configured email address
- I would rather not include Nextcloud or the instance's configured logo

**Describe the solution you'd like**

- Be able to configure an email address to send those invitations from, ideally, this should be configurable per-calendar (i.e.: I may have a personal and a professional calendar and I would like to send invitations from different email addresses in those cases)
- Be able to remove the Nextcloud/instance logo from the invitation, or be able to replace that logo with my user avatar, or be able to define my own invitation template

**Additional context**

See also: https://github.com/nextcloud/calendar/issues/2591

OrfeasLitos:
I luckily found a workaround and thus will unfortunately stop working on this. The whole saga can be found at https://help.nextcloud.com/t/set-per-calendar-organizer/174879/10.

SebastianKrupinski:
Hi @Peque,

We are in the process of implementing changes that will send invitation messages from the users mail app account instead of the system account, which will also have the added benefit of proper Accept/Decline status updates in the Calendar.

But at the moment there is no planned feature to be able to set a outbound email address per calendar. But I will notify the appropriate person of your request, as this feature request has been brought up several times now.

matclab:
Hi @SebastianKrupinski, 
It is nice to know that the *send invitation messages from the users mail address* feature is under work, as it causes some surprise from our the user to receive an invitation from another address. 

Is their an existing issue to follow in order to know when it will be available ? 

Best regards,

PR #5884, opened 2024-03-28
[WIP] Edit email on a per-calendar basis

OrfeasLitos:
This intends to partially address #2617.

The overall idea is that the `calendarEmail` string attribute will be added to each calendar, which will be modifiable, just like the calendar name currently is. When an invitation is prepared, this attribute will be checked. If it isn't empty, the set value will be used, otherwise the organizer's email will be used as the organizer name in the invitation.

I would also like to add the ability to disable sending invitations from Nextcloud and let my calendar client (e.g., Thunderbird) take care of it, using the invitation with the user-selected email as described above. I have not yet implemented this part at all.

Is this a valuable feature? Does this approach make sense? If so, could you please point me to the code that prepares the invitation? I guess it interacts with `dav`, but it's unclear to me how.

Thanks in advance for giving me the pointers I need to contribute!

### DIFF
--- a/src/appointments/main-booking.js
+++ b/src/appointments/main-booking.js
@@ -44,6 +44,7 @@ Vue.prototype.$n = translatePlural
 
 const config = loadState('calendar', 'config')
 const userInfo = loadState('calendar', 'userInfo')
+const calendarEmail = loadState('calendar', 'calendarEmail')
 const visitorInfo = loadState('calendar', 'visitorInfo', {
 	displayName: '',
 	email: '',
@@ -55,6 +56,7 @@ export default new Vue({
 		props: {
 			config,
 			userInfo,
+			calendarEmail,
 			visitorInfo,
 		},
 	}),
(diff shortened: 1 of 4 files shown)


miaulalala:
Hi,

thanks for opening a PR!

I don't think setting the email for the whole calendar will work. There is no single email property for an entire Calendar Collection, but there is the [calendar-user-address-set property](https://datatracker.ietf.org/doc/html/rfc6638#section-2.4.1), which contains a set of emails for each calendar principal. I don't thin it will achieve what I think you want to achieve, though.

Have you looked into the RFCs for calendaring? If not, I'd recommend reading up on them so the whole principal / calendar / CalDAV thing makes sense. It's quite a ride but it will help understanding how it works. All of our supported RFCs are here: https://github.com/nextcloud/calendar/wiki/Developer-Resources/ 

If you have any questions we have a public Talk room you can join as well: https://cloud.nextcloud.com/call/5qb8fujz

What I would do to make it work is implement support for the `EMAIL` property defined by [RFC 7986](https://datatracker.ietf.org/doc/html/rfc7986#section-6.2), so an `ATTENDEE` can decide which email will be used to contact the `ORGANIZER` - that should address some of the concerns mentioned in the issue. The only concern I have is that I don't know how good the interoperability is with other Calendar Clients, so this would have to be tested properly.

Cheers :)

OrfeasLitos:
I luckily found a workaround and thus will unfortunately stop working on this. The whole saga can be found at https://help.nextcloud.com/t/set-per-calendar-organizer/174879/10.

Issue #5746, opened 2024-02-08
🍱 Interactive widget for calendars


ChristophWurst:
## Is your feature request related to a problem? Please describe.

As a Calendar user I want to embed a calendar in a Text document, Collectives page or Talk message.

## Describe the solution you'd like

Offer an interactive widget for calendars. Only people with read access to my calendar should see any contents. For others the widgets has to be empty.

![image](https://github.com/nextcloud/calendar/assets/1374172/d500e885-e58f-4812-b92b-6c3e084259fc)

## Implementation

Shared calendars can be embedded into iframes. Ideally we can leverage this feature.

## Work packages

- [x] Embed shared calendars read-only (accessible to anyone)
- [x] Embed private calendars read-only (but interactive with click handling and navigation)
- [ ]  User docs

## Follow-ups

- Embed private calendars with write support (only for the owner and sharees with write permission)

## Describe alternatives you've considered

N/a

## Additional context

* https://github.com/nextcloud/server/issues/42550
* https://github.com/nextcloud/calendar/issues/5701

ChristophWurst:
@hamza221 could you please check what's left to do? User docs?

hamza221:
and the text styling bug, or should I open a new issue for it?
<img width="932" alt="image" src="https://github.com/nextcloud/calendar/assets/40746210/5ad7926c-c02f-4d4c-ab17-4e443e378dd7">

ChristophWurst:
yes we can tackle that separately :+1:

hamza221:
> yes we can tackle that separately 👍

#5970

PR #5891, opened 2024-04-02
Feat: calendar widget for private calendars

hamza221:
For  https://github.com/nextcloud/calendar/issues/5746

## How to test
1. Create a new calendar
2. Add events to the calendar
3. copy internal link of a calendar
4. Open Talk or Text
5. Paste the link

PS: try testing with multiple calendar, mixing private and public link in one/multiple conversations

### DIFF
--- a/lib/Reference/ReferenceProvider.php
+++ b/lib/Reference/ReferenceProvider.php
@@ -73,24 +73,55 @@ class ReferenceProvider extends ADiscoverableReferenceProvider {
 	public function matchReference(string $referenceText): bool {
 		$start = $this->urlGenerator->getAbsoluteURL('/apps/' . Application::APP_ID);
 		$startIndex = $this->urlGenerator->getAbsoluteURL('/index.php/apps/' . Application::APP_ID);
+		if (preg_match('/^' . preg_quote($start, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1 || preg_match('/^' . preg_quote($startIndex, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1) {
+			return true;
+		}
+		
+		$start = $this->urlGenerator->getAbsoluteURL('/remote.php/dav/calendars');
+		if (preg_match('/^' . preg_quote($start, '/') . '\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+\/$/i', $referenceText) === 1) {
+			return true;
+		}
 
-		return preg_match('/^' . preg_quote($start, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1 || preg_match('/^' . preg_quote($startIndex, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1;
+		return false;
 	}
 
 	public function resolveReference(string $referenceText): ?IReference {
 		if ($this->matchReference($referenceText)) {
-			$token = $this->getCalendarTokenFromLink($referenceText);
-
+			$type = $this->getType($referenceText);
 			$reference = new Reference($referenceText);
 			$reference->setTitle('calendar');
-			$reference->setDescription($token);
-			$reference->setRichObject(
-				'calendar_widget',
-				[
-					'title' => 'calendar',
-					'token' => $token,
-					'url' => $referenceText,]
-			);
+			$reference->setDescription('calendar widget');
+
+			switch ($type) {
+				case 'public':
+					$token = $this->getCalendarTokenFromLink($referenceText);
+					$url = $this->getUrlFromLink($token, 'public');
+					$reference->setRichObject(
+						'calendar_widget',
+						[
+							'title' => 'calendar',
+							'token' => $token,
+							'isPublic' => true,
+							'url' => $url,
+						]
+					);
+					break;
+				case 'private':
+					$url = $this->getUrlFromLink($referenceText, 'private');
+					$reference->setRichObject(
+						'calendar_widget',
+						[
+							'title' => 'calendar',
+							'isPublic' => false,
+							'url' => $url,
+						]
+					);
+					break;
+				default:
+					return null;
+			}
+
+		
 
 			return $reference;
 		}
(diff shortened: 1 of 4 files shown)


st3iny:
I pushed some hacky code to make it work. You should be able to continue from here.

We should probably improve my hacky solution a bit once the PR is ready.

st3iny:
### Fixed Problems

1. FullCalendar was rendered before its options were ready as they are loaded async inside `created()`.
2. `calendarOptions.eventSources` is not reactive and thus no events are shown. The actual calendar is loaded asynchronously but the event sources are not updated any more. -> **It might be worth to investigate why this works in the actual app. (Perhaps the calendars are provided via initial state and are available right away?)**

lib/Reference/ReferenceProvider.php
```diff
@@ -73,24 +73,57 @@ public function getIconUrl(): string {
 	public function matchReference(string $referenceText): bool {
 		$start = $this->urlGenerator->getAbsoluteURL('/apps/' . Application::APP_ID);
 		$startIndex = $this->urlGenerator->getAbsoluteURL('/index.php/apps/' . Application::APP_ID);
+		//public calendars  /\/p\/([a-zA-Z0-9]+)/
```
ChristophWurst, line 76: duplication of the regex one line later?

src/components/CalendarGrid.vue
```diff
@@ -154,6 +168,16 @@ export default {
 			}
 		},
 		eventSources() {
+			if (this.isWidget) {
+				const publicUrl = window.OC.webroot + `/remote.php/dav/public-calendars/${this.token}/`
```
ChristophWurst, line 172: I think by using `@nextcloud/router` we should be able to avoid accessing this global variable

src/components/CalendarGrid.vue
```diff
@@ -154,6 +168,16 @@ export default {
...
+				const url = window.OC.webroot + this.url
```
ChristophWurst, line 173: same here

src/components/CalendarGrid.vue
```diff
@@ -194,6 +218,22 @@ export default {
 			const calendarApi = this.$refs.fullCalendar.getApi()
 			calendarApi.gotoDate(getYYYYMMDDFromFirstdayParam(newDate))
 		},
+		eventSources(sources, oldSources) {
+			const newSources = sources.filter(x => !oldSources.map(x => x.id).includes(x.id))
```
ChristophWurst, line 222: x is used twice. this works but makes the code confusing. rename the vars

src/components/CalendarGrid.vue
```diff
@@ -194,6 +218,22 @@ export default {
...
+			const removedSources = oldSources.filter(x => !sources.map(x => x.id).includes(x.id))
```
ChristophWurst, line 223: same here

src/views/Calendar.vue
```diff
@@ -139,14 +142,26 @@ export default {
 		EditSimple,
 	},
 	props: {
+		// is the calendar in a widget
 		isWidget: {
 			type: Boolean,
 			default: false,
 		},
+		// The reference token for the widget for public share calendars
```
ChristophWurst, line 150: nitpick: comment starts with capital letter :cat:

src/views/Calendar.vue
```diff
@@ -139,14 +142,26 @@ export default {
 		EditSimple,
 	},
 	props: {
+		// is the calendar in a widget
```
ChristophWurst, line 145: here it starts with a lower case letter :crying_cat_face:

ChristophWurst (commented):
Code quality and OCD feedback

src/components/CalendarGrid.vue
hamza221: Actually window.OC.webroot  not needed at all, escaped my testing

lib/Reference/ReferenceProvider.php
hamza221: a bit confused here you mean `\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+\/$/i` and `\/p\/[a-zA-Z0-9]+$/i` ? 
If you mean the code it doesn't exist on the latest commit

hamza221: if you mean the comment , was a testing residual, removed

lib/Reference/ReferenceProvider.php
```diff
@@ -73,24 +73,55 @@ public function getIconUrl(): string {
 	public function matchReference(string $referenceText): bool {
 		$start = $this->urlGenerator->getAbsoluteURL('/apps/' . Application::APP_ID);
 		$startIndex = $this->urlGenerator->getAbsoluteURL('/index.php/apps/' . Application::APP_ID);
+		if(preg_match('/^' . preg_quote($start, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1 || preg_match('/^' . preg_quote($startIndex, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1) {
```
ChristophWurst, line 76:
```suggestion
		if (preg_match('/^' . preg_quote($start, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1 || preg_match('/^' . preg_quote($startIndex, '/') . '\/p\/[a-zA-Z0-9]+$/i', $referenceText) === 1) {
```

lib/Reference/ReferenceProvider.php
```diff
@@ -99,15 +130,31 @@ public function resolveReference(string $referenceText): ?IReference {
...
+		// \/dav\/calendars\/([a-zA-Z0-9-]+)\/([a-zA-Z0-9-]+)
+		if (preg_match('/\/remote.php\/dav\/calendars\/([a-zA-Z0-9-]+)\/([a-zA-Z0-9-]+)\//', $link, $output_array)) {
+			return $output_array[0];
+		}
+		return null;
 
 	}
 
+	private function getType(string $url): string {
+		if(preg_match('/\/p\/([a-zA-Z0-9]+)/', $url) === 1) {
```
ChristophWurst, line 149:
```suggestion
		if (preg_match('/\/p\/([a-zA-Z0-9]+)/', $url) === 1) {
```

ChristophWurst (approved):
We really need better cs-fixer rules for the space after if :laughing: 

Looks good :shipit:

st3iny (changes requested):
Breaks on instances that are hosted under a webroot.

src/components/CalendarGrid.vue
```diff
@@ -154,6 +168,16 @@ export default {
 			}
 		},
 		eventSources() {
+			if (this.isWidget) {
+				const publicUrl = `/remote.php/dav/public-calendars/${this.token}/`
+				const url = this.url
```
st3iny, line 173:
```suggestion
				const publicUrl = OC.webroot + `/remote.php/dav/public-calendars/${this.token}/`
				const url = OC.webroot + this.url
```

st3iny: I'm not sure where the url is generated as I'm not that familiar with the widget backend stuff.

I guess the url should be fixed in the backend where it is generated.

hamza221: url comes from the pasted text, in Talk , text ...

st3iny: I just read Christoph's comment and there is a cleaner variant available: https://nextcloud-libraries.github.io/nextcloud-router/functions/getRootUrl.html

So it should be `getRootUrl()` instead of `OC.webroot`.

st3iny: > url comes from the pasted text, in Talk , text ...

Yeah, I'm confused because I pasted the full url with `https://` and the webroot into Talk. Maybe this bug affects other widgets as well.

hamza221: Oh, yes what we get in the frontend is extracted using regex in `lib/Reference/ReferenceProvider.php` because calendar's in the store looks like this 
image

hamza221: can be added in the backend using `getWebroot` from `IURLGenerator` that way we don't event have to check for `isPublic`

st3iny (approved):
I tested a private share (internal link) and a public share in Talk and they both worked.

Great job fixing the link stuff!

st3iny:
CI fix is at https://github.com/nextcloud/calendar/pull/5928

Issue #5872, opened 2024-03-22
Attendee autosuggestions don't work for user emails


miaulalala:
### Steps to reproduce

1. Create a user jane with email j.doe@gardenia.com
2. Try looking for the user via email address in the Attendee tab
3. No user suggested unless the full email is entered, and even then it's not a match for the user object that is indeed returned from the API call, but simply text.

| Name search | Email search |
|--------|--------|
| Cell | Cell | 

Probably related to https://github.com/nextcloud-libraries/nextcloud-vue/issues/5293

There's also the weird artifact of using backspace and then for a second, the user autosuggestion shows:

![Peek 2024-03-22 13-13](https://github.com/nextcloud/calendar/assets/7427347/0f566b18-3d60-45b4-901b-306a58481c8b)

### Expected behavior

I find a user via name OR email address

### Actual behaviour

I can only find users via name

### Calendar app version

4.7.0-beta.4

### Server operating system

Ubunut 20

### Web server

Apache

### Database engine version

MySQL

### PHP engine version

PHP 8.2

### Updated from an older installed version or fresh install

None

PR #5893, opened 2024-04-02
fix(scheduling): Find attendee via email

ChristophWurst:
Fixes https://github.com/nextcloud/calendar/issues/5872

NcSelect only shows options that **match the current search input**. If you search for user *Alice* with an email of *accounting@acme.com* you will not see her in the suggestions. If her email is *alice@acme.com* you will.

This change adds the email to the property used for searching so NcSelect shows it again.


GVodyanov:
Issues found:
- Users with no set email are not found
- Users with spaces in their usernames don't have their profile pictures rendered

Will get to work on it :)

GVodyanov (approved):

miaulalala (approved):
Wörks

PR #5906, opened 2024-04-08
feat(simple-editor): show additional actions inline

szaimen:
| Before | After |
|---|---|
| image | image |

### DIFF
--- a/src/views/EditSimple.vue
+++ b/src/views/EditSimple.vue
@@ -55,7 +55,7 @@
 
 			<template v-else>
 				<div class="event-popover__top-right-actions">
-					<Actions v-if="!isLoading && !isError && !isNew" :force-menu="true">
+					<Actions v-if="!isLoading && !isError && !isNew" :inline="5">
 						<ActionLink v-if="!hideEventExport && hasDownloadURL"
 							:href="downloadURL">
 							<template #icon>


miaulalala (approved):
Nice!

jancborchardt (changes requested):
Some actions are more important or relevant than others here. Showing _all_ of them inline is a bit much and results in a lot of icons without text, which we want to avoid because it is not readily understandable for people who do not use the software every day. :)

- Download and Copy is not as important as "Delete" and "Edit"
- The actions belong more to "Edit" than they do to the close button at the top

That’s why I propose an adjustment like this:
calendar event

szaimen:
> That’s why I propose an adjustment like this:
> calendar event

Good idea to move them to the bottom! :)

SebastianKrupinski:
> Some actions are more important or relevant than others here. Showing _all_ of them inline is a bit much and results in a lot of icons without text, which we want to avoid because it is not readily understandable for people who do not use the software every day. :)
> 
>     * Download and Copy is not as important as "Delete" and "Edit"
> 
>     * The actions belong more to "Edit" than they do to the close button at the top
> 

@jancborchardt @nimishavijay do we still want to do this?

st3iny:
I like the idea of moving them to the bottom. However, the devil lies in the details :tm:.

For recurring events we have two delete buttons: this and this+all-future. So we need to render either two buttons, show a menu still or come up with a second modal to let the user decide.

SebastianKrupinski:
Closing this in favour of https://github.com/nextcloud/calendar/pull/7491

PR #5907, opened 2024-04-08
feat(sidebar): show actions inline

szaimen:
- [ ] needs https://github.com/nextcloud-libraries/nextcloud-vue/pull/5458

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/src/views/EditSidebar.vue
+++ b/src/views/EditSidebar.vue
@@ -26,7 +26,7 @@
 
 <template>
 	<NcAppSidebar :empty="isLoading || isError"
-		:force-menu="true"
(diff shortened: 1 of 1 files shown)


jancborchardt (changes requested):
Same comment as https://github.com/nextcloud/calendar/pull/5906#pullrequestreview-1986633149
→ I don’t think we should inline all these actions here, but rather move the menu (with Download and Copy) and the out-of-menu "Delete" next to the "Edit" button.

SebastianKrupinski:
Closing as this is no longer relevant

Issue #5198, opened 2023-05-04
Notification of mail to the owner of an appointment, shows the time in format of the time zone of the applicant instead of the time zone of the owner


nonfreegithub:
### Steps to reproduce

1. Config an appointment
2. Request an appointment with diferent time zone
3. Check the mail (to the owner)

### Expected behavior

Notification of mail to the owner of an appointment, shows the time in format of the owner's time zone

### Actual behaviour

Notification of mail to the owner of an appointment, shows the time in format of the applicant's time zone

### Calendar app version

4.3.3

### Browser

Firefox

### Client operating system

GNU/Linux

### Server operating system

GNU/Linux

### Web server

Apache

### Database engine version

PostgreSQL

### PHP engine version

PHP 7.4

### Nextcloud version

25.0.6

### Updated from an older installed version or fresh install

Fresh install

mukil:
I wonder if I have the same issue but to reproduce, can somebody tell me how to do  this?

> Request an appointment with diferent time zone

I would suspect the nextcloud webpage on which guests can choose a date and a timeslot in my calendar **does not** attempt to calculate some timezone client-side, or _does it_?

I am currently trying to figure out how a guest user - no authenticated HTTP session with my nextcloud - who booked an appointment in my calendar - via nextcloud calendar appointment feature - has a calendar entry which starts one hour later then the entry in my nextcloud calendar? Is it maybe the scope of this issue is not only about the "notification mail" but also affects the .ics/event data sent out by nextcloud. 

Sadly, I was unable to reproduce the issue? Therefore it would be helpful if somebody could tell me how to request an appointment with a different timezone.

SebastianKrupinski:
Confirmed.

Both the requestee and requester get a email with the exact same time, even tho the requestee's calendar time zone and requester's appointment time zone are different. Additionally the requestee gets a gui notification with the requester's time zone adjusted time and NOT their own calendar time zone.

SebastianKrupinski:
Issue addressed in PR #5924

PR #5924, opened 2024-04-16
fix: send requestee and requester correct date and time when timezone…

SebastianKrupinski:
Send requestee and requester correct date and time when requestee and requester time zones are different.

Fixes https://github.com/nextcloud/calendar/issues/5198


ChristophWurst (commented):
Code looks good! Could you please add a to `\OCA\Calendar\Tests\Unit\Service\Appointments\MailServiceTest` that covers this scenario? Look at `\OCA\Calendar\Tests\Unit\Service\Appointments\MailServiceTest::testSendConfirmationEmail`. Create a similar test that asserts the values passed to `formatDateTimeRelativeDay`

tests/php/unit/Service/Appointments/MailServiceTest.php
```diff
@@ -505,6 +519,104 @@ public function testSendOrganizerBookingInformationEmail(): void {
...
+			->method('createAttachment');
+		$this->l10n->expects(self::exactly(6))
+			->method('t');
+		$this->lFactory->expects(self::once())
+			->method('findGenericLanguage')
+			->willReturn('en');
+		$this->lFactory->expects(self::once())
+			->method('get');
+		$this->dateFormatter->expects(self::once())
+			->method('formatDateTimeRelativeDay')
```
ChristophWurst, line 581: by using `->with(args)` you can assert the value of the arguments passed to this method. It will allow you to test if the fourth argument matches the expected timezone value. Same for the new ``testSendOrganizerBookingInformationEmailDifferentTZ`` case

lib/Db/AppointmentConfig.php
```diff
@@ -189,6 +189,10 @@ public function setAvailabilityAsArray(array $availability): self {
 		return $this;
 	}
 
+	public function getAvailabilityAsArray(): array {
+		return json_decode($this->getAvailability(), true);
```
ChristophWurst, line 193: nitpick: use https://www.php.net/manual/en/json.constants.php#constant.json-throw-on-error so decoding errors do not go unnoticed.

SebastianKrupinski: Done

lib/Service/Appointments/MailService.php
```diff
@@ -333,11 +339,16 @@ public function sendOrganizerBookingInformationEmail(Booking $booking, Appointme
 	}
 
 	public function sendOrganizerBookingInformationNotification(Booking $booking, AppointmentConfig $config) {
+
+		$tzid = $config->getAvailabilityAsArray()['timezoneId']; // extract time zone from appointment configuration
```
ChristophWurst, line 343:
```suggestion
		$tzid = $config->getAvailabilityAsArray()['timezoneId']; // extract time zone from appointment configuration
```

tests/php/unit/Service/Appointments/MailServiceTest.php
```diff
@@ -551,6 +662,7 @@ public function testSendOrganizerBookingInformationEmail(): void {
 			->method('get');
 		$this->dateFormatter->expects(self::once())
 			->method('formatDateTimeRelativeDay')
+			->with(self::anything(), self::anything(), self::anything(), new \DateTimeZone('Europe/Berlin'))
```
ChristophWurst, line 665: :+1: nice

ChristophWurst (approved):
Code looks good!

ChristophWurst:
@SebastianKrupinski please squash your commits to a reasonable minimum number of commits before merge. It will help us maintain a clean git history, which makes the generation of a changelog a lot easier :pray:

Issue #3786, opened 2021-12-02
Allow shared calendars as appointment conflict calendars


ChristophWurst:
**Is your feature request related to a problem? Please describe.**

As a user I might have calendars shared to me by other users, e.g. a company-wide calendar.

**Describe the solution you'd like**

It would be nice if I could select that shared calendar for appointment conflicts as well. Then company-wide events would also block appointment slots.

**Describe alternatives you've considered**

N/a

**Additional context**

- [x] https://github.com/nextcloud/server/issues/30068

violoncelloCH:
in general, being able to chose which calendars are treated as conflict calendars for appointments would be useful!

ChristophWurst:
> in general, being able to chose which calendars are treated as conflict calendars for appointments would be useful!

Coming with v3.1 https://github.com/nextcloud/calendar/pull/3770

reos-rcrozier:
I was about to make a feature request similar to this, but am going to comment here instead. 

In our case most often Appointments are to be used for arranging meetings with an external party (who can't see shared calendars). For this reason it would be most useful to have an Appointments URL that can be a combination of the public Appointment calendars of several users. 

For example, if several people in the company are trying to arrange an appointment with some external party you would ideally send them one link which provides the appointments available that work for all three employees. This would be really handy.

This seems different from what you propose because one has to set up a new Appointments set for every combination of users to be present at a meeting.

Blechlawine:
would this also apply to subscription calendars as well? Or just calendars shared within nextcloud?

reos-rcrozier:
@Blechlawine if you're rpelying to me, I was only thinking of calendars shared withing Nextcloud, but actually I mean the _Appointments_ calendars shared within Nextcloud, 

My idea is users can set up their own appointments calendars, just as one can now, and I could somehow select from them to make the combined appointments calendar.

Blechlawine:
@reos-rcrozier I was replying to op, but what your're describing sounds very useful as well

ChristophWurst:
I'm sure there are cases where subscriptions as conflict source make sense :+1: 

I just don't know if the change is in scope for this ticket when someone picks it up. We can re-evaluate afterwards and create a another ticket if not.

rasos:
We have an organization wide calendar with team meetings which is shared by an organization-admin account. That should be definitely be taken into account when checking for conflicts to offer free time slots.

pjrobertson:
[...]

PR #5934, opened 2024-04-19
Allow any calendar the user has access to, to be selected for the 'conflict calendar'

pjrobertson:
This is an attempted fix for #3786

Any calendar should be selectable for the 'conflict calendars', even if the user doesn't have write access to it. (e.g. a company has a 'holidays' calendar that a user should be able to select for conflicts.


st3iny (commented):
I tested this and the frontend parts work. However, incoming shared calendars are not considered by the backend when calculating slots during booking.

The CalDAV backend inside server code has to be changed to fix this as far as I can see.

Conflict calendars are queried from the backend by their URIs but shared calendars are stored in another table `oc_dav_shares`. So they won't be listed and thus no slots will be blocked by events in incoming shared (read-only) calendars

https://github.com/nextcloud/server/blob/eeb6ddb176adc7488b74d004dceafee960603ebc/apps/dav/lib/CalDAV/CalDavBackend.php#L613-L617

miaulalala:
> I tested this and the frontend parts work. However, incoming shared calendars are not considered by the backend when calculating slots during booking.
> 
> The CalDAV backend inside server code has to be changed to fix this as far as I can see.
> 
> Conflict calendars are queried from the backend by their URIs but shared calendars are stored in another table `oc_dav_shares`. So they won't be listed and thus no slots will be blocked by events in incoming shared (read-only) calendars
> 
> [nextcloud/server@`eeb6ddb`/apps/dav/lib/CalDAV/CalDavBackend.php#L613-L617](https://github.com/nextcloud/server/blob/eeb6ddb176adc7488b74d004dceafee960603ebc/apps/dav/lib/CalDAV/CalDavBackend.php#L613-L617)

Are you sure? https://github.com/nextcloud/server/blob/9d4b9440986afddc28dad99c7987f9c31513ae9f/apps/dav/lib/CalDAV/CalendarProvider.php#L37 should already search shared calendars

SebastianKrupinski:
Closing this as this is no longer relavent

Issue #3040, opened 2021-04-25
VALARM calendar events missing required DESCRIPTION when using Notifications reminders


ressit:
## Steps to reproduce
- Create a calendar event in nexctloud web interface
- add an alarm 10 min before
- Sync with DAVx5 to android client

## Expected behaviour
The event is synced
## Actual behaviour
- DAVx5 raises a nullptr exception at android client
- validating the calendar entry created by the calendar app with online validator https://icalendar.org/validator.html shows RFC error

Errors
    Missing DESCRIPTION property in VALARM near line # 14
    Reference: [RFC 5545 3.6.6. Alarm Component](http://icalendar.org/iCalendar-RFC-5545/3-6-6-alarm-component.html)
## Created calendar entry
```
BEGIN:VCALENDAR
PRODID:-//IDN nextcloud.com//Calendar app 2.0.5//EN
CALSCALE:GREGORIAN
VERSION:2.0
BEGIN:VEVENT
CREATED:20210425T140900Z
DTSTAMP:20210425T141004Z
LAST-MODIFIED:20210425T141004Z
SEQUENCE:2
UID:6a6537a3-d5b8-42e9-9521-3d96002c4cb9
DTSTART;TZID=Europe/Berlin:20210425T163000
DTEND;TZID=Europe/Berlin:20210425T170000
SUMMARY:test
BEGIN:VALARM
ACTION:DISPLAY
TRIGGER;RELATED=START:-PT10M
END:VALARM
END:VEVENT
BEGIN:VTIMEZONE
TZID:Europe/Berlin
BEGIN:DAYLIGHT
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
TZNAME:CEST
DTSTART:19700329T020000
RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU
END:DAYLIGHT
BEGIN:STANDARD
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
TZNAME:CET
DTSTART:19701025T030000
RRULE:FREQ=YEARLY;BYMONTH=10;BYDAY=-1SU
END:STANDARD
END:VTIMEZONE
END:VCALENDAR
```
## Versions
nextcloud 19.0.10.1
calendar app v 2.0.5
Firefox 88.0

DAVx5 version v3.3.10-gplay

joshtrichards:
[...]

PR #6008, opened 2024-05-15
fix(alarms): also update DISPLAY alarms

miaulalala:
Fixes https://github.com/nextcloud/calendar/issues/3040


st3iny (approved):
Tested and works.

Issue #6011, opened 2024-05-15
Add all holiday calendars from Thunderbird?


nclm:
**Is your feature request related to a problem? Please describe.**

I like the fact that we can add holiday calendars straight from Nextcloud, provided by Thunderbird.

Those are the choices in Nextcloud right now (about 56):
https://github.com/nextcloud/server/assets/923008/2a6713da-1ac4-40f5-8054-d1808fcbeecd

However Thunderbird.net offers more calendars that that (about 86):
https://github.com/nextcloud/server/assets/923008/64d172ec-228c-457e-ba67-9bd93e16a28d
as can been seen on https://www.thunderbird.net/en-US/calendar/holidays/

**Describe the solution you'd like**
Include all calendars offered by Thunderbird in Nextcloud.
Additionally, if there could be a way to automatically update the list in Nextcloud if Thunderbird adds or change their calendar, that would be useful to avoid differences like that in the future.

joshtrichards:
The list is currently manually maintained. It's located here:

https://github.com/nextcloud/calendar/blob/b2d23f855269ab34494701a9df1bcdbd9792a904/src/resources/holiday_calendars.json

It was just updated a few months ago. I'm not sure if some were left out at the time or if new ones have since been added @ Thunderbird.

Feel free to make some additions to it and submit them via a PR. :-)

nclm:
Alright, it seems that https://github.com/nextcloud/calendar/blob/main/src/resources/holiday_calendars.json is simply a copy of https://github.com/thunderbird/thunderbird-website/blob/master/media/caldata/calendars.json, which makes sense. 

But Thunderbird is now using https://github.com/thunderbird/thunderbird-website/blob/master/media/caldata/autogen/calendars.json since last year, with (more) calendars auto-updated every quarter from a “calendar provider”. If you are up for using these, replacing `holiday_calendars.json` with the contents of `autogen/calendars.json` could do the job (as long as Nextcloud can handle the extra fields (`locale`, `language`, `updated`).

miaulalala:
[...]

PR #6014, opened 2024-05-16
Update holiday_calendars.json to be in sync with Thunderbird

nclm:
From conversation in #6011:

Imported content from https://github.com/thunderbird/thunderbird-website/blob/master/media/caldata/autogen/calendars.json (version from Apr 1, 2024, 9:39 PM GMT+2) bringing the number of calendars from about 56 to about 86.

Adds 'locale', 'language', and 'updated' fields that are not used by Nextcloud at the moment but present in Thunderbird’s source file.

### DIFF
--- a/src/resources/holiday_calendars.json
+++ b/src/resources/holiday_calendars.json
@@ -1,336 +1,776 @@
 [
-	{
-		"country":"Algeria",
-		"filename":"AlgeriaHolidays.ics",
-		"datespan":"2024-2027",
-		"authors":"Imad Tbahriti"
-	},
-	{
-		"country":"Australia",
-		"filename":"AustraliaHolidays.ics",
-		"datespan":"2024-2027",
-		"authors":"RGreyman"
-	},
-	{
-		"country":"Austria",
-		"filename":"AustrianHolidays.ics",
-		"datespan":"2024-2027",
-		"authors":"boe"
-	},
-	{
-		"country":"Belgium",
-		"filename":"BelgianHolidays.ics",
-		"datespan":"2024-2027",
-		"authors":"Hubertus Verdonck"
-	},
-	{
-		"country":"Bolivia",
-		"filename":"BoliviaHolidays.ics",
-		"datespan":"2024-2027",
-		"authors":"Rebelde Boliche"
-	},
-	{
-		"country":"Brazil",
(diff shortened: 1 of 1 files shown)


miaulalala:
The problem is see is that old subscriptions aren't updated to the new endpoint, which means even if I have a subscription, the button in the subscription calendar list isn't greyed out. Instead I can subscribe to the same holiday calendar twice - in my test case for Germany. That's a bit confusing for the users.

Looks like the subscription is found in \Sabre\CalDAV\Backend\SubscriptionSupport::getSubscriptionsForUser.

I think this will need a bit of a code change in the backend, maybe a migration step that updates the subscriptions to use the new `autogen` endpoint automatically.

p4p4j0hn:
> maybe a migration step that updates the subscriptions to use the new `autogen` endpoint automatically.

A more immediate change that is necessary is the initial subscription url for the holiday calendars should have the `autogen` endpoint. The Canadian calendar being pulled in does not work for me, since Nextcloud Calendar is pointing to an outdated endpoint. If I subscribe to the updated calendar directly from Thunderbird all the holidays display correctly.

cwerstuik:
@miaulalala If it's as easy as deleting the old calendar and adding it again, maybe a migration step can be added later...

I also was hoping to rely on the Canadian calendar, but the holidays are all wrong so as it is, no one in Canada is going to be using this calendar. Thanks for considering.

miaulalala (approved):

nclm:
Great thanks, 

Would there be any way to quarterly/biyearly/yearly update the file with the latest from Thunderbird?
They seem to update it “quarterly”: https://github.com/thunderbird/thunderbird-website/commits/master/media/caldata/autogen/calendars.json

miaulalala:
> Great thanks,
> 
> Would there be any way to quarterly/biyearly/yearly update the file with the latest from Thunderbird? They seem to update it “quarterly”: [thunderbird/thunderbird-website@`master`/media/caldata/autogen/calendars.json (commits)](https://github.com/thunderbird/thunderbird-website/commits/master/media/caldata/autogen/calendars.json)

not automatically yet - maybe there's some script we could run but I'm not sure.

nclm:
Or maybe some kind of yearly repeating task in Github or whichever project management tool, just to remember to do a manual PR regularly.

PR #6058, opened 2024-06-13
fix: composer out of sync files

miaulalala:
background: https://github.com/nextcloud/calendar/pull/5607 is incomplete. composer.json was changed without composer.lock, so the two files got out of sync


ChristophWurst (approved):

ChristophWurst:
FYI it's not just CI. Composer 2.7.7 fails locally too

PR #6129, opened 2024-07-08
Improve sidebar organization

GVodyanov:
Close https://github.com/nextcloud/calendar/issues/5083

| A | B |
| - | - |
| image | image |

### DIFF
--- a/src/components/AppNavigation/CalendarList.vue
+++ b/src/components/AppNavigation/CalendarList.vue
@@ -9,17 +9,66 @@
 		v-bind="{swapThreshold: 0.30, delay: 500, delayOnTouchOnly: true, touchStartThreshold: 3}"
 		draggable=".draggable-calendar-list-item"
 		@update="update">
+		<CalendarListNew />
 		<template v-if="!isPublic">
-			<CalendarListItem v-for="calendar in calendars"
+			<CalendarListItem v-for="calendar in sortedCalendars.personal"
 				:key="calendar.id"
 				class="draggable-calendar-list-item"
 				:calendar="calendar" />
 		</template>
 		<template v-else>
-			<PublicCalendarListItem v-for="calendar in calendars"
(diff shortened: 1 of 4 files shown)


src/components/AppNavigation/CalendarList.vue
```diff
@@ -99,3 +196,20 @@ export default {
 	},
 }
 </script>
+
+<style scoped>
+:deep(.list-item) {
+	padding: 4px 8px 4px 8px;
+	margin-bottom: 0 !important;
```
ChristophWurst, line 203: deep and !important are often a code smell. would it possible to achieve this without? e.g. by moving the styles into the child or upstreaming conditional styles via a prop in nc/vue?

nimishavijay (approved):
Awesome! Looks super nice :) 

Only one minor point of feedback: since there is now a heading for the deck section, would it be possible to remove the repeated "Deck: " in the beginning of those calendars? 

Can't tell from the screenshot, but if a section doesn't have any calendars it should be hidden. 

Approving to not block, they can also be follow ups 🚀

Ornanovitch:
If I may, I love this improvement but the "hidden" section seems too much: a disabled calendar might be a Deck one, a Shared one... Disabled is more a state than a category, IMO a disabled calendar should be grayed out in its proper section. :)

GVodyanov:
Hey everyone!

Screenshot from 2024-07-12 12-22-39

Here is what it looks like now: no custom CSS, sections are hidden if empty, "Deck: " removed, and the hidden menu is implemented with `allowCollapse` of `NcAppNavigationItem`.

@nimishavijay @ChristophWurst

ChristophWurst:
The headings and the checkboxes are marginally misaligned. Perhaps this could be addressed in nc/vue? I assume other apps using the same components could have similar alignment issues

ChristophWurst (approved):
Please squash the three commits into one for a clean history and atomic changes :)

Issue #3193, opened 2021-06-11
Option to set a calendar as transparent for free/busy


ChristophWurst:
**Is your feature request related to a problem? Please describe.**

If you create a calendar it's currently always considered for free/busy scheduling. There are cases where you don't want this, e.g. for a calendar that is supposed to be used for a group, then the owner (the person who shared) shouldn't be seen as blocked for the shared events. But also sometimes you might create a calendar for something that isn't an actual event where you are busy, and you don't want to set your availability to free for every event you create in there.

**Describe the solution you'd like**

Offer users a checkbox on the calendar action menu where they can toggle the free/busy transparency.

**Describe alternatives you've considered**

N/a

**Additional context**

* The standardized property https://icalendar.org/CalDAV-Scheduling-RFC-6638/9-1-caldav-schedule-calendar-transp-property.html
* Handling in Sabre https://github.com/sabre-io/dav/blob/c1afdc77a95efea6ee40c03c45f57c3c0c80ec22/lib/CalDAV/Schedule/Plugin.php#L890-L894

naelfe:
Or if you have a separate calendar for your cities garbage disposal, where you simply import the ical for the whole year but are marked busy for every other day.
Would be indeed handy to set the hole calendar as transparent. It's too much extra work for me to change the ical every time manually, because they tend to change it a couple of times during the year :D

miaulalala:
Seems group / shared calendars are always treated as transparent as per:

https://github.com/nextcloud/server/blob/ad78f7e48e7f03fc6c3f47a8972d122cde275d97/apps/dav/lib/CalDAV/CalDavBackend.php#L416

so any calendar that is shared should not have this checkbox.

ChristophWurst:
I suggest to show the checkbox read-only and unchecked, or some text. Then this will also be easy to understand for users.

miaulalala:
> I suggest to show the checkbox read-only and unchecked, or some text. Then this will also be easy to understand for users.

You mean for shared and subscription calendars?

PR #6169, opened 2024-07-15
feat: allow calendar-wide transparency settings

miaulalala:
Fixes https://github.com/nextcloud/calendar/issues/3193

~Needs https://github.com/nextcloud/cdav-library/pull/908~
~Needs https://github.com/nextcloud/cdav-library/pull/909~

How to test:

- Uncheck the checkbox in the Edit Calendar Modal
- Check oc_calendars row for the calendar - the value for the transparency should change to 1 for transparent when the checkbox is unchecked.

### DIFF
--- a/src/components/AppNavigation/EditCalendarModal.vue
+++ b/src/components/AppNavigation/EditCalendarModal.vue
@@ -24,7 +24,11 @@
 					:placeholder="$t('calendar', 'Calendar name …')"
 					@input="calendarNameChanged = true">
 			</div>
-
+			<template v-if="canBeShared">
+				<NcCheckboxRadioSwitch :checked.sync="isTransparent">
+					{{ $t('calendar', 'Never show me as busy (set this calendar to transparent)') }}
+				</NcCheckboxRadioSwitch>
+			</template>
 			<template v-if="canBeShared">
 				<h2 class="edit-calendar-modal__sharing-header">
 					{{ $t('calendar', 'Share calendar') }}
@@ -72,7 +76,7 @@
 </template>
 
 <script>
-import { NcModal, NcColorPicker, NcButton, NcAppNavigationSpacer } from '@nextcloud/vue'
+import { NcModal, NcColorPicker, NcButton, NcCheckboxRadioSwitch, NcAppNavigationSpacer } from '@nextcloud/vue'
 import PublishCalendar from './EditCalendarModal/PublishCalendar.vue'
 import SharingSearch from './EditCalendarModal/SharingSearch.vue'
 import ShareItem from './EditCalendarModal/ShareItem.vue'
@@ -101,11 +105,13 @@ export default {
 		CheckIcon,
 		InternalLink,
 		NcAppNavigationSpacer,
+		NcCheckboxRadioSwitch,
 	},
 	data() {
 		return {
 			calendarColor: undefined,
 			calendarColorChanged: false,
+			isTransparent: false,
 			calendarName: undefined,
 			calendarNameChanged: false,
 		}
@@ -164,6 +170,7 @@ export default {
 			this.calendarColor = calendar.color
 			this.calendarNameChanged = false
 			this.calendarColorChanged = false
+			this.isTransparent = calendar.transparency === 'transparent'
 		},
 	},
 	methods: {
@@ -187,6 +194,24 @@ export default {
 				logger.error('Failed to save calendar color', {
 					calendar: this.calendar,
 					newColor: this.calendarColor,
+				})
+				throw error
+			}
+		},
+
+		/**
+		 * Save the calendar transparency.
+		 */
+		async saveTransparency() {
+			try {
+				await this.calendarsStore.changeCalendarTransparency({
+					calendar: this.calendar,
+					transparency: this.isTransparent ? 'transparent' : 'opaque',
+				})
+			} catch (error) {
+				logger.error('Failed to save calendar transparency', {
+					calendar: this.calendar,
+					transparency: this.isTransparent ? 'transparent' : 'opaque',
 				})
 				throw error
 			}
@@ -220,6 +245,7 @@ export default {
 				if (this.calendarColorChanged) {
 					await this.saveColor()
 				}
+				await this.saveTransparency()
 				if (this.calendarNameChanged) {
 					await this.saveName()
 				}
@@ -243,7 +269,7 @@ export default {
 }
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
 .edit-calendar-modal {
 	padding: 20px;
 	display: flex;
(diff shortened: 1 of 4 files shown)


st3iny:
https://www.npmjs.com/package/@nextcloud/cdav-library/v/1.5.0

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -24,7 +24,19 @@
 					:placeholder="$t('calendar', 'Calendar name …')"
 					@input="calendarNameChanged = true">
 			</div>
-
+			<template v-if="canBeShared">
+				<NcListItem :name="$t('calendar', 'Include calendar in free/busy checks')"
```
miaulalala, line 28:
```suggestion
				<NcListItem :name="$t('calendar', 'All events in this calendar show you as busy by default')"
```

Also document that a transparent calendars VEVENTS will never show you as busy even if the event is set to busy.

st3iny:
The cdav library was released again: https://www.npmjs.com/package/@nextcloud/cdav-library/v/1.5.1

It looks like it was already updated on main. So another rebase will pull in the dependency.

Ref https://github.com/nextcloud/calendar/pull/6182

hamza221:
There's enough space to show the whole text, but it should be fixed uptream
image

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -24,7 +24,19 @@
 					:placeholder="$t('calendar', 'Calendar name …')"
 					@input="calendarNameChanged = true">
 			</div>
-
+			<template v-if="canBeShared">
+				<NcListItem :name="$t('calendar', 'All events in this calendar show you as busy by default')"
+					@click="isTransparent = !isTransparent">
+					<template #icon>
```
hamza221, line 30: [NcCheckboxRadioSwitch](https://nextcloud-vue-components.netlify.app/#/Components/NcCheckboxRadioSwitch) can be used here instead

hamza221 (approved):

hamza221:
Works otherwise ✅

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -24,7 +24,19 @@
 					:placeholder="$t('calendar', 'Calendar name …')"
 					@input="calendarNameChanged = true">
 			</div>
-
+			<template v-if="canBeShared">
+				<NcListItem :name="$t('calendar', 'All events in this calendar show you as busy by default')"
```
susnux, line 28: This is not within a list and thus will be invalid HTML. Only use `NcListItem` within `NcAppNavigationList` or a plain `ul` or `ol` tag.

This also is accessible wise really bad, besides the list item being not in a list you are using a link as checkbox without setting the role correctly.
Instead I would recommend just using `NcCheckboxRadioSwitch` with checkbox type and overwrite the color of the checkbox (we could make this also an option in the library but also works with styles here).

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -24,7 +24,11 @@
 					:placeholder="$t('calendar', 'Calendar name …')"
 					@input="calendarNameChanged = true">
 			</div>
-
+			<template v-if="canBeShared">
+				<NcCheckboxRadioSwitch :checked.sync="isTransparent">
+					{{ $t('calendar', 'All events in this calendar show you as busy by default') }}
```
ChristophWurst, line 29: "Events in this calendar show you as busy if the events are marked busy."

I thinkt the current wording is misleading. This is not about defaults. It's about whether the calendar is checked for business at all. Events with status free still show you as free.

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -164,6 +170,7 @@ export default {
 			this.calendarColor = calendar.color
 			this.calendarNameChanged = false
 			this.calendarColorChanged = false
+			this.isTransparent = calendar.transparency === 'transparent'
```
ChristophWurst, line 173:
```suggestion
			this.isTransparent = calendar.transparency !== 'opaque'
```

Then we assume the calendar to be transparent when calendar.transparency is undefined.

From the code comment below

> If this property is not present on a calendar collection,
> then the default value CALDAV:opaque MUST be assumed.

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -220,6 +245,7 @@ export default {
 				if (this.calendarColorChanged) {
 					await this.saveColor()
 				}
+				await this.saveTransparency()
```
ChristophWurst, line 248: nit: only save if it changed?

src/models/calendar.js
```diff
@@ -54,6 +54,8 @@ const getDefaultCalendarObject = (props = {}) => Object.assign({}, {
 	calendarObjects: [],
 	// Time-ranges that have already been fetched for this calendar
 	fetchedTimeRanges: [],
+	// Scheduling transparency
+	transparency: '',
```
ChristophWurst, line 58:
```suggestion
	transparency: 'opaque',
```

for the correct default?

ChristophWurst (commented):
Looks good!

src/components/AppNavigation/EditCalendarModal.vue
GVodyanov: We have that here: https://github.com/nextcloud/calendar/blob/e98cd45aa9ae45477de8a2ccefb4c7a3fb43ed71/src/store/calendars.js#L549

I thought it would make more sense for the check to happen in the store

ChristophWurst: That works too!

src/components/AppNavigation/EditCalendarModal.vue
miaulalala: Or we could switch the logic (using `!isTransparent`) and use "Never show me as busy" since for "opaque" you can still set the VEVENT as FREE but for transparent, a BUSY will never be used.

ChristophWurst: That works, too :+1:

st3iny, line 29: There is a logic error here: The state is inverted.

Checking means that we want an opaque calendar but will cause a transparent calendar and vice versa.

Either invert the wording or rename the variable to `isOpaque`.

st3iny (changes requested):
Works otherwise

src/components/AppNavigation/EditCalendarModal.vue
miaulalala: Is there? `this.isTransparent = calendar.transparency !== 'opaque'`

st3iny (approved):
Tested and works now. The wording is now consistent with the actual state.

PR #6172, opened 2024-07-16
Make the new calendar caption more coherent with the appointments caption

GVodyanov:
| A | B |
| - | - |
| Screenshot from 2024-07-16 14-37-59 | Screenshot from 2024-07-16 14-37-24 |

(Also adds hover text)


miaulalala (approved):
Can't see a difference but sure :wink:

PR #6175, opened 2024-07-17
Fix some issues in free/busy suggestions

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6162

Improve the slot suggestions and make the algorithm more understandable

### DIFF
--- a/src/components/Editor/FreeBusy/FreeBusy.vue
+++ b/src/components/Editor/FreeBusy/FreeBusy.vue
@@ -456,13 +456,13 @@ export default {
 			try {
 				// for now search slots only in the first week days
 				const endSearchDate = new Date(startSearch)
-				endSearchDate.setDate(startSearch.getDate() + 7)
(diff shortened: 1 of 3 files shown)


GVodyanov:
For people reviewing, this is how the suggested slots should work:
- If the time you set for your event is free, it will be suggested, without being rounded (e.g. I set the time to 11:24-12:56, there are no blocking events in that period, that period will be suggested)

- Aside from suggesting the current time if free, it will look at all the events found by the free busy service, and see if everyone is free immediately after. So say two people are unavailable until 08:00 and 09:00, immediately after 08:00 is invalid, but immediately after 09:00 is.

- The time after the events will also be rounded to the nearest quarter-hour (so :00, :15, :30, :45), but always rounding up. There is always at least a minute of break between events, so say you have a blocking event ending at 13:00, the next event will be suggested at 13:15. If that blocking event were to end at 13:14, the next event would still be suggested at 13:15.

miaulalala (approved):

PR #6200, opened 2024-07-24
PropertyColor: now color is selected with submit button instead of directly

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6145

Based on @st3iny's idea https://github.com/nextcloud/calendar/issues/6145#issuecomment-2247834790

### DIFF
--- a/src/components/Editor/Properties/PropertyColor.vue
+++ b/src/components/Editor/Properties/PropertyColor.vue
@@ -21,9 +21,9 @@
 		<div v-else
 			class="property-color__input">
 			<ColorPicker :value="selectedColor"
-				:open.sync="isColorPickerOpen"
+				:shown.sync="isColorPickerOpen"
 				:advanced-fields="true"
-				@input="changeColor">
+				@submit="changeColor">
 				<NcButton class="property-color__color-preview"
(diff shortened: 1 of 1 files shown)


src/components/Editor/Properties/PropertyColor.vue
```diff
@@ -104,25 +103,12 @@ export default {
 	methods: {
 		/**
 		 * Changes / Sets the custom color of this event
-		 *
-		 * The problem we are facing here is that the
-		 * color-picker component uses normal hex colors,
-		 * but the RFC 7986 property COLOR requires
-		 * css-color-names.
-		 *
-		 * The color-space of css-color-names is smaller
```
ChristophWurst, line 113: We don't need debounce because submit happens only once?

ChristophWurst (approved):
Makes sense

src/components/Editor/Properties/PropertyColor.vue
GVodyanov: Yep and it doesn't happen immediately after choosing the color

ChristophWurst: Nice!

PR #6202, opened 2024-07-24
fix: enable attendee selection on shared calendars

SebastianKrupinski:
enable attendee selection on shared calendars


miaulalala (dismissed):
NVM, wrong ics

miaulalala (changes requested):
Popover modal is not in sync with sidebar modal:

Popover:
image

Popover Editor:

image



EditorSidebar:

image

SebastianKrupinski:
@miaulalala the popup modal is NOT out of sync, it just does not have enough room to show all the attendees, it has the same behavior with a personal calendar. This has nothing to do with this PR and would need to be addresses in another PR.

miaulalala:
> @miaulalala the popup modal is NOT out of sync, it just does not have enough room to show all the attendees, it has the same behavior with a personal calendar. This has nothing to do with this PR and would need to be addresses in another PR.

fair enough, can you please open a ticket with these findings?

ChristophWurst:
> I don't know what's the plan for [nextcloud/server#45054](https://github.com/nextcloud/server/pull/45054) (backports ?), but in any case, this should depend on the server version to be activated.

Calendar main is 30+ and we won't backport this to my knowledge. This should be alright :)

SebastianKrupinski:
> lice shares with her assistant

Correct.

SebastianKrupinski:
Additional thoughts...

Some front end UI changes that would be beneficial...

    Drop down on attendee tab to selected the organizer of the event. Use cases...

    Sharee is an assistant with access to the Managers calendar (Sharer)... The organizer of the event should be the Sharer.
    Sharee is a employee with access to the Company calendar (Sharer)... The organizer of the event should be the Sharee.

Some back end changes that would be beneficial...

    Additional calendar property to control access to send on behalf of the Sharer.

Draw backs...

    The above features would only work in the NC UI. CalDAV dose support the "CALDAV:schedule-send" ACL permission but I doubt that there are many if any clients that actually support this feature.

SebastianKrupinski:
> An organizer selection is still missing for the assistant use case, right? As in, Alice shares with her assistant Bob, who invites Charly. Alice is saved as organizer, bob is an attendee.

Correct. See comments above.

PR #6216, opened 2024-08-01
fix: add border to sidebar button

miaulalala:
Give a border to the sidebar button

| Before | After |
|--------|--------|
| image | image |
| image | image | 

cc @nextcloud/designers


GVodyanov (approved):
Now I know this would usually go into Nextcloud vue but calendar has a pretty irregular UI so I think having this in custom styling only here makes sense

st3iny (approved):
Makes sense in the context of Calendar. I'm no designer though.

PR #6229, opened 2024-08-05
fix: recurrance selection

### LINKED ISSUE #6228: Recurrence options no longer work
### Steps to reproduce

1. Create a new event or open an existing one
2. Try to add a recurrence option (like repeat daily)

### Expected behavior

I can select the recurrence interval, then the details

### Actual behaviour

image

Visually broken NcSelect, nothing happens on selection.

### Calendar app version

5.0

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

_No response_

None

### List of activated apps

_No response_

### Nextcloud configuration

_No response_

### Web server error log

_No response_

### Log file

_No response_

### Browser log

_No response_

### Additional info

_No response_

SebastianKrupinski:
Recurrence selection was not working, now it works. 

Fixes #6228

### DIFF
--- a/src/components/Editor/Repeat/Repeat.vue
+++ b/src/components/Editor/Repeat/Repeat.vue
@@ -224,7 +224,8 @@ export default {
 		 * @param {string} frequency Allowed values: NONE, DAILY, WEEKLY, MONTHLY, YEARLY
 		 */
 		changeFrequency(frequency) {
-			this.calendarObjectInstanceStore.changeRecurrenceFrequencyMutation({
+			this.calendarObjectInstanceStore.changeRecurrenceFrequency({
+				calendarObjectInstance: this.calendarObjectInstance,
 				recurrenceRule: this.recurrenceRule,
 				frequency,
 			})


ChristophWurst (approved):
Works! Thanks!

This seems to come from https://github.com/nextcloud/calendar/commit/88f6dd2bb294e6610550b832bb69e70920872658. @GVodyanov are there any similar method calls we should check?

SebastianKrupinski:
> Works! Thanks!
> 
> This seems to come from [88f6dd2](https://github.com/nextcloud/calendar/commit/88f6dd2bb294e6610550b832bb69e70920872658). @GVodyanov are there any similar method calls we should check?

I will test all the calendar functionality this week and see if anything else is broken.

GVodyanov:
> Works! Thanks!
> 
> This seems to come from [88f6dd2](https://github.com/nextcloud/calendar/commit/88f6dd2bb294e6610550b832bb69e70920872658). @GVodyanov are there any similar method calls we should check?

Yeah this was my bad, I just looked through the other usages and everything else should be fine though

ChristophWurst:
Happens to the best. Thanks for checking!

PR #6231, opened 2024-08-05
fix: recurrance until selection

SebastianKrupinski:
Recurrence until selection was broken. Now its fixed.


st3iny (approved):
Tested and works. I see you are slowly becoming a frontender :smile:

Issue #6185, opened 2024-07-22
Default reminder at event start labeled as "a few seconds"


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Open the settings menu at the bottom left
3. Open the default reminders select menu

### Expected behavior

Something like 
![image](https://github.com/user-attachments/assets/89f5b5ad-c665-4682-a196-4abd15a668e3)

### Actual behaviour

![Bildschirmfoto vom 2024-07-22 08-54-06](https://github.com/user-attachments/assets/bb693e83-7e8b-4914-a394-e35b783bdbc6)

### Calendar app version

4.8

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

abdou6666:
Hello, I am new to OSS and interested in contributing. Is someone working on this issue?

ChristophWurst:
Nobody is working on this. Feel free to experiment with it. 

https://nextcloud.com/developer/ for assistance.

GretaD:
closed here: https://github.com/nextcloud/calendar/pull/6249/files

PR #6249, opened 2024-08-09
fix: default reminder ui bug

abdou6666:
#6185 

before : 
image

after : 
image

### DIFF
--- a/src/components/AppNavigation/Settings.vue
+++ b/src/components/AppNavigation/Settings.vue
@@ -89,8 +89,7 @@
 				</template>
 				{{ $t('calendar', 'Copy iOS/macOS CalDAV address') }}
 			</ActionButton>
-			<ActionLink :href="availabilitySettingsUrl"
-				target="_blank">
+			<ActionLink :href="availabilitySettingsUrl" target="_blank">
 				<template #icon>
 					<OpenInNewIcon :size="20" decorative />
 				</template>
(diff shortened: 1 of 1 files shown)


hamza221:
The text is too long, it will never be displayed fully.  @ChristophWurst would only changing the default reminder be enough to close the issue ?

src/components/AppNavigation/Settings.vue
```diff
@@ -256,8 +255,9 @@ export default {
 		},
 		defaultReminderOptions() {
 			const defaultAlarms = getDefaultAlarms().map(seconds => {
+				const label = seconds === 0 ? t('calendar', "at event's start") : t('calendar', "{name} before the event's starts", { name: moment.duration(Math.abs(seconds) * 1000).locale(this.locale).humanize() })
```
miaulalala, line 258: Too much text for the width of the field unfortunately.

```suggestion
				const label = seconds === 0 ? t('calendar', "At event start") : moment.duration(Math.abs(seconds) * 1000).locale(this.locale).humanize()
```

miaulalala (changes requested):

src/components/AppNavigation/Settings.vue
```diff
@@ -256,8 +255,9 @@ export default {
 		},
 		defaultReminderOptions() {
 			const defaultAlarms = getDefaultAlarms().map(seconds => {
+				const label = seconds === 0 ? t('calendar', "At event start") : moment.duration(Math.abs(seconds) * 1000).locale(this.locale).humanize()
```
miaulalala, line 258:
```suggestion
				const label = seconds === 0 ? t('calendar', 'At event start') : moment.duration(Math.abs(seconds) * 1000).locale(this.locale).humanize()
```

CI doesn't like that :see_no_evil: 

If you could change that then we can merge - thanks so much!

miaulalala (changes requested):

hamza221 (approved):

miaulalala (approved):

miaulalala:
No backport due to translations

Issue #6226, opened 2024-08-05
Event bar is too short and has visible background "shadow"


miaulalala:
### Steps to reproduce

Create a new event in weekly view:

![image](https://github.com/user-attachments/assets/a5b2cb9c-230e-49cd-80cb-c62911162898)

### Expected behavior

The bar is wide enough to fit inside the calendar grid

### Actual behaviour

The bar is too short and exposes a background

### Calendar app version

5.0.0-beta.3

PR #6269, opened 2024-08-19
FullCalendar CSS: make events in weekly view have the correct width

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6226

The solution is a bit hacky, but I can't think of anything better...

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -235,6 +235,11 @@
 	}
 }
 
+.fc-timegrid-event-harness-inset {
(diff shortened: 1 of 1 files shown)


miaulalala (commented):
Did you test this on mobile?

GVodyanov:
> Did you test this on mobile?

Yeah, It's not always pixel perfect, but it works

miaulalala (approved):

PR #6290, opened 2024-08-26
fix(teams): resolve undefined variable error and add logging

miaulalala:
Signed-off-by: Anna Larch <anna@nextcloud.com>

### DIFF
--- a/lib/Controller/ContactController.php
+++ b/lib/Controller/ContactController.php
@@ -7,7 +7,9 @@ declare(strict_types=1);
  */
 namespace OCA\Calendar\Controller;
 
+use Exception;
 use OCA\Calendar\Service\ServiceException;
+use OCA\Circles\Api\v1\Circles;
 use OCA\Circles\Exceptions\CircleNotFoundException;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
@@ -17,6 +19,7 @@ use OCP\AppFramework\QueryException;
 use OCP\Contacts\IManager;
 use OCP\IRequest;
 use OCP\IUserManager;
+use Psr\Log\LoggerInterface;
 
 /**
  * Class ContactController
@@ -40,11 +43,14 @@ class ContactController extends Controller {
 	 * @param IRequest $request
 	 * @param IManager $contacts
 	 */
-	public function __construct(string $appName,
+	public function __construct(
+		string $appName,
 		IRequest $request,
 		IManager $contacts,
 		IAppManager $appManager,
-		IUserManager $userManager) {
+		IUserManager $userManager,
+		private LoggerInterface $logger,
+	) {
 		parent::__construct($appName, $request);
 		$this->contactsManager = $contacts;
 		$this->appManager = $appManager;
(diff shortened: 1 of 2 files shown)


lib/Controller/ContactController.php
```diff
@@ -173,32 +178,32 @@ public function searchAttendee(string $search):JSONResponse {
...
 	 * @return JSONResponse
 	 * @throws Exception
-	 * @throws \OCP\AppFramework\QueryException
 	 *
 	 * @NoAdminRequired
 	 */
 	public function getCircleMembers(string $circleId):JSONResponse {
-		if (!$this->appManager->isEnabledForUser('circles') || !class_exists('\OCA\Circles\Api\v1\Circles')) {
+		if (!class_exists('\OCA\Circles\Api\v1\Circles') || !$this->appManager->isEnabledForUser('circles')) {
+			$this->logger->debug('Circles not enabled');
```
kesselb, line 186: Why the debug log?

miaulalala: Why not?

lib/Controller/ContactController.php
```diff
@@ -173,32 +178,32 @@ public function searchAttendee(string $search):JSONResponse {
 	 * @param string $circleId CircleId to query for members
 	 * @return JSONResponse
 	 * @throws Exception
-	 * @throws \OCP\AppFramework\QueryException
 	 *
 	 * @NoAdminRequired
 	 */
 	public function getCircleMembers(string $circleId):JSONResponse {
-		if (!$this->appManager->isEnabledForUser('circles') || !class_exists('\OCA\Circles\Api\v1\Circles')) {
+		if (!class_exists('\OCA\Circles\Api\v1\Circles') || !$this->appManager->isEnabledForUser('circles')) {
```
SebastianKrupinski, line 185: Question, why are we checking if the class exists? Would the check for isEnabledForUser not be sufficient enough to  determine if the circles app is installed, enabled on the system and permitted for this user?

kesselb: The original pr mentions a problem with Psalm https://github.com/nextcloud/calendar/pull/5420#pullrequestreview-1710099584 ;)

I haven't done any profiling but would expect, especially now with the flipped condition, a little performance gain for systems without circles. 

isEnabledForUser does a lot (fetch the installed app versions for all apps, see if the app is installed, fetch the user session, fetch the groups for a user, see if the user has a group that the app is enabled for).

SebastianKrupinski: > isEnabledForUser does a lot (fetch the installed app versions for all apps, see if the app is installed, fetch the user session, fetch the groups for a user, see if the user has a group that the app is enabled for).

I checked the code... Thats not the case, I think the original author just didn't know there was a build in function to see if the app was enabled, and checked for the existence for the class.

SebastianKrupinski (approved):

lib/Controller/ContactController.php
kesselb: > Thats not the case

What is not the case?

lib/Controller/ContactController.php
kesselb: The method is only called, from the frontend, when circles is enabled.

miaulalala: apparently not otherwise we wouldn't need the class_exists check / app enabled check

PR #6322, opened 2024-09-04
enhanc: keep calendar always open

### LINKED ISSUE #6321: Use calendly design and layout
- [x] Time zone and date picker don't have the same width
- [x] Center avatar + name
- [x] Avatar is way too big
- [x] Show the date picker opened permanently instead of toggled from a dropdown?
- [x] Scroll only the times and not the entire page (similar to calendly)
- [x] Mention time zone
- [ ] 

Changes to container

- [x]  Width is fixed at 900px (size of the large modal)
- [x]  margin-bottom: 50px so that the container is horizontally centered
- [x]  use 3 columns, each column is fixed equal width
- [x]  the entire modal does not scroll, only the time slot column
- [x] Changes for first column
- [x]  avatar is 44*44px
- [x] Changes for second column
- [x]  Change font size of "Select date" to normal size (only styling can be changed, semantically it can be h3)
- [x]  disable dates for which no slots are available
- [x]  Time zone picker is full width of the column
- [x] Changes to third column
- [x]  Change font size of "Select slot" to normal size (only styling can be changed, semantically it can be h3)
- [x]  Time slots are full width of the column
- [x]  Time slots list is scrollable

GretaD:
fixes #6321

- [x] requires: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6095

Screenshot from 2024-11-18 13-42-29

### DIFF
--- a/src/views/Appointments/Booking.vue
+++ b/src/views/Appointments/Booking.vue
@@ -11,35 +11,40 @@
 					:display-name="userInfo.displayName"
 					:disable-tooltip="true"
 					:disable-menu="true"
-					:size="180" />
+					:size="44" />
 				<div class="booking__display-name">
 					<strong>{{ userInfo.displayName }}</strong>
 				</div>
-				<h2 class="booking__name">
+				<h5 class="booking__name">
 					{{ config.name }}
-				</h2>
+				</h5>
+				<!-- Description needs to stay inline due to its whitespace -->
 				<span class="booking__description">{{ config.description }}</span>
 			</div>
 			<div class="booking__date-selection">
-				<h3>{{ $t('calendar', 'Select date') }}</h3>
+				<h5 class="booking__date-header">
+					{{ $t('calendar', 'Select a date') }}
+				</h5>
 				<div class="booking__date">
 					<DateTimePicker v-model="selectedDate"
 						:disabled-date="disabledDate"
 						type="date"
+						:open="true"
 						@change="fetchSlots" />
 				</div>
 				<div class="booking__time-zone">
-					<TimezonePicker v-model="timeZone" @change="fetchSlots" />
+					<TimezonePicker v-model="timeZone"
+						:aria-label="$t('calendar', 'Select a date')"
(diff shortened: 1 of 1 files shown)


nimishavijay:
Nice! :) Also see https://github.com/nextcloud/calendar/issues/6324 for how we can improve the design for this as it looks a bit outdated atm

GretaD:
@nimishavijay please have another look at the screenshot. In the appointment case, we cannot use the native datetimepicker, because the native one doesnt allow for the dropdown to be always open, because that is browser defined behavior.

I tried to adopt the normal one, as much as i could, to your mockup in the other ticket.

jancborchardt (commented):
Looks quite good! :) 
- [ ] One of the most important improvements UX-wise is: "disable dates for which no slots are available", is that possible here?
- [x] The time zone picker doesn’t need a full heading, or just "Time zone", not "Select a time zone"
- [ ] The "Select a date and time slot" would be good to split, or if the time slots are shifted a bit down so it fits with the layout in @nimishavijay’s mockup
- [ ] Regarding the third column: Currently the vertical gap seems rather large, it could be reduced a little bit and instead there can be a larger gap for every hour as a little divider. In many cases like here, there would then be slots of 4 grouped together to get a better feeling for the time rather than a long list.

SebastianKrupinski:
Tested. Looks and works great

SebastianKrupinski (approved):

GVodyanov:
Yeah it's the same now even without this PR, I just discovered it, I will open an issue for it.

image

SebastianKrupinski:
> Yeah it's the same now even without this PR, I just discovered it, I will open an issue for it.

Yeah, I noticed that and thought it was a new feature!

st3iny:
> > Yeah it's the same now even without this PR, I just discovered it, I will open an issue for it.
> 
> Yeah, I noticed that and thought it was a new feature!

Can't click the dates if you can't seem em :laughing:

PR #6356, opened 2024-09-18
feat(appointments): add timezone to all emails

miaulalala:


lib/Service/Appointments/MailService.php
```diff
@@ -217,21 +217,24 @@ private function addBulletList(IEMailTemplate $template,
...
 			$dtstart,
 			'long',
 			'short',
-			new \DateTimeZone($tzid),
+			$relativeTimezone,
 			$this->lFactory->get('calendar', $l)
 		);
 
-		$template->addBodyListItem($relativeDateTime, $l10n->t('Date:'));
+		$timestring = $relativeDateTime . ' (' . $relativeTimezone->getName() . ')';
```
tcitworld, line 235: A nice future improvement we could have is using `IntlTimeZone::getDisplayName` to have the timezone name localized, if the `intl` module is available.

tcitworld, line 235: Do we need to localise this? For locales that don't use brackets or invert the order of things?

miaulalala: I asked Christoph and he said it won't be possible to localise it - so I agree with your future improvement idea there.

st3iny (approved):
Tested and works.

PR #6365, opened 2024-09-24
feat: disable autocompletion for event property field

kesselb:
Motivation:

I just added 8 events to my calendar, and my password manager asked me to unlock the vault for the auto suggestions for every event title. The password manager should do better, but autocompletion for the field seems to have little use and Google Calendar also uses autocomplete off for the event title.


st3iny (approved):
Makes sense and looks good.

PR #6371, opened 2024-09-25
"Appointment schedule" notion instead of only "Appointments" for a better UX

Jerome-Herbinet:
"Appointment" notion is no clear and creates confusion because this feature don't directly create an appointment but create an "appointment request form". I propose a new notion : "Appointment request form".
I made some wording change test with my browser dev tools. Please check below (real tests from reviewers are welcome) : 

Capture d’écran du 2024-09-25 11-40-30
Capture d’écran du 2024-09-25 11-41-37
Capture d’écran du 2024-09-25 11-48-00
Capture d’écran du 2024-09-25 11-51-43
Capture d’écran du 2024-09-25 11-52-29

### DIFF
--- a/src/components/AppNavigation/AppointmentConfigList.vue
+++ b/src/components/AppNavigation/AppointmentConfigList.vue
@@ -11,7 +11,7 @@
 	<div v-if="hasAtLeastOneCalendar"
 		class="appointment-config-list">
 		<AppNavigationCaption class="appointment-config-list__caption"
-			:name="t('calendar', 'Appointments')">
+			:name="t('calendar', 'Appointment schedules')">
 			<template v-if="hasUserEmailAddress"
 				#actions>
 				<ActionButton :close-after-click="true"
@@ -19,7 +19,7 @@
 					<template #icon>
 						<PlusIcon :size="20" decorative />
 					</template>
-					{{ t('calendar', 'Add new') }}
+					{{ t('calendar', 'Create new') }}
 				</ActionButton>
 			</template>
 		</AppNavigationCaption>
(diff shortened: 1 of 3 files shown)


nimishavijay (changes requested):
This exact issue came up in a design review call! :) The discussion led to the conclusion that "Appointment schedule" is more appropriate. This is also on parity with Google Calendar (with uses the wording "Appointment schedule" to set this up) and Calendly (which calls itself an "Appointment scheduling software"). What do you think?

Jerome-Herbinet:
> Appointment schedule

> This exact issue came up in a design review call! :) The discussion led to the conclusion that "Appointment schedule" is more appropriate. This is also on parity with Google Calendar (with uses the wording "Appointment schedule" to set this up) and Calendly (which calls itself an "Appointment scheduling software"). What do you think?

Thanks for your reply @nimishavijay ! Despite "Appointment request form" remains my favorite wording (personal preference), your suggestion is of course the best one because it fits wordings used by Google and it will make UX transition easier for end users. So, I'll make some new commits.

Jerome-Herbinet:
@nimishavijay changes are done ; can you check and test it ?

nimishavijay (changes requested):
Nice! Only a small change that I didn't notice before: "Appointment schedules" (plural form)

Jerome-Herbinet:
> Nice! Only a small change that I didn't notice before: "Appointment schedules" (plural form)

Done @nimishavijay

nimishavijay (approved):

st3iny (approved):

Issue #6385, opened 2024-10-01
Improve Date/Time selector to be more intuitive.


SebastianKrupinski:
One of the recommendations during the brain storming sessions was the improvement of the time selector. This was one of the top requested items.

Currently,
Clicking the Date/Time selector always brings up the time picker if the event is not all day. The time picking part of the selector is also ticky to understand.

![image](https://github.com/user-attachments/assets/3b568fd7-7a7b-488f-9bc9-ac4714f9a014)

Possible Improvements,

1. Leaving the current selection box as is but showing a different drop down depending on if the user clicks the date or time. 

2. Splitting the Date and Time selector in to two parts like the tasks app dose.

![image](https://github.com/user-attachments/assets/59ee8f31-c69c-4f70-9246-b450097b0c08)

3. Also improving the look of the date/time selector to be more understandable. Maybe something like the material design date and time pickers.

![image](https://github.com/user-attachments/assets/be61052f-2ded-45d0-978d-06c07ca4f73e)

![image](https://github.com/user-attachments/assets/1cdb9635-a534-4871-9a61-07dc13153678)

ChristophWurst:
This might actually be a duplicate of https://github.com/nextcloud/calendar/issues/6324

ChristophWurst:
There is also https://github.com/nextcloud/calendar/issues/4626

Before we spend too much time on redesigning the current picker we should re-evaluate if it can actually be made accessible. The last time we checked it was not. Many apps and components switched to a native time picker. We might want to do the same for Calendar.

SebastianKrupinski:
The native date/time picker is an option but there are some down side, they can't really be styled and the styling varies between browser versions, and the native time pickers are horrible.

You can test them on here... 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

There are some options for javascript based accessible date/time picker.

https://www.digitala11y.com/accessible-date-pickers-roundup/

The Duet Date picker seems to be a good alternative.

ChristophWurst:
See https://github.com/nextcloud-libraries/nextcloud-vue/issues/1666 as well. We evaluated some pickers in the past.

GVodyanov:
My proposal is using PrimeVue's components https://primevue.org/datepicker/#time. There is support for time and date, but not timezone, unfortunately. It's also under the MIT license. My only doubt would be the accessibility as I don't really know how to test for that.

SebastianKrupinski:
[...]

Issue #4626, opened 2022-10-14
Replace datetime picker with native one on new event modal


JuliaKirschenheuter:
### Steps to reproduce

1. Click into your calendar grid to create an event
2. Pick start/end time

### Expected behavior

Native datetime picker

### Actual behaviour

![Bildschirmfoto vom 2022-10-10 09-34-46](https://user-images.githubusercontent.com/1374172/194819096-b6bbe649-9770-4c63-9cdb-9c92fca450d6.png)

### Calendar app version

4.0.0

corbin-auriti:
Is anyone else working on this? Otherwise I would like to along with a "natural language" title datetime grabber similar to Elementary OS Calendar and others https://github.com/elementary/calendar/commit/8a2063007acd73cbecf8d5c6877aafe20ee8b214

jancborchardt:
I checked with Firefox, also referencing here from https://github.com/nextcloud/calendar/issues/6385#issuecomment-2432512262

> On the Firefox side, the bug about a time picker is tracked as https://bugzilla.mozilla.org/show_bug.cgi?id=1811581 and/or https://bugzilla.mozilla.org/show_bug.cgi?id=451832
>
> So I would say we indeed just accept the tradeoff that Firefox does not have a visual time picker. The keyboard-only input is accessible at least, and then hopefully they will soon activate the visual timepicker.
>
> And as written in the issue, they actually have a timepicker interface, but it unfortunately is hidden behind the `about:config` switch `dom.forms.datetime.timepicker`. If set to true, you see the picker, e.g. on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time
> ![Image](https://github.com/user-attachments/assets/ee5cf201-6216-4166-8502-ba6c35613d35)

PR #6423, opened 2024-10-14
feat/Separate date and time picker

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6385

### DIFF
--- a/src/components/Editor/Properties/PropertyTitleTimePicker.vue
+++ b/src/components/Editor/Properties/PropertyTitleTimePicker.vue
@@ -30,6 +30,7 @@
 				@change="changeEnd"
 				@change-timezone="changeEndTimezone" />
 		</div>
+		<TimePicker :initial-date="startDate" />
 		<div v-if="isReadOnly"
 			class="property-title-time-picker__time-pickers property-title-time-picker__time-pickers--readonly">
 			<div class="property-title-time-picker-read-only-wrapper property-title-time-picker-read-only-wrapper--start-date">
@@ -68,6 +69,7 @@
 <script>
 import moment from '@nextcloud/moment'
 import DatePicker from '../../Shared/DatePicker.vue'
+import TimePicker from '../../Shared/TimePicker.vue'
 import IconTimezone from 'vue-material-design-icons/Web.vue'
 import CalendarIcon from 'vue-material-design-icons/Calendar.vue'
 import { NcCheckboxRadioSwitch } from '@nextcloud/vue'
@@ -78,6 +80,7 @@ export default {
 	name: 'PropertyTitleTimePicker',
 	components: {
 		DatePicker,
+		TimePicker,
 		IconTimezone,
 		CalendarIcon,
 		NcCheckboxRadioSwitch,
@@ -280,3 +283,9 @@ export default {
 	},
 }
 </script>
+
+<style scoped>
+:deep(.button-vue--icon-only), :deep(.button-vue__icon) {
+	width: 7rem !important;
+}
+</style>
--- a/src/components/Shared/TimePicker.vue
+++ b/src/components/Shared/TimePicker.vue
@@ -4,62 +4,54 @@
 -->
 
 <template>
-	<DateTimePicker ::clearable="false"
-		:first-day-of-week="firstDay"
-		:format="format"
-		:lang="lang"
-		:minute-step="5"
-		:show-second="false"
-		type="time"
-		:use12h="showAmPm"
-		:value="date"
-		v-bind="$attrs"
-		v-on="$listeners"
-		@change="change" />
+	<NcActions>
+		<template #icon>
+			<NcTextField :value.sync="date">
+				<template #trailing-button-icon>
+					<ClockOutline/>
+				</template>
+			</NcTextField>
+		</template>
+		<NcActionButton v-for="time in timeList" :key="time" @click="change(parse(time))">
+			<template #icon></template>
+			{{ time }}
+		</NcActionButton>
+	</NcActions>
 </template>
 
 <script>
-import { NcDateTimePicker as DateTimePicker } from '@nextcloud/vue'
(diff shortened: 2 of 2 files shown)


GVodyanov:
Hey @nimishavijay I know this doesn't exactly match your spec, but I wasn't able to include the text and icons inside the date picker input, as it's the native browser one. Please tell me if it could still work and if you might have any ideas on how to make it better seeing the limitations.

st3iny (commented):
Looks amazing so far!

I noticed two small issues during my testing:
1. All day events can not have time zones. The time zone select should be hidden when `All day` is checked.
2. Some design issues during my testing. See below for screenshots.

## Popover editor read-only wrapping

grafik

## Sidebar editor overflowing

grafik

GVodyanov:
@st3iny Thanks for the review! 

I hadn't looked at the read-only part, but now that I did... it was a wonder it (barely) worked, lol. Anyways here's how I made it look:

image

And the other stuff you mentioned is fixed now too :)

GVodyanov:
Screenshot from 2024-11-12 20-37-31

I noticed that this thing here needs to be migrated too... figuring out with Christoph what to do about it, but ignore it in reviews for now

SebastianKrupinski:
Hey, just tested this. This is what I noticed. 

Also some of these observations might need more input from the team on what the logic should be. 

- (popup modal only) When changing the start time to a time greater then the end time, the end time does not change to reflect that change and stays in the past (before start time), should adjust automatically with the same time span as before.

image

- (popup modal only) When changing the date, after changing the time, the time resets to the original value

- Time zone input does not align with date input. (looks odd) I think is should be inline with date input.

image

- Read only pop-up should show both time zones if they differ

image

- Changing start time zone automatically changes end time zone. Hard to say what the logic should be here.

- Changing the start date to a date prior to set date automatically changes the end date, I don't think this should automatically change the end date unless the end date is before the start date.

nimishavijay:
Looks really nice! :) Most of the comments are regarding spacing and alignment: 
- [ ] the width of the entire container can be reduced if the only elements shown are the date picker and time picker. Right now in the first screenshot, both pickers are rather wide. So if the width of the whole card can be reduced by ~90px, then even the width of the pickers would be lesser
- [ ] the width of the timezone dropdown is more than needed. Would it be possible to adjust that to be ~30px lesser?
- [ ] possibly beyond the scope of this issue but is it possible to reduce the width of the dropdown in the reminder selector? It's way too wide for a single word
- [ ] could we un-bold the text in the timezone button? And possibly remove the edit icon? (It's a deviation from the mockups but I checked with the rest of the team and we agree that the button is too eye catching if it's bold + has 2 icons)

> I wasn't able to include the text and icons inside the date picker input, as it's the native browser one. 

Do you think we could use the floating labels like we do for input fields? 

Nice-to-have also in the mockups: the first time someone edits a time zone, the change is reflected across both the input fields (because it most likely that you want to change the timezone for both the start and end time).

GVodyanov:
Right, so, @nimishavijay @SebastianKrupinski 

> (popup modal only) When changing the start time to a time greater then the end time, the end time does not change to reflect that change and stays in the past (before start time), should adjust automatically with the same time span as before.

Yeah didn't notice, will fix

> (popup modal only) When changing the date, after changing the time, the time resets to the original value

I thought I was just seeing things because it worked in the sidebar... weird, will look into it

> Time zone input does not align with date input. (looks odd) I think is should be inline with date input.

Fair enough

> Read only pop-up should show both time zones if they differ

Well here I tried to do only graphical changes, this is the way it worked before, but sure seeing as we're here may as well change that

> Changing start time zone automatically changes end time zone. Hard to say what the logic should be here.

Intentional, see https://github.com/nextcloud/calendar/issues/6460

> the width of the entire container can be reduced if the only elements shown are the date picker and time picker. Right now in the first screenshot, both pickers are rather wide. So if the width of the whole card can be reduced by ~90px, then even the width of the pickers would be lesser

Yep, previously the container was `600px` fixed, now I gave it the possibility to expand. I didn't think you would want shrinking as well, but sure will do.

> the width of the timezone dropdown is more than needed. Would it be possible to adjust that to be ~30px lesser?

Sure

> possibly beyond the scope of this issue but is it possible to reduce the width of the dropdown in the reminder selector? It's way too wide for a single word

EZ

>  could we un-bold the text in the timezone button?

Yeah, sure it's easy for me to do, but `NcButton` doesn't currently support it. https://nextcloud-vue-components.netlify.app/#/Components/NcButton. What I would do here is make an exception and CSS override the boldness, but maybe if we want to start having non-bold buttons in the app in general, we could add it to nextcloud-vue? I feel like this is the sort of thing you would want to standardize

> And possibly remove the edit icon? (It's a deviation from the mockups but I checked with the rest of the team and we agree that the button is too eye catching if it's bold + has 2 icons)

| | |
| - | - |
| image | image |

The [mockup](https://github.com/nextcloud/calendar/issues/6385#issuecomment-2401043906) first didn't have the icon... and then did later in the video. I could remove the button altogether, or make it appear only in specific situations, but I didn't understand what this condition was supposed to be from the video.

> Do you think we could use the floating labels like we do for input fields?

I talked about this with Christoph, and it's really best not to. First of all, chrome already has the icon actually, it's just firefox which is sadder. Hacking it on is possible but very prone to breaking in the future and could possibly compromise functionality in the case of a browser redesign, so it's probably best not to.

GVodyanov:
Hey peeps! I'm back with all the changes from my previous comment. What do you think?

image

image

image

image

SebastianKrupinski:
Tested Again. Works Great.

SebastianKrupinski (approved):

ChristophWurst (approved):
Tested & works!

So this not only fixes https://github.com/nextcloud/calendar/issues/6385 but also https://github.com/nextcloud/calendar/issues/4626?

PR #6435, opened 2024-10-19
fix: added required composer package for psalm

SebastianKrupinski:
psalm was failing to run due to missing package.

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
     },
     "require": {
         "php": ">= 8.1 <=8.3",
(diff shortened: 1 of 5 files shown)


st3iny (changes requested):
We omit the OCP package on purpose because it
1. pollutes the autoloader,
2. is hard to keep in sync,
3. and will be installed automatically during CI.

You can run the following command install the package before running psalm:
`composer require --dev nextcloud/ocp:dev-master --ignore-platform-reqs --with-dependencies`

You can also run it against stable30 by replacing `dev-master` with `dev-stable30` etc.

PR #6436, opened 2024-10-19
fix: Entity parameter types

SebastianKrupinski:
Psalm was failing on the entity parameter types.

image


ChristophWurst:
See https://github.com/nextcloud/server/pull/47329#discussion_r1807934745. 'int' might return.

st3iny (approved):
Thanks!

PR #6507, opened 2024-11-20
[stable5.0] ci: Add release automation

ChristophWurst:
* Add a fully automated workflow for releases. It does the same as a manual process, except for any milestone cleanups.
* Change the old workflow to only run manually to avoid a double release with the new workflow pushing over to the releases org
* Set up the repository release environment and add necessary keys

This workflow released Mail v4.0.2 yesterday.


ChristophWurst:
Reviewers unavailable right now and a release is due -> merge without review

PR #6524, opened 2024-11-26
Remove 500ms focus delay from PropertyTitle

ochen1:
02a2914 was a great commit but 3 years later I am unsure why the arbitrary 500ms delay is necessary. I want to remove it.

Use case: I have a userscript that enables Google Calendar keybindings (eg. e to edit) and this delay significantly impedes my workflow (when editing the name of an event). The entire name-editing process should take 500ms total, but with the current configuration the user must wait *at least* 500ms for the input to focus.

I tested the removal of the delay via devtools resource override on Edge and the focus still works consistently. I hope this minor change can make the event name editing pipeline more performant for everyone.

### DIFF
--- a/src/directives/focus.js
+++ b/src/directives/focus.js
@@ -5,8 +5,6 @@
 
 export default {
 	inserted(el) {
-		setTimeout(() => {
-			el.focus()
-		}, 500)
+		el.focus()
 	},
 }


ochen1:
I tested event creation and I now understand the original intention behind the 500ms delay. The popover takes some time to mount if it didn't exist before (e.g. as a quick preview), while the directive to focus is called immediately. The `e.focus()` call within the directive has no effect if the PropertyTitle input is not yet mounted. As such, I propose calling it twice to accommodate for both cases. In the original case, we can hope nobody finishes making title edits in under 100ms and decides to tab away to edit the date as well. Even if they do, I believe it would be an improvement from having to wait 500ms to start making edits.

A more elegant solution would probably involve finding a way to avoid triggering the focus directive until the necessary components are fully mounted and visible.

SebastianKrupinski:
Hi @ochen1

I see no reason not to accept your PR but we need to see what the affects are first.

On that note we have made significant changes to the calendar in the last two months in our main branch, have you pulled the latest changes to see how it all looks and works now?

Also, a screenshot or video of what we are looking for would be helpful for testing purposes.

st3iny (commented):
Hi, thank you for the PR.

We discussed this and we would like to keep the current delay. A very short delay like 100 ms might not work on all machines.

I experimented with this a bit and I don't think that we can get rid of the timeout completely. The problem is that the focus has to be triggered after the input field (and popover) is fully rendered.

ochen1:
Understood, thank you for the review.

My power tools userscript contains this snippet, in case there are any other power users out there also interested in removing the 500ms focus delay. Tested with 4.7.13. Vibe coded.

```js
    // Monkey patch window.setTimeout to remove 500ms focus delay
    const originalSetTimeout = window.setTimeout;
    window.setTimeout = function(callback, delay, ...params) {
        const isFiveHundredMs = (delay === 500);
        if (!isFiveHundredMs) return originalSetTimeout.apply(window, [callback, delay, ...params]);
        const logStyle = isFiveHundredMs ? 'font-weight: bold; color: orange;' : 'color: dodgerblue;';
        const callbackType = typeof callback;
        let callbackRepresentation;

        if (callbackType === 'function') {
            // Try to get a concise representation of the function
            callbackRepresentation = callback.toString().split('\n')[0].substring(0, 100); // First line, up to 100 chars
            if (callback.toString().length > 100 || callback.toString().includes('\n')) {
                callbackRepresentation += '...';
            }
        } else if (callbackType === 'string') {
            // setTimeout can also take a string to eval (bad practice, but supported)
            callbackRepresentation = `"${callback.substring(0, 100)}${callback.length > 100 ? '...' : ''}" (string eval)`;
        } else {
            callbackRepresentation = `Unsupported type: ${callbackType}`;
        }
        console.log(`%c Call registered: ${callbackRepresentation}`, logStyle);
        if (isFiveHundredMs) delay = 50;
        // Call the original setTimeout with the correct 'this' context (window) and arguments
        // and return its ID so the calling code works as expected.
        return originalSetTimeout.apply(window, [callback, delay, ...params]);
    };
```

Issue #6528, opened 2024-11-27
Appointment slots are generated for the wrong day due to UI sending date in epoch


SebastianKrupinski:
### Steps to reproduce

1. Create appointment schedule in local time zone (Europe/Berlin) UTC+1
2. Change browser time zone to (Asia/Jerusalem) UTC+2
3. Select a date e.g. 2024-11-26

### Expected behavior

Should show slots for 2024-11-26

### Actual behaviour

Shows slots for 2024-11-25

### Calendar app version

Main

### Test Results

Selected November 26, 2024 in Asia/Jerusalem which is epoch 1732579200000.

Epoch conversion for Asia/Jerusalem time zone
```
1732579200000 converts to November 26, 2024 00:00:00 (am) in time zone Asia/Jerusalem (IST)
```

Epoch conversion for Europe/Berlin
```
1732579200000  converts to November 25, 2024 23:00:00 (pm) in time zone Europe/Berlin (CET)
```

Epoch conversion for America/Toronto
```
1732579200000  converts to November 25, 2024 19:00:00 (pm) in time zone America/Toronto (EST)
```

### Issue
UI sends the date in the browsers time zone as epoch without sending the actual time zone. The issue is that epoch is a static number of seconds from a specific start point and defines a different time for every time zone, so once the configuration time zone is applied to the date, this causes the date to roll back one day causing slots to be generated for the wrong date. Epoch time will also differ during Standard and Daylight Savings time.

### Solution
Use string date instead of epoch time stamp

PR #6532, opened 2024-11-27
fix: send date as string instead of epoch

SebastianKrupinski:
Resolves: https://github.com/nextcloud/calendar/issues/6528

Testing Recommendation
1. Test with existing appointment schedule and fresh appointment schedule (make sure when creating a new appointment schedule the browser time zone matches the calendar time zone, as there is another bug that will affect this)
2. Test by selecting different time zones negative (North America/South America), positive (Australia/China/Etc) and matching the appointment schedule
3. Test by changing browser time zone (either override browser setting or change system time zone). You can check your browser time zone here: https://webbrowsertools.com/timezone/
4. Test by selecting time zones with an offset greater than the current time. (if the current time is 10.00 select zone that has offset of +11 hours, if the time is 14.00 select timezone that is -11 hours, this would in the past roll the date back one day)
5. Test by selecting dates close to year end and start
6. Compare all slot times to a external source I recommend https://www.worldtimebuddy.com/ or https://www.timeanddate.com/worldclock/converter.html you can preset all testing time zones and compare the times

### DIFF
--- a/src/views/Appointments/Booking.vue
+++ b/src/views/Appointments/Booking.vue
@@ -195,12 +195,14 @@ export default {
 			this.slots = []
 			this.loadingSlots = true
 
-			const startOfDay = new Date(this.selectedDate.getTime())
+			const selectedDay = this.selectedDate.getFullYear().toString() + '-'
+								+ (this.selectedDate.getMonth() + 1).toString() + '-'
+								+ this.selectedDate.getDate().toString()
 
 			try {
 				this.slots = await findSlots(
 					this.config,
-					Math.round(startOfDay.getTime() / 1000),
+					selectedDay,
 					this.timeZone,
 				)
 			} catch (e) {
--- a/lib/Controller/BookingController.php
+++ b/lib/Controller/BookingController.php
@@ -8,7 +8,6 @@ declare(strict_types=1);
 namespace OCA\Calendar\Controller;
 
 use DateTime;
-use DateTimeImmutable;
 use DateTimeZone;
 use InvalidArgumentException;
 use OCA\Calendar\AppInfo\Application;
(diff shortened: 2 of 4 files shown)


src/views/Appointments/Booking.vue
```diff
@@ -188,12 +188,14 @@ export default {
 			this.slots = []
 			this.loadingSlots = true
 
-			const startOfDay = new Date(this.selectedDate.getTime())
+			const selectedDay = this.selectedDate.getFullYear().toString() + '-' + 
+								(this.selectedDate.getMonth() + 1).toString() + '-' + 
+							    this.selectedDate.getDate().toString()
```
st3iny, line 193: We need proper formatting here. Month and date might only have a single digit.

Please use the `getYYYYMMDDFromDate(date)` from `src/utils/date.js` which already implements formatting as `YYYY-MM-DD`.

st3iny (changes requested):
Now it seems to make sense. I left a comment regarding the formatting in the frontend.

src/views/Appointments/Booking.vue
st3iny:
```js
import { getYYYYMMDDFromDate } from '../../utils/date.js'

// [...] further down

const selectedDay = getYYYYMMDDFromDate(this.selectedDate)
```

st3iny: Yeah, that is why Georg reverts the shift. See `date.getTime() - (date.getTimezoneOffset() * 60000))`

I tested it and it should work correctly.

SebastianKrupinski: Yeah, I don't trust the calculation in getYYYYMMDDFromDate

Tested with:

- Value In
Wed Nov 27 2024 14:44:54 GMT-0500 (Eastern Standard Time)

- Value produced by "new Date(date.getTime() - date.getTimezoneOffset() * 60000)"
Wed Nov 27 2024 09:44:54 GMT-0500 (Eastern Standard Time)

As you can see it shifted the time by 5 hours. This would have the effect of shifting the date, if the input value time was less then the offset.

E.g.
Input 
Wed Nov 27 2024 2:00:00 GMT-0500 (Eastern Standard Time)

Result
Wed Nov 26 2024 21:00:00 GMT-0500 (Eastern Standard Time)

Then toISOString produces this 
2024-11-27T14:44:54.790Z

The end result seems correct in my browser and time zone but will this be correct for every time zone and browser. IDK

st3iny: I get that it is confusing. The thing is, that JavaScript's Date is not time zone aware. It is **always** in the local time zone, regardless which time zone you pass in the constructor.

We have unit tests in place to ensure that this implementation works and it is used throughout Calendar.

If it breaks we can fix it. The important thing is that we have one central implementation and not many copied across the code base because this is way harder to maintain.

st3iny: See also https://stackoverflow.com/a/29774197/7840137

st3iny: But yeah, let's not get pedantic over 3 lines of code.

st3iny (approved):

ChristophWurst (approved):

ChristophWurst:
@GVodyanov could you give this a test with the instructions provided? :)

GVodyanov:
Hey, would you be able to explain this? I'm pretty confused as to why just that Tuesday doesn't work (all other days as far as I can see are fine)

Screenshot from 2024-12-09 19-33-42
Screenshot from 2024-12-09 19-33-28

https://github.com/user-attachments/assets/1a6acdba-aeba-490a-83c6-c0af7d4935c9

SebastianKrupinski:
> Hey, would you be able to explain this? I'm pretty confused as to why just that Tuesday doesn't work (all other days as far as I can see are fine)

Do you have an "all day event" on that Tuesday? That would cause the slots to be filtered out as a conflict during that time.

GVodyanov:
> > Hey, would you be able to explain this? I'm pretty confused as to why just that Tuesday doesn't work (all other days as far as I can see are fine)
> 
> Do you have an "all day event" on that Tuesday? That would cause the slots to be filtered out as a conflict during that time.

I'm dumb

GVodyanov (approved):

SebastianKrupinski:
> > Do you have an "all day event" on that Tuesday? That would cause the slots to be filtered out as a conflict during that time.
> 
> I'm dumb

🤣 don't worry, I did the same thing

PR #6534, opened 2024-11-29
Event attachments: Enable directory selection

kesselb:
**Summary:**

It’s possible to attach files to an event. An internal or public share link is stored within the event.

On the main branch, adding a folder currently doesn’t work (while selection is possible, the folder isn’t actually added). It’s unclear to me from the initial implementation https://github.com/nextcloud/calendar/pull/4251 whether folder support was intended or not.

Folders typically don’t have a content type returned, so I added a fallback to display the folder icon.

| B | A |
|--------|--------|
| Screenshot From 2024-11-29 15-57-02 | Screenshot From 2024-11-29 15-39-05 | 

> [!WARNING]
> Adding another attendeed will generate a public share link instead of the internal link. This is apparently broken as well, I will try to address that in a follow up or log an issue.

**How to test:**

- Create an event
- Try to attach a folder
- Nothing happens on main, it works on this pr.

### DIFF
--- a/src/components/Editor/Attachments/AttachmentsList.vue
+++ b/src/components/Editor/Attachments/AttachmentsList.vue
@@ -220,7 +220,9 @@ export default {
 			if (attachment.xNcHasPreview) {
 				return generateUrl(`/core/preview?fileId=${attachment.xNcFileId}&x=100&y=100&a=0`)
 			}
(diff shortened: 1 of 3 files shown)


SebastianKrupinski (approved):
Tested. Works

Issue #6501, opened 2024-11-17
Drag to reorder calendars in the browser is broken


DireMunchkin:
### Steps to reproduce

1. In the calendar view in the web app, drag to reorder the calendars.

### Expected behavior

The calendars are reordered according to user preference

### Actual behaviour

A JavaScript error is thrown in the browser and the order isn't changed.

### Calendar app version

5.0.1

### Browser

Librewolf 132.0.1-1

### Client operating system

Linux Mint

### Server operating system

Linux Mint

### Web server

Nginx

### Database engine version

PostgreSQL

### PHP engine version

Other

### Nextcloud version

30.0

### Updated from an older installed version or fresh install

Updated from an older version

### List of activated apps

```Text
Enabled:
  - app_api: 4.0.0
  - bruteforcesettings: 3.0.0
  - calendar: 5.0.1
  - cloud_federation_api: 1.13.0
  - contacts: 6.1.1
  - dashboard: 7.10.0
  - dav: 1.31.1
  - federatedfilesharing: 1.20.0
  - files: 2.2.0
  - files_downloadlimit: 3.0.0
  - files_external: 1.22.0
  - lookup_server_connector: 1.18.0
  - oauth2: 1.18.1
  - provisioning_api: 1.20.0
  - settings: 1.13.0
  - theming: 2.5.0
  - twofactor_backupcodes: 1.19.0
  - viewer: 3.0.0
  - webhook_listeners: 1.1.0-dev
  - workflowengine: 2.12.0
Disabled:
  - activity: 3.0.0 (installed 2.17.0)
  - admin_audit: 1.20.0
  - circles: 30.0.0 (installed 25.0.0)
  - comments: 1.20.1 (installed 1.15.0)
  - contactsinteraction: 1.11.0 (installed 1.6.0)
  - encryption: 2.18.0
  - federation: 1.20.0 (installed 1.15.0)
  - files_pdfviewer: 3.0.0 (installed 2.6.0)
  - files_reminders: 1.3.0 (installed 1.1.0)
  - files_sharing: 1.22.0 (installed 1.17.0)
  - files_trashbin: 1.20.1 (installed 1.15.0)
  - files_versions: 1.23.0 (installed 1.18.0)
  - firstrunwizard: 3.0.0 (installed 2.14.0)
  - logreader: 3.0.0 (installed 2.10.0)
  - nextcloud_announcements: 2.0.0 (installed 1.14.0)
[...]

PR #6536, opened 2024-11-30
Fix calendar lists not being orderable any more

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6501

So the problem with this bug is that the way that dragging the calendars previously worked is conceptually incompatible with the division of calendars into different sections, mostly because of the hidden calendars. The hidden calendars being part of the order, but being in a different place from all the others, causes the order coming from `vue-draggable` to be wrong. Another issue discussed with Christoph is the drag to reorder feature is pretty hard to discover.

Here are a few things I thought about:
- Instead of having one big orderable section, have 3, one for personal, one for shared, and one for deck. This however still doesn't fix the issue of hidden calendars.
- One potential solution to the hidden calendars is making all calendars unhidden when the users starts dragging, but this would be pretty jarring. 
- Have some sort of button indicating that you are able to reorder the calendars (should probably talk with designers)

In the end to make this happen I thought of adding a button which opens a modal with the three sections: personal, shared, and deck, and WITHOUT the hidden calendars section.

### DIFF
--- a/src/components/AppNavigation/CalendarList.vue
+++ b/src/components/AppNavigation/CalendarList.vue
@@ -4,17 +4,19 @@
 -->
 
 <template>
-	<draggable v-model="calendars"
-		:disabled="disableDragging"
-		v-bind="{swapThreshold: 0.30, delay: 500, delayOnTouchOnly: true, touchStartThreshold: 3}"
-		draggable=".draggable-calendar-list-item"
-		@update="update">
+	<div class="calendar-list-wrapper">
 		<CalendarListNew />
 		<template v-if="!isPublic">
-			<CalendarListItem v-for="calendar in sortedCalendars.personal"
-				:key="calendar.id"
-				class="draggable-calendar-list-item"
-				:calendar="calendar" />
+			<draggable v-model="sortedCalendars.personal"
+				:disabled="disableDragging"
+				v-bind="{swapThreshold: 0.30, delay: 500, delayOnTouchOnly: true, touchStartThreshold: 3}"
+				draggable=".draggable-calendar-list-item"
+				@update="updateInput">
+				<CalendarListItem v-for="calendar in sortedCalendars.personal"
+					:key="calendar.id"
+					class="draggable-calendar-list-item"
+					:calendar="calendar" />
+			</draggable>
 		</template>
 		<template v-else>
 			<PublicCalendarListItem v-for="calendar in sortedCalendars.personal"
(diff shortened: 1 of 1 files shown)


GVodyanov:
What it could look like:

image

GVodyanov:
@ChristophWurst The issue with ignoring hidden calendars is that even though they are hidden, they still make up the "order" of the calendars. So if you have:

- Calendar ID 1
- Calendar ID 2 (hidden)
- Calendar ID 3
- Calendar ID 4

With the order [1, 2, 3, 4]

The draggable list would show:
- Calendar ID 1
- Calendar ID 3
- Calendar ID 4

Now imagine you move Calendar ID 4 to the second position, resulting in:
- Calendar ID 1
- Calendar ID 4
- Calendar ID 3

The order could either be [1, 2, 4, 3] or [1, 4, 2, 3] seeing as the hidden calendar, even if not displayed in the UI, still is part of the list. This causes draggable to give weird results.

There are a few ways that come to mind to solve this: show the hidden calendars when dragging (jarring because they appear out of nowhere), append hidden calendars to the end (messing up the order when u want to hide a calendar), etc. The modal is also one.

ChristophWurst:
> append hidden calendars to the end (messing up the order when u want to hide a calendar)

This :)

See hidden list as extension of the visible calendars list. In your example, the order will be  [ 1 4 3 2 ]. When you unhide the calendar it will be at the bottom of the visible calendars, which seems acceptable to me.

SebastianKrupinski (approved):
Tested. Works. 

Arranging calendar list order works, unhiding a calendar adds the calendar to the end of the list.

LnLcFlx:
> When you unhide the calendar it will be at the bottom of the visible calendars, which seems acceptable to me.

To me, unfortunately, it is not.
I, and I assume others too, often hide and un-hide calendars based on what I have to see rn (I have lots of calendars), resulting in an unordered mess which one would have to reorder upon every hide/un-hide.

Would the following "align-at-top" solution work, technically?:
- When I drag a calendar B to a new position, what I usually intend to do is to *place said calendar B below another calendar A (in the visible list)*.
  (At least that's how my brain works, after all the list expands from top to bottom).
- Would it be possible to fetch calendar A above calendar B and just insert the calendar ID of B right after A?
- In the above example: When I move calendar 4 up in the visible list to *1, 4, 3* 
  This would turn into a unique ordering [1, 4, 2, 3] because 1 is the calendar in the visible list above 4 after reordering.
- If I move a calendar to the top of the visible list, this ofc should result in an ordering that starts with that calendar (no matter if previously another hidden calendar took the first spot).

It is not perfect but the most intuitive approach, that solves this issue, I could come up with.

ChristophWurst:
Thanks for the proposal, @LnLcFlx. Could you please put this into a new feature request ticket so the idea doesn't get lost?

PR #6551, opened 2024-12-05
Fixes Sidebar editor repetition count for "days"

### LINKED ISSUE #6429: Sidebar editor repetition count for "days" is not put on separate line
### Steps to reproduce

Edit an event with a screen size lower than 960px

Image

### Expected behavior

The form field is put on a separate line

### Actual behaviour

The form field breaks the sidebar editor

### Calendar app version

5.0.2 (main)

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

_No response_

None

### List of activated apps

_No response_

### Nextcloud configuration

_No response_

### Web server error log

_No response_

### Log file

_No response_

### Browser log

_No response_

### Additional info

_No response_

keremsemiz:
Fixes #6429

Adjusted CSS for proper centering of the Repeat section in the sidebar at low px counts.
image


SebastianKrupinski:
Tested. Works.

But it does not seem to inline when the screen is large enough.

image

GretaD:
yes, @keremsemiz the "repeat every" field should be smaller, so both input field can fit in one row, because the content they hold is very small. So remove the flex-direction: column for that.

The second part: end repeat. 
The second field(the date one) should not always be positioned below the first field, should only slide below if theres no space. As you can see on the Sebastian screen, the fields are still above each other even though theres space, it should be next to each other.

What you can do is, have a set width for the fields, and wrap them in a flex box to move below if theres no space.

GretaD:
> > The second part: end repeat. The second field(the date one) should not always be positioned below the first field, should only slide below if theres no space. As you can see on the Sebastian's screen, the fields are still above each other even though theres space. It should be next to each other.
> > What you can do is, have a set width for the fields, and wrap them in a flex box to move below if theres no space.
> 
> Yes, I understand what you mean. Originally I did it that way but I thought that the issue always needed them to be on seperate lines 👍

no problem, try this new approach when you have the time

keremsemiz:
I've went over the issues again, here is the revised version:
image

SebastianKrupinski:
Hey, tested again the counts is now properly in line in all screen sizes.

But the On Date still shows on two lines.

image

keremsemiz:
Thank you for pointint it out, I'll fix that tomorrow

SebastianKrupinski:
Tested. Again. Looks good to me, probably the best we going to get.

On Date
image

On Count
image

SebastianKrupinski (approved):

SebastianKrupinski:
@keremsemiz can you squash all the commits in to one please

keremsemiz:
yep, did it.

PR #6559, opened 2024-12-10
chore: change behaviour to behavior

kesselb:
To use the same spelling for expected and actual.


SebastianKrupinski (approved):
What's wrong with the Canadian spelling of Behaviour? 🤣 I mean if we do this for Behaviour, we are going to have to do it for Colour and Favourite 🤣

kesselb:
I don't have a strong opinion here, just found it weird to have "expected behavior" and "actual behaviour".

PR #6580, opened 2024-12-18
fix: avatar status icon text misalignment

GretaD:
before
Screenshot from 2024-12-18 17-51-04

after
Screenshot from 2024-12-18 18-14-13

### DIFF
--- a/src/components/Editor/AvatarParticipationStatus.vue
+++ b/src/components/Editor/AvatarParticipationStatus.vue
@@ -176,8 +176,8 @@ export default {
 </script>
 <style lang="scss" scoped>
 :deep(.avatar-participation-status__indicator) {
-	bottom: 20px;
-	left: 43px;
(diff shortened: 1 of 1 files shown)


ChristophWurst (dismissed):
This doesn't restore how it looked.

The status icon was part of the subline. This is also the reason why the subline is indented.

ChristophWurst:
Check out https://docs.nextcloud.com/server/29/user_manual/en/groupware/calendar.html#invite-attendees-to-an-event for what the view is intended to look like.

GretaD:
> This doesn't restore how it looked.
> 
> The status icon was part of the subline. This is also the reason why the subline is indented.

ok, changed it to how it looked. It was confusing what was supposed to be done, because the initial screenshot you sent its not reproducible.

ChristophWurst:
The main line (attendee name) and the icon in the subline do not align

PR #6596, opened 2024-12-30
feat: OCC and OCS Calendar Import/Export

SebastianKrupinski:
* Requires: https://github.com/nextcloud/server/pull/49995

## Summary
This adds the ability to export calendars via the OCS and OCC.

### OCC Export
Command: calendar:export
Arguments: userId calendarId format filepath

### OCC Import
Command: calendar:import
Arguments: userId calendarId format filepath
Options: errors, validation, supersede, show-created, show-updated, show-skipped, show-errors

### OCS Export
Endpoint: /ocs/v2.php/calendar/export
Request: GET/POST
{
    "id": "personal",
    "fmt": "xcal", (optional "ical, jcal, xcal", defaults to "ical")
	"user": "user1" (optional admin permissions required)
}

## Performance
- Export performance bottle neck seems to be CPU core max out due to single process execution, tests on dev setup with 16k of events takes about 5min or 60 items / second
- Import performance bottle neck seems to be database write performance,  tests on dev setup with 16k of events, takes about 30min or 10 items / second

## TODO

- [ ] Tests


tcitworld:
I think it feels odd to have this in the calendar as other calendar commands are into the `dav` namespace.

lib/Service/Export/ExportService.php
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	/**
+	 * Generates appropriate output start based on selected format
+	 */
+	private function exportStart(string $format): string {
+		return match ($format) {
+			'jcal' => '["vcalendar",[["version",{},"text","2.0"],["prodid",{},"text","-\/\/IDN nextcloud.com\/\/Calendar App\/\/EN"]],[',
+			'xcal' => '<?xml version="1.0" encoding="UTF-8"?><icalendar xmlns="urn:ietf:params:xml:ns:icalendar-2.0"><vcalendar><properties><version><text>2.0</text></version><prodid><text>-//IDN nextcloud.com//Calendar App//EN</text></prodid></properties><components>',
+			default => "BEGIN:VCALENDAR\nVERSION:2.0\nPRODID:-//IDN nextcloud.com//Calendar App//EN\n"
```
tcitworld, line 70: I don't like hardcoding things here. Would it be possible to serialize an empty calendar to get the start string?

SebastianKrupinski: I agree, I don't like hard coding this either. But I could not think of a better way to do this.

Yes, I could serialize a empty object but then I would still have to do fancy string manipulations to separate the beginning form the closing. Then output the beginning, then the objects, then the closing, as the output is a generated stream directly to the http buffer.

The only way to do this without hard coding would be to generate the calendar in memory and serialize it all as one. So it was a trade of hard coding or php memory exhaustion when exporting large calendars.

SebastianKrupinski:
All code moved to server

PR #6613, opened 2025-01-08
fix: max-width for long timezones names

### LINKED ISSUE #6570: New Appointment Schedule view breaks with some selected time zones
### Steps to reproduce

1. Create appointment schedule 
2. Got to appointment schedule page

### Expected behavior

View should remain consistant

### Actual behavior

Some time zones are fine
Image

Time zones with long names break the view
Image

GretaD:
fixes  #6570

to avoid any breaking changes on other parts where ncselect is used, lets force it on calendar

Screenshot from 2025-01-08 12-39-08

### DIFF
--- a/src/views/Appointments/Booking.vue
+++ b/src/views/Appointments/Booking.vue
@@ -287,6 +287,10 @@ export default {
 .booking__time-zone {
 	margin-top: 280px;
 	position: relative;
+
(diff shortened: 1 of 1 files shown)


tcitworld (commented):
I'd scope it into `.booking__time-zone` to make sure it doesn't affect anything else that would be added afterwards.

PR #6623, opened 2025-01-13
feat: Add overlay between calendar and open simple editor

### LINKED ISSUE #6582: Add overlay between calendar and open simple editor

As a Nextcloud user I don't want other calendar elements to be clickable while the simple editor is open.

E.g.
1. Open my calendar with existing events
2. Click on event a -> see the simple editor for a
3. Click on event b -> see the simple editor for b

1. Open my calendar with existing events
2. Click on event a -> see the simple editor for a
3. I'm not able to click anything outside of the simple editor within the view of the calendar. This is similar to modals.

https://github.com/nextcloud/calendar/issues/5513

### Additional context

N/a

GretaD:
the editing state has now a overlay, similar to modal or dialog. It closes if you click the close button on the popover. The viewing state stays the same.
fixes #6582

before
Screenshot from 2025-01-13 17-24-04
after
Screenshot from 2025-01-13 17-31-12

### DIFF
--- a/src/views/EditSimple.vue
+++ b/src/views/EditSimple.vue
@@ -4,166 +4,178 @@
 -->
 
 <template>
-	<NcPopover ref="popover"
-		:shown="showPopover"
-		:auto-hide="false"
-		:placement="placement"
-		:boundary="boundaryElement"
-		popover-base-class="event-popover"
-		:triggers="[]">
-		<template #trigger="{ attrs }">
-			<!-- Dummy slot to silence vue warning regarding a custom trigger -->
-			<button v-bind="attrs" style="display: none" />
-		</template>
-		<div class="event-popover__inner">
-			<template v-if="isLoading && !isSaving">
-				<PopoverLoadingIndicator />
+	<div>
+		<div v-if="showPopover && !isViewing"
+			ref="mask"
+			class="modal-mask"
+			:class="{
+				'modal-mask--opaque': dark,
+				'modal-mask--light': lightBackdrop,
+			}"
+			role="dialog"
+			aria-modal="true"
+			tabindex="-1" />
+		<NcPopover ref="popover"
+			:shown="showPopover"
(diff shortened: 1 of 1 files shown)


src/views/EditSimple.vue
```diff
@@ -416,10 +448,28 @@ export default {
...
+<style lang="scss" scoped>
 .event-popover__inner {
 	width: unset !important;
 	max-width: unset !important;
 	min-width: 300px !important;
 }
+.modal-mask {
+	position: fixed;
+	z-index: 9998;
+	//the height of header
```
ChristophWurst, line 460: Please test this with an embedded calendar

GretaD: sure, how can i add/create an embed calender?

GretaD: looks good. Why did you specify this, did you expect to be broken? (Just to make sure i didnt miss something.)

ChristophWurst: Because of the specific value used to position the mask relative to the header. If the header sizes are not identical, this solution wouldn't work.
Thank you for checking.

src/views/EditSimple.vue
```diff
@@ -334,6 +358,14 @@ export default {
 		window.removeEventListener('keydown', this.keyboardDuplicateEvent)
 	},
 	methods: {
+		closePopover() {
+			this.showMask = false
+		},
+		cssVariables() {
+			return {
+				'--slideshow-duration': '300ms',
```
ChristophWurst, line 366: is this used?

GretaD: no, leftover of hasPreview and hasNext button, from NcModal that we dont need here :). Removed

ChristophWurst (approved):
Code looks good as https://github.com/nextcloud/calendar/pull/6623/files?w=1

PR #6624, opened 2025-01-13
Visually distinguish invitations in calendar grid

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/3869

swappy-20250121_193855

Based on https://github.com/nextcloud/calendar/issues/3869#issuecomment-1026921383

### DIFF
--- a/src/fullcalendar/eventSources/eventSourceFunction.js
+++ b/src/fullcalendar/eventSources/eventSourceFunction.js
@@ -11,7 +11,7 @@ import {
 } from '../../utils/color.js'
 import logger from '../../utils/logger.js'
 import { getAllObjectsInTimeRange } from '../../utils/calendarObject.js'
-
+import usePrincipalsStore from '../../store/principals.js'
 /**
  * convert an array of calendar-objects to events
  *
@@ -23,6 +23,8 @@ import { getAllObjectsInTimeRange } from '../../utils/calendarObject.js'
  * @return {object}[]
  */
 export function eventSourceFunction(calendarObjects, calendar, start, end, timezone) {
+	const principalsStore = usePrincipalsStore()
+
 	const fcEvents = []
 	for (const calendarObject of calendarObjects) {
 		let allObjectsInTimeRange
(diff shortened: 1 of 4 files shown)


src/fullcalendar/eventSources/eventSourceFunction.js
```diff
@@ -32,10 +32,30 @@ export function eventSourceFunction(calendarObjects, calendar, start, end, timez
...
+									didEveryoneDecline = false;
+								}
+							}
+						})
+					}
+				}
+				if (didEveryoneDecline) {
+					classNames.push('fc-event-nc-all-declined')
+				}
+			}
```
st3iny, line 57: The data you need is already exposed by calendar-js. Take a look at the following pseudo-code:

```js
for (const attendeeProperty of object.getPropertyIterator('ATTENDEE')) {
	const hasDeclined = attendeeProperty.participationStatus === 'DECLINED'
	// TODO: write the rest of the code
}
```

GVodyanov: ❤️

nimishavijay:
Looking great! Could you also post a screenshot in light mode, and one with the entire screen? to check the contrast. 
So far I have only one suggestion, which is to use the filled [warning symbol](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:warning:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=warning&icon.size=24&icon.color=%23222222) instead of the outlined one so that it's visible more clearly.

GVodyanov:
> Looking great! Could you also post a screenshot in light mode, and one with the entire screen? to check the contrast. 

Here you go @nimishavijay 

image

The outlined yellow event on a white background is the worst we can get in terms of contrast (at least out of the default colors)... I agree it's not great, I guess the spec could be changed to either modify the text and have it be the normal text color, or leave the background and have some other way of showing it's declined, what do you think is best?

nimishavijay:
Yep, exactly what I thought too. 
- [ ] We can definitely bold the text for full day events, considering the regular events are bold too. 
- [ ] To be on the safer side I'd also agree that we can make it the normal text color. 
- [ ] Other than that I can only think of maybe using a thicker border (maybe 2px?). We can try that and see, let me know what you think. 

- [ ] Also unrelated nitpick but I noticed that today's events are like 1px too low (possibly because of the box around the date). Is it possible to fix that?

GVodyanov:
First things first, here's the new icon

GVodyanov:
How is this looking? @nimishavijay 

image

image

SebastianKrupinski:
Looking good! 👍

nimishavijay:
Awesome! Thanks for the comprehensive screenshots. 2px works! Looks much clearer now. Couple of more nitpicks:

- [ ] I see that the height of the needs-action and confirmed full day events are different (needs-action is slightly more because of the border). Could we make this consistent? Maybe by using outline instead of border or decreasing the height and adding a border to the confirmed event in the same colour, you get the idea 
- [ ] The outlined circle stroke also seems very thin, possible to make that also 2px? (Also make sure to keep the size consistent with the filled circle like the previous point)
- [ ] add 2px spacing between the warning icon and the text, rn it seems very close 

Other than that it is ready to go! Super nice work! :)

nimishavijay (approved):
Awesome! Looks perfect now!

src/fullcalendar/eventSources/eventSourceFunction.js
```diff
@@ -32,16 +34,46 @@ export function eventSourceFunction(calendarObjects, calendar, start, end, timez
...
+					}
+					if (didEveryoneDecline) {
+						classNames.push('fc-event-nc-all-declined')
+					}
+				}
+				if (object.status === 'CANCELLED') {
+					classNames.push('fc-event-nc-cancelled')
+				} else if (object.status === 'TENTATIVE') {
+					classNames.push('fc-event-nc-tentative')
+				}
```
st3iny, line 60: This is a regression. The cancelled and tentative states are global for the event and not tied to organized event. You can set the status on events that have no organizer and attendees.

I think you should move the block outside of the `if ('ORGANIZER')` again.

The frontend tests also fail because of this.

st3iny: Can be reproduced by creating an event and setting the status to cancelled.

**Before:** Background is slightly transparent and the title is crossed.
**Here:** Event is shown as a regular without any styling changes.

st3iny (changes requested):
Left some feedback. Eslint is not yet happy.

src/fullcalendar/eventSources/eventSourceFunction.js
GVodyanov: Thanks for noticing!

GVodyanov:
The tests also seem to be failing because of the usage of Pinia outside of a vue component, I think that the tests have to be modified, but I will check if that is the best way to solve this.

st3iny (approved):
Tested and works. Great job, this is a neat improvement!

SebastianKrupinski:
@GVodyanov 

How difficult would it be to add the highlighting recommendation to this PR? from this ticket.

https://github.com/nextcloud/calendar/issues/6708

SebastianKrupinski:
We NVM, @st3iny merged it as I posted the comment 😆

PR #6626, opened 2025-01-14
fix: simple calendar view width

### LINKED ISSUE #6625: Location and description with long text, in viewing mode doesn wrap
### Steps to reproduce

1. Go to calendar
2. Create and event or open an existing one with long locationa nd description
3. See the event after saving

### Expected behavior

Good design

### Actual behavior

Image

### Calendar app version

latest

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

_No response_

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

_No response_

GretaD:
fixes #6625 

Screenshot from 2025-01-14 11-03-09

### DIFF
--- a/src/views/EditSimple.vue
+++ b/src/views/EditSimple.vue
@@ -419,7 +419,6 @@ export default {
 <style>
 .event-popover__inner {
 	width: unset !important;
-	max-width: unset !important;
 	min-width: 300px !important;
 }
 </style>


GretaD:
@GVodyanov you changed the max-width to unset. Do you remember why you did that? Also the reason why its smaller now, its because of the min-width, can we have the min-width bigger?

GVodyanov:
> @GVodyanov you changed the max-width to unset. Do you remember why you did that? Also the reason why its smaller now, its because of the min-width you added in the same class, can we have the min-width bigger?

Hey! Yeah this was a result of the timepicker change, Nimisha asked to have the modal be a dynamic width
 https://github.com/nextcloud/calendar/pull/6423#issuecomment-2474743608 
https://github.com/nextcloud/calendar/issues/6385#issuecomment-2401043906

We can certainly change the sizes if you want, just wanted to show why we have it the way it is now. Maybe it would be better to ask Nimisha before changing though.

GretaD:
> > @GVodyanov you changed the max-width to unset. Do you remember why you did that? Also the reason why its smaller now, its because of the min-width you added in the same class, can we have the min-width bigger?
> 
> Hey! Yeah this was a result of the timepicker change, Nimisha asked to have the modal be a dynamic width [#6423 (comment)](https://github.com/nextcloud/calendar/pull/6423#issuecomment-2474743608) [#6385 (comment)](https://github.com/nextcloud/calendar/issues/6385#issuecomment-2401043906)
> 
> We can certainly change the sizes if you want, just wanted to show why we have it the way it is now. Maybe it would be better to ask Nimisha before changing though.

yes, but that broke those field so, we should have a better solution for dynamic fields.

GVodyanov:
Sure that sounds like a good idea

ChristophWurst:
Is this ready for testing and review?

GretaD:
> Is this ready for testing and review?

yes, if we only want to fix the width. But, i opened another ticket to improve simple editor for other issues too.

st3iny (approved):
Tested and works.

GretaD:
i would say lets merge this one and then work on the rest whenever we pickup the overview ticket, since there are too many people who noticed the issue.

PR #6639, opened 2025-01-21
Fix: avoid hotkeys in contenteditable

janepie:
fixes https://github.com/nextcloud/assistant/issues/16

[NcRichContenteditable](https://nextcloud-vue-components.netlify.app/#/Components/NcRichContenteditable) uses a `div` for text input so shortkeys defined via `vue-shortkey` do not get avoided unless they are explicitly excluded


st3iny (approved):
Looks good, thanks.

Issue #6660, opened 2025-01-24
Closing the free/busy modal doesn't update time in sidebar


GVodyanov:
### Steps to reproduce

1. ![Image](https://github.com/user-attachments/assets/e95ab156-0f78-4b67-bc1d-2536236718b8)

### Expected behavior

The time should change in the sidebar to the one selected in the modal

### Actual behavior

The time doesn't change

### Calendar app version

main

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

PR #6690, opened 2025-02-02
fix: free busy not updating date

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6660

I think this comes from an unfound bug in the Pinia migration


SebastianKrupinski (approved):

st3iny (approved):
Did not test but looks good.

PR #6706, opened 2025-02-11
fix: force height for descr and location

GretaD:
the problem
Screenshot from 2025-02-11 16-21-20

fixed

Screenshot from 2025-02-11 16-19-23

### DIFF
--- a/src/views/EditSidebar.vue
+++ b/src/views/EditSidebar.vue
@@ -109,6 +109,7 @@
 				:is-read-only="isReadOnly"
 				:prop-model="rfcProps.description"
 				:value="description"
+				:is-description="true"
 				:linkify-links="true"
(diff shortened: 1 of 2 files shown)


src/views/EditSidebar.vue
```diff
@@ -109,6 +109,7 @@
 				:is-read-only="isReadOnly"
 				:prop-model="rfcProps.description"
 				:value="description"
+				:is-description="true"
```
GretaD, line 112: this change is needed, because when you click the textarea it jumbs to a smaller side, and i fixed it some time ago for simple editor and forgot to add it here

ChristophWurst (approved):
Looks good but didn't test

SebastianKrupinski (approved):

Issue #6711, opened 2025-02-12
Changing event date resets time to midnight


ChristophWurst:
### Steps to reproduce

1. Create an event at 10AM tomorrow
2. Save the event
3. Open the event sidebar
4. Change the date to the day after tomorrow

### Expected behavior

Time stays at 10AM

### Actual behavior

Time is reset to 12AM (midnight)

### Calendar app version

5.1.0-rc2

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

PR #6725, opened 2025-02-16
Fix date selector resetting time

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6711


SebastianKrupinski (approved):

ChristophWurst:
@GVodyanov @SebastianKrupinski this needs backports to all effected stable branches

GVodyanov:
There's another issue here with the end time changing when it shouldn't, will fix

PR #6768, opened 2025-02-25
feat(talkintegration): add object type to talk room creation

miaulalala:
For https://github.com/nextcloud/spreed/issues/14401 we need an object type for the calendar- created Talk rooms.

This PR adds the object type "event" and passes is as an `objectType` when creating a talk room. Since there is only one room type created directly from the calendar, it should be ok to have this as a static value (same as the room type always being `3`).

@st3iny is there any chance to also pass the UUID for the `objectId`? I'm guessing it doesn't exists yet at this point in the process?

### DIFF
--- a/src/services/talkService.js
+++ b/src/services/talkService.js
@@ -9,6 +9,7 @@ import { loadState } from '@nextcloud/initial-state'
 import { getCurrentUser } from '@nextcloud/auth'
 import logger from '../utils/logger.js'
 import { removeMailtoPrefix } from '../utils/attendee.js'
+import md5 from 'md5'
 
 /**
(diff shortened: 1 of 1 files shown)


SebastianKrupinski (approved):

SebastianKrupinski:
@miaulalala  Hey Hey! Long time!
 
> @st3iny is there any chance to also pass the UUID for the `objectId`? I'm guessing it doesn't exists yet at this point in the process?

The UI creates the ID and UID fo the object so an I'd would be possible. We would just need to check if both of those are created at button click or creation of the new object. Either way it is possible.

SebastianKrupinski (approved):

Issue #6424, opened 2024-10-15
there should be space/gap next to events, so you can create new events at the same time as existing events


alexanderdd:
### Steps to reproduce

1. create an event
2. try to create a new event at the same time by dragging mouse on desired time

### Expected behavior

there should be space at the right side next to events where you can click and drag.

It was still there in 29.0.6, calendar 4.7.16

![Image](https://github.com/user-attachments/assets/2b6e7a3f-2da6-42ae-b83b-65a611bc54da)

### Actual behaviour

Starting in NC 30 (c.nc.com) there is no space, not possible to create event via click and drag.

### Calendar app version

c.nc.com

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

PR #6775, opened 2025-02-28
style: add a gap for event dragging

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6424

| Before | After |
| - | - |
| image | image |


ChristophWurst:
:ping_pong: @GVodyanov this seems ready except for the conventional commit message

Issue #6701, opened 2025-02-05
Changes by attendee propagating to organiser


hamza221:
### Steps to reproduce

1. Create event as user1
2. Invite User 2
3. Login as User 2
4. Change date 
5. Accept event

### Expected behavior

User1 get a notification that only mentions the attendance status

### Actual behavior

<img width="688" alt="Image" src="https://github.com/user-attachments/assets/7b955bdb-5f67-45fd-a67c-6e6c068a9b22" />

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

PR #6778, opened 2025-03-02
fix: restrict attendees edit priveleges in the frontend

hamza221:
Don't allow attendee to edit 
- dates
- Title
- Attendees
- Resources 

to avoid running into  #6701

### DIFF
--- a/src/fullcalendar/interaction/eventDrop.js
+++ b/src/fullcalendar/interaction/eventDrop.js
@@ -8,7 +8,10 @@ import logger from '../../utils/logger.js'
 import { getObjectAtRecurrenceId } from '../../utils/calendarObject.js'
 import useCalendarsStore from '../../store/calendars.js'
 import useCalendarObjectsStore from '../../store/calendarObjects.js'
-
+import usePrincipalsStore from '../../store/principals.js'
+import { isOrganizer } from '../../utils/attendee.js'
+import { showWarning } from '@nextcloud/dialogs'
+import { translate as t } from '@nextcloud/l10n'
 /**
  * Returns a function to drop an event at a different position
  *
@@ -18,6 +21,7 @@ import useCalendarObjectsStore from '../../store/calendarObjects.js'
 export default function(fcAPI) {
 	const calendarsStore = useCalendarsStore()
 	const calendarObjectsStore = useCalendarObjectsStore()
+	const principalsStore = usePrincipalsStore()
 
 	return async function({ event, delta, revert }) {
 		const deltaDuration = getDurationValueFromFullCalendarDuration(delta)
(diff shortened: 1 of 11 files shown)


SebastianKrupinski:
Tested. Works. Looks good.

Modal:
image

Side bar:
image

SebastianKrupinski (approved):

src/fullcalendar/interaction/eventDrop.js
```diff
@@ -56,6 +59,12 @@ export default function(fcAPI) {
 			return
 		}
 
+		if (!isOrganizer(principalsStore.getCurrentUserPrincipalEmail, eventComponent.organizer)) {
+			revert()
+			showWarning(t('calendar', 'You are not allowed to edit this event as an attendee.'))
+			return
+		}
```
st3iny, line 66: How would I trigger this warning? Given that most fields are read-only.

st3iny (changes requested):
The `Add talk conversation` button needs to be hidden as well.

image

src/fullcalendar/interaction/eventDrop.js
st3iny: Nvm, I can't read. Dragging and dropping will trigger the warning.

st3iny:
Would it be possible to move the help text to the right of the calendar picker? Also, it is missing in the simple editor (edit mode).

hamza221:
> Would it be possible to move the help text to the right of the calendar picker? Also, it is missing in the simple editor (edit mode).

I tried quickly yesterday, and failed since we're using [NcAppSidebar](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppSidebar?id=ncappsidebar-1)   I'll give it another try now

SebastianKrupinski:
> The `Add talk conversation` button needs to be hidden as well.

Good catch! I don't have the talk app enabled!

hamza221:
> Would it be possible to move the help text to the right of the calendar picker? Also, it is missing in the simple editor (edit mode).


is this what you meant by simple editor edit mode? it's already there 
image

hamza221:
> > Would it be possible to move the help text to the right of the calendar picker? Also, it is missing in the simple editor (edit mode).
> 
> I tried quickly yesterday, and failed since we're using [NcAppSidebar](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppSidebar?id=ncappsidebar-1) I'll give it another try now

I misunderstood your comment earlier, I thought you meant next to the close button like the screenshot above. That's definitely possible, I went with left in case of a long calendar name

st3iny (approved):
Tested and works.

Pushed a small commit to fix the calendar picker overlapping with the warning in case a calendar with a very long name is picked.

makuser:
While this does indeed fix #6701, it does not take into account, if the calendar is shared with write permission, which should then (also as attendee) allow to rightfully modify the event.
This broke our workflow and so far, I did not find any other configurable way around this.

PR #6794, opened 2025-03-09
fix: do not show attendee actions in viewing mode

SebastianKrupinski:
Resolves: Issue with attendee action showing in popover viewing mode. The action also had no affect since the modal was only in viewing mode.

Before:
Screenshot 2025-03-09 102519

After:
Screenshot 2025-03-09 105310


ChristophWurst (approved):
Nice!

PR #6795, opened 2025-03-09
fix: do not show attendee list when there are no attendees in viewing mode

SebastianKrupinski:
Resolves: Issue with attendee list showing blank place holder when there are no attendees. We do not show blank place holders for location or description, attendees list should function the same way, less in more.

Before:
image

After - Viewing Mode - No attendees
Screenshot 2025-03-09 111552

After - Editing Mode - No attendees
Screenshot 2025-03-09 112142

After - Viewing Mode - With attendees
Screenshot 2025-03-09 112058

After - Editing Mode - With attendees
Screenshot 2025-03-09 112109

### DIFF
--- a/src/views/EditSimple.vue
+++ b/src/views/EditSimple.vue
@@ -148,7 +148,8 @@
 						:is-description="true"
 						@update:value="updateDescription" />
 
-					<InviteesList class="event-popover__invitees"
+					<InviteesList v-if="!isViewing || (isViewing && hasAttendees)"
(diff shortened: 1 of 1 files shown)


GVodyanov (approved):
I like this

hamza221 (approved):
Let's wait for @nimishavijay  before merging

nimishavijay (approved):
Looking great 👌🏼

PR #6797, opened 2025-03-09
fix: yearly recurrance options - month selection

SebastianKrupinski:
Resolves: Issue with month selection not showing selected months

Before:
image

After:
image


GVodyanov (approved):
I wasn't able to reproduce the original issue actually but code looks and works fine

PR #6801, opened 2025-03-10
[stable5.0] fix: do not show attendee actions in viewing mode

SebastianKrupinski:
Manual backport of #6794


ChristophWurst:
5.1 supersedes 5.0. Like, all instances that support 5.0.x will update to 5.1. Thus we don't release stable5.0 anymore.

https://apps.nextcloud.com/apps/calendar

5.1 and 4.7 are currently maintained.

PR #6827, opened 2025-03-16
Fix calendar grid transparency overlapping issues

GVodyanov:
| Before | After |
| - | - |
| image | image |


ChristophWurst:
423531483-dfc14c2f-a79d-45c7-83af-ae21ba11f3d5

these are still weird. IMO the transparent events should stay as transparent as non overlapping events.

GVodyanov:
Well, I made it work with some pretty weird geometry. I hope it isn't too hacky.

image

PR #6829, opened 2025-03-17
Make time text color in calendar grid be main text color

GVodyanov:
| Before | After |
| - | - |
| image | image |


st3iny (approved):
LGTM!

Issue #6521, opened 2024-11-25
Disabled days in appointment preview calendar look bad (dark mode)


GVodyanov:
### Steps to reproduce

1. Create an appointment
2. Preview it
3. Click on "Select date"
4. See this: ![Image](https://github.com/user-attachments/assets/ce24368a-e47c-4d1a-8e9b-b75217ddb430)

### Expected behavior

The days not belonging to the current month should also be disabled, and the fact that they are disabled should be communicated better. Currently IMO they just look broken.

### Actual behaviour

Currently IMO they just look broken.

### Calendar app version

main

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

PR #6845, opened 2025-03-20
Booking: change color of disabled days in datepicker

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6521

| Before | After |
| - | - |
| image | image |
| image | image |


nimishavijay (dismissed):
Definitely looks great in dark mode! The disabled dates stand out more than the available dates in light mode though. Could we try using background-dark instead of background-darker, and change the text color of the available dates to main-text? That could help make them stand out more.

st3iny (approved):
Tested and works.

Issue #5513, opened 2023-10-02
Full screen event detail view


drankinatty:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

It is frustrating when viewing a calendar entry in a pop-up to select "more" and have the pop-up embed itself in the right side of the browser calendar display. This reduces the width of the information display, cuts off information, and wraps any descriptive text greater than 32 characters (or so). For example, when clicking an entry, a nice pop-up appears:

![nextcloud-cal-event](https://github.com/nextcloud/server/assets/9997638/7ec0c07b-0d04-43b3-a68f-2f84ab059b74)

It could be a little wider, but that's likely browser dependent. However, when you click more, the pop-up disappears and the entire entry is embedded in the right side of the calendar. In the format of a slide out that w3schools was providing as a "hot" example a couple of years ago. E.g.

![nextcloud-cal-event-more](https://github.com/nextcloud/server/assets/9997638/402bf16c-1fb1-4e82-b546-141965f360da)

Here, the description text is wrapped and jumbled as well as the reminder information. The irony is you actually see *"Less"* information after pressing *"More"* and you then must use a scrollbar in the reduced with right-side display anyway to see any additional calendar event information. What's the point?? 

**Describe the solution you'd like**

Just add a scrollbar (auto-scroll) in the existing pop-up. Nothing could be simpler, and that would preserve the dialog width and allow the descriptive text and controls to display without wrapping. Leave the trendy JS slide-out to the w3scools example pages. KISS is always the best policy.

**Describe alternatives you've considered**

There really are not a lot of alternatives. The one obvious solution is by far the simplest to implement and would save Nextcloud 20-50 lines of code. Just use the existing pop-up...

**Additional context**

This really goes to readability and proper display of information. Trying to wedge the information in a slide-out on the right side, leaves the calendar event without enough room to display properly and at the same time compacts the actual calendar display to the point other events are unreadable.

miaulalala:
We're currently working on a design improvement for the Calendar, so this might be solved soon:

https://github.com/nextcloud/calendar/issues/3543

drankinatty:
[...]

PR #6864, opened 2025-03-28
Add full page event editor

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/5513

Here are some screenshots:
image

image

When scrolling down the top part remains anchored:
image


SebastianKrupinski:
Hi, 

Spent some time reviewing and testing this, functionality wise, all functions seem to work.

- Tested Singleton Events
- Tested Recurring (Day, Week, Month, Yearly)
- Tested Creating, Updating, Deleting

Visually I have some recommendations.

### In view only mode (as attendee) 
- Tentative button does not show properly
- Event status (top right column) should be disabled attendee should not be able to cancel the event
- Resources separator still shown, in right column
- No attendee place holder still showing in attendee list 
 
Screenshot 2025-05-13 191051

### In read/write mode (as organizer)
- Use transparent overlay to show the calendar in the background on the sides at the moment there is a lot of gray space which makes it feel off
- Change the X to button and place it beside the save button
- Make buttons always visible (don't hide after scrolling to the bottom)
- Repeat explanation squeezes the all day option
- Split screen in to columns, adding alarms or attachments unnecessarily moves the resource section down

image

### In repeat modal
- Move X on repeat modal to in line with title the extra spacing makes it look off. 
- Inline Repeat interval/frequency, relative position, and end
- Remove wording times when end after is selected

image

SebastianKrupinski:
Minus some minor changes it looks amazing!

nimishavijay (changes requested):
Super cool work! :) Discussed this in the design team call and everyone loved it as well, here are the comments:

- [x] The spacing between the 2 big columns can be reduced to something like 12 or 16px. The entire content should still be centered. We can have the following logic:
  - [x] Width of the content is limited to 700px 
  - [x] Gap between 2 columns is 16px
  - [x] First column is (2/3 * max width) - gap/2 = 458px
  - [x] Second column is (1/3 * max width) - gap/2  = 225px
  - [x] If the widths look really off we can just adjust the ratio (instead of 2/3 and 1/3 we can try 3/5 and 2/5 or something)
- [ ] The width of the location, description, etc fields are a bit lesser than the name, start/end time etc. They should all be left and right aligned
- [ ] The export, duplicate and delete seem very present as the first actions we see on the screen. Maybe we can move them into a 3 dot menu next to the update button
- [ ] The update button and the 3 dot menu can go next to the name, the the calendar button (` 🔵Personal`) can go on the last row next to the repeat button
- [ ] We also need an "Add Talk room" button, and that can be on the right of the location field 

Small adjustments for each field:
- [ ] Tags: it can go to the right side, above the color field
- [ ] Reminders: Remove the duplicated icon for each reminder
- [ ] Attendees:
  - [ ] Move "Find a time" to next to the attendees heading so that it doesn't get buried under a long list of attendees
  - [ ] Adjusted the width of the attendees search field so that it is left anf right aligned with the rest of the fields
- [ ] Resources: Remove the "Suggested" heading 

Very janky mockup for how it could look after moving everything
image


Component-level fixes possibly
- [ ] The borders for the different fields are all different thicknesses and colours. They should all be the same thickness (1px) and color (--color-border-maxcontrast)

I also have some notes for improving the rooms and resources and other fields like tags, reminder, etc. But that can be a follow up. 

Looks awesome! Can't wait to see it in our calendar :)

nimishavijay (changes requested):
Looks amazing! :) Thanks for incorporating all the changes!

> About the reminder icons (the little bells), the idea is to not have duplicated items, but I see you kept the icon for the "reminder entry" and not the "reminder selector". Seeing as you can have multiple reminder entries, that would result in having multiple icons. May I suggest having it for the selector and not for the entries? I think that makes that brings it closer to the objective of not having repetitions.

Makes sense, in that case we can move the selector to the top as the first item. I have a couple more changes for reminders in mind which we can also do in a follow up :) 

> Also, I see that the "All day" checkbox is all the way to the left: that's a little inconvenient as it requires overriding the NcCheckboxRadioSwitch styling. It's not a problem to override something if necessary, but maybe the default styling looks good too with the new alignment (in the interest of sticking with Nextcloud vue)?

No problem, not an issue. Since the checkbox itself is nicely aligned with the rest of the elements, it looks alright :) 

> A small note is that I used the measurements that you suggested, so 700px total and 16px gap. However, what came out looks a tad bit smaller than your mockup. Would you like me to keep those measurements or to make them a little larger to be more similar to the mockup? It isn't hard to adjust, I set variables for everything.

Indeed things seem a bit smaller. we could increase it to 900px maybe? The main thing I'm looking for is that no text is ellipsized. 

Other small points I noticed:
- [x] The date/time fields are ever so slightly misaligned, they should be aligned 
image
- [x] If possible, could we add maybe 4px (`--default-grid-baseline`) on the left of the date picker? The "From" and "To" labels seem very close to the fields.
- [x] The icon for the timezone picker seems to have shifted to the right in the latest screenshot?
- [x] Initially I thought the name and the time can stay sticky, but looking at the screenshot right now, it seems like that's around 200px that's constantly taking up screens space (and that's not even counting the 50px of the app bar). Could we instead keep only the title+update button sticky? :) 
- [x] As a result of that we can remove the horizontal separator line and reduce the space above and below the "All day/does not repeat" row to the same as the space between the other fields
- [x] I can't tell from looking at the screenshot, but is there a 8px gap on the edges, like there is for the content normally?

GVodyanov:
Hey @nimishavijay! I adjusted to your latest comment, but I have a small note:

swappy-20250525_183819

Anchoring the time input field to the same X coordinate as the title field means that the text label of the input field has to exist outside the main structure, and when you have a long text string it might start going off-screen. I solved this by wrapping soon. Please let me know if you have any comments on the mobile version too.

nimishavijay:
> Anchoring the time input field to the same X coordinate as the title field means that the text label of the input field has to exist outside the main structure, and when you have a long text string it might start going off-screen. I solved this by wrapping soon. 

Good point, yes we should wrap at some point, maybe after 200px (is that a lot idk)?

> Please let me know if you have any comments on the mobile version too.

Could you post screenshots of the latest mobile version?
Also of read only events with all the changes?

GVodyanov:
Hey! How does this look? I tried to have the responsive version have the same logic @nimishavijay 

image

image

image

image

nimishavijay:
Thank you! Looks nice on mobile as well :) 

Small changes on both desktop and mobile:
- [x] The time zone picker and the "Add Talk conversation" is missing in the latest screenshots?
- [x] Reminders:
  - [x] Could we move the "Add reminder" NcSelect and the bell icon to above the list of reminders?
  - [x] Also get rid of the "+" in the label on the NcSelect so that it's just "Add reminder"
  - [x] In each reminder, move the 3 dot menu to the end of that row such that they are all aligned with the "+" and the "x" of the attachments
- [x] The location field is also a bit taller than the rest of the input fields (I'm guessing it's also a component-level issue). Any way it can be overridden to be 34px like the rest of the input fields?
- [x] align the icon of description to the top of the description field, not the center

On mobile:
- [x] It's bothering me that there is no way to close the form if you scroll as the "x" button scrolls away and I'm not sure how to solve this. We can't keep it next to the "Save" button as there might be misclicks and then you've lost all the info you typed
- [ ] Could be kind of a radical change, but I'm thinking about removing the icons on the left on mobile. That way we have more screen space, and everything can be nicely aligned on the left. I'll check both of these points with the rest of the design team and confirm.

Read-only:
- [x] It's unclear which fields are editable? Is it possible for the recipient to add a description, reminder, change the status, etc?
- [x] Align "Does not repeat" icon and text to the date/time icon and text above
- [x] Don't show "Does not repeat" if it doesn't repeat
- [x] Align the start of the accept/decline row with the description input field, it looks like it's sticking out more than it should be right now

Follow ups (to do for Nim):
- [ ] Improve design of resources section 
- [ ] Improve placement of find a time and possibly link it to the external find a time feature
- [ ] Possibly change the design of the calendar picker. I'm thinking now that we have the space, we could move the calendar picker into an NcSelect and move it above "Confirmed"

GVodyanov:
Heyo!

>  The time zone picker and the "Add Talk conversation" is missing in the latest screenshots?

Hey they are actually still there I just forgot to include them in the screenshot. Both are conditionally displayed:
- "Add Talk conversation" when there is talk configured on the instance
- "Time zone picker" when the event isn't a full day event

>  It's bothering me that there is no way to close the form if you scroll as the "x" button scrolls away and I'm not sure how to solve this. We can't keep it next to the "Save" button as there might be misclicks and then you've lost all the info you typed

Misclicks won't be an issue soon ;) https://github.com/nextcloud/calendar/pull/6890

So what I did is make it so that on desktop you have the normal icon in the corner and this on mobile:
image

>  Could be kind of a radical change, but I'm thinking about removing the icons on the left on mobile. That way we have more screen space, and everything can be nicely aligned on the left. I'll check both of these points with the rest of the design team and confirm.

I'll wait for this then

>  It's unclear which fields are editable? Is it possible for the recipient to add a description, reminder, change the status, etc?

Yes all of that is possible

nimishavijay:
Notes from review:

- [x] Don't show resources in read only mode if it's empty
- [x] Fix jumping around when you open and close dropdown 

We will have some follow ups with for eg. better read-only mode, resources, and find a time but that's for later. This looks awesome! :)

GVodyanov:
Fixed tests, should be ready to go now!

EDIT Smh codecov... In theory, those components are covered by existing tests anyway

SebastianKrupinski:
Tested. Looks good and works as expected

SebastianKrupinski:
Looks like I spoke to soon! lol

Pressed add talk conversation and the location line went wacky...
image

GVodyanov:
> Looks like I spoke to soon! lol
> 
> Pressed add talk conversation and the location line went wacky... image

Fixed! Thanks for testing :)

GretaD:
The location field is still broken and i see 2 **_find time_** buttons.

Screenshot from 2025-07-03 14-29-30

GVodyanov:
@GretaD Finally figured out what was causing the location issue, should be all fixed now. Thanks for checking it out!

Issue #6881, opened 2025-04-03
End time resetting when changing date


GVodyanov:
### Steps to reproduce

1. Create an event from 1pm to 3pm
2. Set the day to the day after tomorrow

### Expected behavior

Times remain 1pm, 3pm

### Actual behavior

Times become 1pm, 1pm

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

PR #6888, opened 2025-04-06
Fix end time resetting when changing day

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/6881


SebastianKrupinski (approved):
Tested Works

PR #6920, opened 2025-04-17
fix(talkintegration): allow room creation with description

miaulalala:
Fixes https://github.com/nextcloud/spreed/issues/14886

Due to changes in the Talk API, rooms with type `event` cannot be modified after they're created.

To reproduce this issue on `main`, create a new event with a description, and only after writing the description, create a Talk room.

This PR aims to keep backwards compatibility for Nextcloud 30 and lets older versions still set the description if Talk supports it. 
Integrating with a newer Talk version ignores the error that will be returned when trying to set a description on a room type `event` and lets the user proceed. The description will be set in the Talk Event Listener instead.

The `description` API call can be removed as soon as Calendar min support is Nextcloud 31.

### DIFF
--- a/src/services/talkService.js
+++ b/src/services/talkService.js
@@ -22,27 +22,38 @@ import md5 from 'md5'
  */
 export async function createTalkRoom(eventTitle = null, eventDescription = null, attendees = []) {
 	const apiVersion = loadState('calendar', 'talk_api_version')
+	let response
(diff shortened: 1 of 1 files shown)


st3iny (approved):
Tested and works. Seems to fix the issue.

Thank you very much :)

PR #6936, opened 2025-04-25
Make event margin be in vw instead of percentage

GVodyanov:
| Before | After |
| - | - |
| image| swappy-20250425_171731 |


SebastianKrupinski (approved):

st3iny (approved):
LGTM!

Issue #25, opened 2016-09-17
improve tab index in editor


georgehrke:

ccoenen:
my main gripe is with the `all day event` checkbox, as it is in the wrong place in my opinion. If you're entering stuff via keyboard, currently you have to

* title <kbd>Tab</kbd>
* calendar selection <kbd>Tab</kbd>
* start date <kbd>Tab</kbd> end date <kbd>Tab</kbd>
* now at all day event <kbd>spacebar</kbd> (uncheck all day event)
* <kbd>Shift-Tab</kbd> <kbd>Shift-Tab</kbd> <kbd>Shift-Tab</kbd> (get back to start time)
...

georgehrke:
From a visual point of view I like the position of the all day button. we should just adjust the tab index :)

eppfel:
As mentioned in #72 I would not want to block users. Maybe you can tab to the time fields to enable them? And if you actually want daily events you hit space after tabbing through. Not sure...

PR #6959, opened 2025-05-06
feat: rework freebusy modal

### LINKED ISSUE #6611: Change view and colors of the find a time modal (free/busy)

hello there, 
every time I have to look for an available slot, I find my self very confused in front of the find a time modal...
- one need to horizontally scroll to browse the whole day, a full day view, or at least 9 to 5 would be great
- the colors are completely confusing me:
  - the colors in the legend look slightly different from the colors in the grid
  - usually the out of office is a rather dark color but in our case is a light color
  - usually the meetings are represented in a light color, but are here reprensented in a medium contrasted rather dark color
  - the available slots are usually the same color as the background to show as empty, but in our case it seems there is a grey

this really asks me a massive brain effort to try to understand this, and I always end a bit unsure whether or not the attendee is available.

screenshot of our find a time 
Image

screenshot of TB find a time
Image

Outlook
https://answers.microsoft.com/en-us/outlook_com/forum/all/send-availability-on-outlook-web-like-you-can-do/15dde2a7-cccf-419c-8307-2c46aecc0ffc
Image

- a 9 to 5 view of availability
- available slots should be same color as background (as if there were nothing, no event)
- I would prefer taken slots to be in a light color, as it is usually the case is your own calendar
- I would prefer out of office slots to be in a darker color
- make the legend colors exactly match the find a time grid colors

_No response_

### Additional context

_No response_

hamza221:
fix #6611
| web | Mobile |
|--------|--------|
|image|image |

- [x] Test potential regressions for resources 
- [x] Multiple Day event 
- [x] Wrong end date bug 
- [x] remove unused method 
- [x] Organize Mobile conditional rendering 
- [x] show own OOO as grey
- [x] RTL support
- [x] Cleanup code


nimishavijay (changes requested):
Looks great! Much more easy to understand! 

Addressing questions from the issue here: 

> So all attendees would be the same color ? and should the attendee's display name be the block title ?

So this was with the assumption that different colours was not possible. If it is possible then we can definitely do that!

I am only worried about the case when there are many people (eg 5+) where it becomes not very clear. In this case would it be possible to open the day view by default?

Other than that, some general feedback:
- [ ] We can add a border of 2px to the outside of each event block
- [x] We can also round the corners more (maybe 6px?)
- [ ] Can we theme the "week | day" switcher to look like NcButtons?
- [ ] Can we reduce the height of the "all day" row?
- [ ] If we do show a different color for each user, I'm wondering if there's a way to indicate that better. How about we show the list of users on the left vertically with a colored circle for each? This would be completely outside the fullcalendar component
  - [ ] Nice-to-have: Also show a small avatar 
Very low level mockup:
```
🔵  😊 Nimisha Vijay
🟢  🧔 Charlie Doe
🟣  🙂 Alice Bobb
```
I can post a mockup soon if it will be helpful 👍

ChristophWurst:
@hamza221 have the change requests by Nimisha been addressed already?

nimishavijay:
Nice! Looks very cool! 
- [x] Only major change would be to have the list of participants on the left as a vertical list of list items below the search field  (still outside the fullcalendar component)

Minor changes at first glance
- [ ] Can we show a shortened date on the top row? or is it according to the locale?
- [ ] Could we automatically scroll to a more meaningful time for the user? for eg. 1 hour before the start of their workday according to their availability, or 8:00 AM as a fallback
- [x] As a follow up we could maybe show the names or avatars of the attendees inside their blocks in the fullcalendar component if possible

nimishavijay:
Notes from the design review :) 

- [x] Show week # in the middle, \[today\] \[<\] \[date picker\] \[>\] \[?\]
- [x] day view as full buttons (secondary selected and tertiary unselected)
- [x] Attendee list on the left with colour indicator
- [x] Full screen dialog
- [x] Unclear if available or unavailable -- use the striked out bg with colors
- [x] 2 events side to side if more than one person is busy in a slot
- [x] Show names of attendees in the block? show colors with stripes with the avatar? 
- [x] show the summary in the sticky section with done
- [x] show available times section on the top? (do we even need it)?
- [x] Show a selected time already in the calendar
- [x] show own OOO as grey
- [x] Show others as striked block in their own color 

## Mobile

- [x] Switch to day view in mobile
- [x] Keep user bubbles in mobile

hamza221:
@nimishavijay  I updated the screenshots, let me know what should get changed once you get a chance

st3iny (commented):
Nvm, it should look like that.

nimishavijay:
Super nice! The fullcalendar itself looks very clear now and the attendees are easily recognizable :)
Some spacing and wording changes:

- [x] The modal itself can be fullscreen but the content inside it can have a max-width of 1200px so that it doesn't get too stretched on large screens. Everything should fit inside this. we can also have multiple breakpoints for different screen sizes in the follow ups
- [x] `Week #25` --> `Week 25 of 2025` (so "Week [week number] of [year]" to be consistent with the regular calendar)
- [x] Move the `Today` `<` `2025.06.20` `>` `?` to be aligned with the start of the fullcalendar component
- [x] Move the available times dropdown to the column with the participants, and remove the external label
- [x] Inside the dropdown, `Select automatic slot` --> `Suggested times`
- [x] Add a small heading "Attendees" above the attendees dropdown to differentiate it from the other dropdown
- [x] Instead of the trash icon use the X icon
- [x] Add a background behind the summary and done button

Rough mockup of all the changes except the max-width: 1200px
image

For mobile:
- [x] Left align all rows
- [x] The order: available times, then attendee search, then attendees
- [x] Dropdowns can be full width

image

I'm wondering if there's something we can do about the stripes? they look a bit uneven and jarring, I was thinking if it's possible to have something like when an event is set as tentative. I think @GVodyanov worked on that feature

image

GVodyanov:
Hey! This looks really cool, the suggesting mechanism for empty slots seems to work. The only thing is after accepting the suggestion and pressing done, the value isn't updated in the sidebar.

SebastianKrupinski:
Hey! Looks great and works good too.

Only two small visual issues, I am using a laptop with a screen res of 1920x1080. 

- [x]  The footer overlapping the contents (wording and buttons are on top of the calendar view)

image

- [x]  Is it possible to scroll the calendar instead of the entire modal view? If you need to scroll to the bottom of the calendar you then have to scroll back up to make any changes like adding a participant or using the day controls to forward a week, etc

image

src/components/Editor/AddTalkModal.vue
```diff
@@ -137,7 +137,7 @@ export default {
 					&& (conversation.type === CONVERSATION_TYPE_GROUP
 						|| (conversation.type === CONVERSATION_TYPE_PUBLIC
 							&& conversation.objectType !== CONVERSATION_OBJECT_TYPE_VIDEO_VERIFICATION))
-					&& conversation.objectType !== CONVERSATION_OBJECT_TYPE_EVENT
+					&& conversation.objectType !== CONVERSATION_OBJECT_TYPE_EVENT,
```
ChristophWurst, line 140: Good call :+1:

src/components/Editor/FreeBusy/FreeBusy.vue
```diff
@@ -4,129 +4,240 @@
...
+									</template>
+								</NcButton>
+							</template>
+							<template #default>
+								<div class="freebusy-caption">
+									<div class="freebusy-caption__calendar-user-types" />
+									<div class="freebusy-caption__colors">
+										<div class="freebusy-caption-item">
+											<div class="freebusy-caption-item__color"
+												:style=" { 'background': 'repeating-linear-gradient(45deg, #dbdbdb, #dbdbdb 1px, transparent 1px, transparent 3.5px)'}" />
```
ChristophWurst: nit: why inline style and not via a component style sheet?

src/components/Editor/FreeBusy/FreeBusy.vue
```diff
@@ -4,129 +4,240 @@
...
+							<NcListItemIcon v-for="attendee in attendees"
+								:key="attendee.id"
+								:name="attendee.commonName"
+								class="modal__content__body__sidebar__attendees__item">
+								<NcActions>
+									<NcActionButton @click="removeAttendee(attendee)">
+										<template #icon>
+											<Close :size="20" />
+										</template>
+										Delete
```
ChristophWurst: l10n

src/components/Editor/FreeBusy/FreeBusy.vue
```diff
@@ -401,14 +509,61 @@ export default {
...
+
+		},
+		eventDidMount(e) {
+			const eventElement = e.el
+			if (e.event.id === 'selected-event-slot') {
+				eventElement.style.setProperty('border', `2px solid ${this.personalCalendarColor}`, 'important')
+				return
+			}
+			if (e.event.title === this.organizer.attendeeProperty.commonName && e.el.classList.contains('free-busy-busy-unavailable')) {
+				eventElement.style.background = 'repeating-linear-gradient(45deg, #dbdbdb, #dbdbdb 1px, transparent 1px, transparent 3.5px)'
```
ChristophWurst: nit: could this be done with a dynamically added class to avoid styles in js strings?

src/components/Editor/FreeBusy/FreeBusy.vue
```diff
@@ -512,46 +678,130 @@ export default {
...
+				width: 260px;
+			}
+			&__date{
+				display: flex;
+				justify-content: space-between;
+				align-items: center;
+				& > *{
+					margin-inline-start: var(--default-grid-baseline);
+				}
+			}
```
ChristophWurst: this …

src/components/Editor/FreeBusy/FreeBusy.vue
```diff
@@ -512,46 +678,130 @@ export default {
...
+			&__title{
+				&__date{
+					display: flex;
+					justify-content: space-between;
+					align-items: center;
+					& > *{
+						margin-inline-start: var(--default-grid-baseline);
+					}
+				}
```
ChristophWurst: … looks like a duplicate of this?

ChristophWurst (commented):
Code review

nimishavijay (approved):
Latest screenshots look really nice! Only change is to align the items in the footer with the rest of the items as well, so that on very large screens the done button is not off in the corner of the screen. 
- [ ] left align the summary with the attendees list
- [x] right align the Done button with the end of the fullcalendar. 

Approving to not block as everything else looks great, nice work! :)

hamza221:
image

SebastianKrupinski (approved):

PR #6964, opened 2025-05-08
Fix timepicker not adjusting end date

GVodyanov:
Regression after https://github.com/nextcloud/calendar/pull/6888

To test: have an event with different start and end times, change start time, end time should also change to be the same interval as before


st3iny (approved):
Tested this and the old PR. Both cases still work.

Issue #6838, opened 2025-03-19
Error when removing invitees from event


kirlue:
### Steps to reproduce

/edit: I'm editing this bug to separate between the reoccurrence of #3325 and this (new?) bug.

The errors happen when changing events with Thunderbird as well as in the NC-interface of the calendar. All calendars are owned by one account and shared with other accounts. The changes are made with the user, that owns the calendars. So there it's not an issue of writing permissions.

The errors started to show a week ago. 

1. Create an event in a shared or not-shared calender.
2. Invite a person to the event, save. -> person gets invitation, doesn't even have to respond to this.
3. Try to delete the invitee -> error.

### Expected behavior

1. I would like to delete an invited person in the event -> invited person is deleted and gets a message about the cancellation.

### Actual behavior

1. I would like to delete an invited person in the event **-> NC calendar shows "error in saving the event"; the invited person gets a mail with "event cancelled" but actually is still invited** 
2. See error logs posted underneath.

I can still change the title or the description of the event. 
Sometimes  I was able to nevertheless change the events calendar (move it to another calendar) and then delete the invited person. Sometimes this wasn't possible neither. I can't see why this happens because for me up to now there is no pattern.

It is also possible to delete one invitee and invite another one -> the event is saved without a problem.

### Calendar app version

5.2.0

### CalDAV-clients used

Thunderbird 128.6.0

### Nextcloud version

Nextcloud Hub 9 (30.0.6)

### Browser log

Mar 19 21:31:01 xxxx.your-storageshare.de nextcloud-nxxxxx[640100]: 
{
    "reqId": "hcCaVxqSCF9dJOcV5scu",
    "level": 3,
    "time": "2025-03-19T20:31:01+00:00",
    "remoteAddr": "2a02:8109:9f1b:2600:f5aa:13d5:aed4:873a",
    "user": "buero",
    "app": "PHP",
    "method": "PUT",
    "url": "/remote.php/dav/calendars/xxx/yyyy/12407c6c-a2bb-4d87-a01b-4097bd8f98c7.ics",
    "message": "Undefined array key \"master\" at /var/www/html/apps/dav/lib/CalDAV/TipBroker.php#104",
    "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:135.0) Gecko/20100101 Firefox/135.0",
    "version": "30.0.6.2",
    "data": {
        "app": "PHP"
    }
}

AND 

[...]

PR #6980, opened 2025-05-15
fix: remove organizer when there are no attendees

SebastianKrupinski:
## Issue
* Resolves: #6838

## Resolution
- added logic to remove the organizer once the last attendee is removed


st3iny (approved):
Pair programming go brrrrrr

PR #7051, opened 2025-06-20
chore: update composer to php 8.4 max

SebastianKrupinski:
- Updated composer to php 8.4 max

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     },
     "require": {
-        "php": ">= 8.1 <=8.3",
+        "php": ">= 8.1 <=8.4",
         "bamarni/composer-bin-plugin": "^1.8.2"
     },
     "scripts": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "492b5cfee734ba356e1be314ca01accb",
+    "content-hash": "4e8c9e638551f81a2d3798c2f26e0e68",
     "packages": [
         {
             "name": "bamarni/composer-bin-plugin",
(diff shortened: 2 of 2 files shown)


SebastianKrupinski:
Looks like a bunch of the test will need to be fixed for php 8.4

st3iny (changes requested):
It looks like you accidentally updated phpunit hence the failing unit tests.

```
Error: Call to undefined method PHPUnit\Framework\MockObject\Builder\InvocationMocker::withConsecutive()
```

Please revert the changes in the psalm and phpunit vendor-bins.

SebastianKrupinski:
> It looks like you accidentally updated phpunit hence the failing unit tests.
> 
> ```
> Error: Call to undefined method PHPUnit\Framework\MockObject\Builder\InvocationMocker::withConsecutive()
> ```
> 
> Please revert the changes in the psalm and phpunit vendor-bins.

Yeah, composer did that automatically on me. It's better now.

st3iny (commented):
I'm not a fan of bundling manual dependency updates if not really required. We already have renovate to bump dependencies.

Is there a reason why we have to bump the minimum version of phpunit.

SebastianKrupinski:
> I'm not a fan of bundling manual dependency updates if not really required. We already have renovate to bump dependencies.
> 
> Is there a reason why we have to bump the minimum version of phpunit.

Without enforcing ^9 for phpunit "composer test" would not run.

image

Also any other dependencies that got bumped where a result of just bumping the coding standard with composer (granted maybe I didn't do it correctly lol)

st3iny:
That is probably related to accidentally bumping phpunit. Manually updating composer dependencies is error prone unfortunately.

I moved the cs-fixer related changes to the renovate PR here: https://github.com/nextcloud/calendar/pull/7052

For this PR it makes sense to only keep the PHP version bump:
> Updated composer to php 8.4 max

SebastianKrupinski:
> > Updated composer to php 8.4 max

Okay, I removed all the other changes

st3iny (approved):

PR #7072, opened 2025-07-01
feat: Meeting Proposals

SebastianKrupinski:
### Summary

Resolves: https://github.com/nextcloud/server/issues/50084

Proposal Editor
image

Proposal Overview
image

Proposal Participant Availability
image

Proposal Status and List
image

### DIFF
--- a/lib/Migration/Version5050Date20250701000001.php
+++ b/lib/Migration/Version5050Date20250701000001.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Calendar\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version5050Date20250701000001 extends SimpleMigrationStep {
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	#[\Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('calendar_proposal_dts')) {
+			return $schema;
+		}
+
+		$table = $schema->createTable('calendar_proposal_dts');
+		$table->addColumn('id', Types::BIGINT, [
+			'autoincrement' => true,
+			'notnull' => true,
+			'unsigned' => true
+		]);
+		$table->addColumn('uid', Types::STRING, [
+			'notnull' => true,
+			'length' => 255
+		]);
+		$table->addColumn('uuid', Types::STRING, [
+			'notnull' => true,
+			'length' => 36
+		]);
+		$table->addColumn('title', Types::TEXT, [
+			'notnull' => false,
+			'length' => 65535
+		]);
+		$table->addColumn('description', Types::TEXT, [
+			'notnull' => false,
+			'length' => 65535
+		]);
+		$table->addColumn('location', Types::TEXT, [
+			'notnull' => false,
+			'length' => 65535
+		]);
+		$table->addColumn('duration', Types::INTEGER, [
+			'notnull' => true,
+			'length' => null
+		]);
+
+		$table->setPrimaryKey(['id']);
+
+		return $schema;
+	}
+}
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -23,7 +23,7 @@
 * 📎 **Attachments!** Add, upload and view event attachments
 * 🙈 **We’re not reinventing the wheel!** Based on the great [c-dav library](https://github.com/nextcloud/cdav-library), [ical.js](https://github.com/mozilla-comm/ical.js) and [fullcalendar](https://github.com/fullcalendar/fullcalendar) libraries.
 	]]></description>
-	<version>5.6.0-dev.1</version>
+	<version>5.6.0-dev.2</version>
 	<licence>agpl</licence>
 	<author homepage="https://github.com/st3iny">Richard Steinmetz</author>
 	<author homepage="https://github.com/SebastianKrupinski">Sebastian Krupinski </author>
--- a/lib/Controller/ProposalController.php
+++ b/lib/Controller/ProposalController.php
@@ -1,0 +1,224 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Calendar\Controller;
+
+use OCA\Calendar\AppInfo\Application;
+use OCA\Calendar\Objects\Proposal\ProposalObject;
+use OCA\Calendar\Objects\Proposal\ProposalResponseObject;
+use OCA\Calendar\Service\Proposal\ProposalService;
+use OCP\AppFramework\ApiController;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\AnonRateLimit;
+use OCP\AppFramework\Http\Attribute\ApiRoute;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
+use OCP\AppFramework\Http\Attribute\PublicPage;
+use OCP\AppFramework\Http\Attribute\UserRateLimit;
+use OCP\AppFramework\Http\JSONResponse;
+use OCP\IGroupManager;
+use OCP\IRequest;
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\IUserSession;
+
+class ProposalController extends ApiController {
+
+	public function __construct(
+		IRequest $request,
+		private IUserSession $userSession,
+		private IUserManager $userManager,
+		private IGroupManager $groupManager,
+		private ProposalService $proposalService,
+	) {
+		parent::__construct(Application::APP_ID, $request);
+	}
+
+	private function authorize(?string $userId = null): JSONResponse|IUser {
+		// evaluate if user is logged in and has permissions
+		if (!$this->userSession->isLoggedIn()) {
+			return new JSONResponse([], Http::STATUS_UNAUTHORIZED);
+		}
+		if ($userId !== null) {
+			if ($this->userSession->getUser()->getUID() !== $userId
+				&& $this->groupManager->isAdmin($this->userSession->getUser()->getUID()) === false) {
+				return new JSONResponse([], Http::STATUS_UNAUTHORIZED);
+			}
+			if (!$this->userManager->userExists($userId)) {
+				return new JSONResponse(['error' => 'user not found'], Http::STATUS_BAD_REQUEST);
+			}
+			$user = $this->userManager->get($userId);
+		} else {
+			$user = $this->userSession->getUser();
+		}
+		if ($user === null) {
+			return new JSONResponse(['error' => 'user not found'], Http::STATUS_BAD_REQUEST);
+		}
+		return $user;
+	}
+
+	/**
+	 * Retrieve list of available proposals
+	 */
+	#[ApiRoute(verb: 'POST', url: '/proposal/list', root: '/calendar')]
+	#[NoAdminRequired]
+	#[UserRateLimit(limit: 10, period: 60)]
+	public function list(?string $user = null): JSONResponse {
+		// authorize request
+		$authorization = $this->authorize($user);
+		if ($authorization instanceof JSONResponse) {
+			return $authorization;
+		}
+		$userObject = $authorization;
+		// retrieve proposals for the user
+		$proposals = $this->proposalService->listProposals($userObject);
+
+		return new JSONResponse($proposals, Http::STATUS_OK);
+	}
+
+	/**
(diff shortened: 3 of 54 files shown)


SebastianKrupinski:
Only 3.2K of lines... I can do better 😆

ChristophWurst (dismissed):
Extract the typescript introduction into a separate PR

SebastianKrupinski:
Basic functionality is implemented.
- Proposal Creation
- Proposal Modification
- Proposal Deletion
- Proposal Management (List of proposals)
- Proposal Response (Participant date selection)
- Mail notifications on Create / Modify / Delete
- iMip Messaging to internal users on Create / Modify / Delete (Time blockers in calendar)

SebastianKrupinski:
Remain Tasks,
- Discuss TODO's with @ChristophWurst 
- Investigate why UI only shows one instance
- Implement Group expansion
- Implement Tests
- Functional Review with @ericfischereu and @nimishavijay 
- Design Review with @ericfischereu and @nimishavijay

ChristophWurst:
> * Mail notifications on Create / Modify / Delete
> 
> * iMip Messaging to internal users on Create / Modify / Delete (Time blockers in calendar)

These two sound optional. Does it bring your PR closer to a reviewable size when you split them off?

SebastianKrupinski:
> > * Mail notifications on Create / Modify / Delete
> > * iMip Messaging to internal users on Create / Modify / Delete (Time blockers in calendar)
> 
> These two sound optional. Does it bring your PR closer to a reviewable size when you split them off?

Those two where part of the original spec, also they are part of the feature that we are trying to replicate, and kind of necessary. The one thing we could drop is the iMip messaging but that is only 100 lines so not a huge amount of code.

SebastianKrupinski:
Completed initial review with @nimishavijay @jancborchardt @hamza221 today

nimishavijay:
Notes from the design review:

## Overall

* [x] When you fill in users list them vertically using listitems instead of userbubbles
* [x] Fullcalendar buttons should be adjusted to match NC design 
* [x] remove zoom buttons? unintuitive. instead make default snapto to 15 mins and bring up default editor to adjust
* [x] Instead of showing a blank block for past dates, show just the heading 
* [x] Instead of showing available blocks, show busy blocks with the same striked out pattern as in new free/busy 
* [x] Show a close button on a selected time in the fullcalendar 
* [x] When you click on a date, 
* [x] Don't show delete button if it can't be deleted 
* [x] Don't show close or cancel button
* [x] Track who voted 
* [ ] Follow up for next versions: Once you fill out organizers and duration, automatically suggest a few timeslots based on free/busy

## Navigation

- [x] Show list of proposals in the left navigation below appointment schedules in a new section 
- [x] Similar to apppointment schedules show a plus button next to the navigation caption which opens the create new proposals 
- [x] Each item can have an icon or 2 to indicate responses 

## Papercuts

- [x] "Create Meeting Proposal" --> "Create meeting proposal" use sentence case, not title case
- [x] Also similar for buttons in the bottom row
- [x] "today" "week" "day" --> "Today" "Week" "Day" in fullcalendar component
- [x] Move all the buttons to the right, and follow the order of having the primary button at the end, and then the rest of the buttons

## Voting screen

- [x] Always show organizer with avatar and name 
- [x] Show other participants if possible? 
- [x] Make the dates nicer, use the full words 
  - [x] If it is the current year, don't show the year
- [x] Possibly copy exactly the appointment voting screen
  - [x] Show a list of dates in the middle column, possibly arranged in a a clearer way (eg divide into sections based on month)
  - [x] On the left column show all the info about the proposal
  - [x] Add a timezone picker! 
  - [x] Possibly show who already responded to a date like in MS

## Email

- [x] Title, description, location, duration, organizer for first version
- [x] Participants and proposed times as a follow up? 

To do for Nim:
- [x] Design of time slots inside the proposal creation/update view 
- [x] Design of showing who responded to what inside the proposal view
- [x] Layout improvements to the voting page

ChristophWurst:
> Those two where part of the original spec, also they are part of the feature that we are trying to replicate, and kind of necessary. The one thing we could drop is the iMip messaging but that is only 100 lines so not a huge amount of code.

Okay. Thank you for the insights.

nimishavijay:
For the voting page:

Here are 3 versions of what it could look like:

1. First one is with just the exact info we have rn (with the addition of organiser + avatar)
(in case it's not obvious, the user will have to click on each date to select it and then select the yes/no/maybe for each timeslot)

image

The second and third one are with the addition of who voted for what

2. This is with the same layout as (1)
image

3. This may involve scrolling, but is less clicks to finish the whole task as we are not switching dates
image

ccing @jancborchardt as he is also at the GW hackweek :)

jancborchardt (changes requested):
Talked with @SebastianKrupinski about a first pass on information architecture regarding the general navigation:

Calendar navigation needs to have entries for the proposals:
- "Meeting proposals" section title
- "+" button to create a new meeting proposal on the right of it
- Every entry has: Status icon for [in progress](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:event_upcoming:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=calendar&icon.size=24&icon.color=%23e3e3e3) and [all voted](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:event_available:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=calendar&icon.size=24&icon.color=%23e3e3e3
), title, and respondent count like "0/2", "3/5" etc.
- Clicking on it will open it in View mode.
- Menu with "View", "Edit" and "Delete" to quickly do the functions

nimishavijay:
**Navigation**

- [x] Exchange positions of appt schedules and meeting proposals
- [ ] Maybe include explanation somewhere
- [x] "Configured email address.." should wrap
- [x] Todo Nim: meeting overview 

**While creating a proposal**

- [x] "Virtual" not clear --> "Add Talk conversation after everyone votes" checkbox, onclick fills out the location field with "Talk conversation"
- [ ] Add calendar selector 
- [x] Use checkboxradioswitch for predefined times
- [x] Participant list
  - [x] If required, remove icon
  - [x] If optional show optional icon and sort to the bottom
- [x] Reverse styling of available and busy slots 
- [x] Show all availabilities in free/busy (default free/busy view)
- [x] Change icon of zoom in and out
- [x] Dates
  - [x] trash icon --> close icon
- [ ] When you click on create add a loading spinner inside the button

**Voting page**

- [x] 404 page: use sentence case for "No meeting proposals..."
- [x] Todo Nim: add all responses 

**After voting**

- [ ] Send an email to the organizer when everyone has voted --> follow up

**Mobile**

- [ ] Free busy is missing! Second page with free busy

**General**

- [ ] Add duration hh:mm component (upstream)

nimishavijay:
As discussed, new designs for meeting overview and voting page:

**Meeting overview:**

Edit: updated mockups with @jancborchardt's feedback about clearer "yes" responses

image

<details> <summary> Old mockups </summary>

image

</details>

- [x] min-width 600px, it can increase in width for upto 10 votes, and after that it becomes fullscreen
- [x] Meeting details on the top, votes on the bottom
- [x] meeting details section:
  - [x] one row with title
  - [x] one row with description (no "Description" title necessary)
  - [x] one row with location if set or Talk room selected
  - [x] one row with duration in bold and edit proposal tertiary button
- [ ] votes section:
  - [ ] border: 1px solid var(--color-border)
  - [x] each row is minimum 34px to make sure we fit an accessible button in there
  - [x] rows have a gap of 2px between them, and all the cells with the votes have a gap of 2px
  - [x] scrollable
  - [x] 1st row with names of participants 
    - [ ] required participants sorted to the beginning
    - [x] since they will pretty much always ellipsize, on hover show a tooltip with full name
    - [x] nice-to-have: avatars
  - [ ] rest of the rows show time, date is used as a separator
    - [x] edit: "yes" responses are marked with a background of `rgba(var(--color-success-rgb), 0.1)` and the icon color is `var(--color-success)`
    - [ ] if everyone responded yes, highlight that row in `rgba(var(--color-success-rgb), 0.1)` and add a border-radius: var(--border-radius-element)

**Voting page:**
image

Pretty much exactly the same as the meeting overview with some changes in buttons for the different functionality
- [x] Meeting details section:
  - [x] no edit proposal button, instead there is the meeting organizer's name and avatar (20px) as the first element
- [ ] Votes section
  - [x] There is a "Share your availability" heading inside the votes section 
  - [ ] no highlight if everyone responded yes
  - [x] No create button, instead we have the yes/no/maybe buttons after the time
  - [x] I thought we could use the NcCheckboxRadioSwitch component for the yes/no/maybe buttons. if we do some overrides (`flex-direction: row; gap: 4px`, removing that border and changing the background color to --primary-light), we could make them look like how they do in the mockup 
image

SebastianKrupinski:
@nimishavijay 

The mockups look great, I will do my best to make the real thing look as amazing!

jancborchardt:
Nice, great updated mockups! :)

SebastianKrupinski:
Latest version of the Organizer Overview Screen

Avatars are present they just don't show right on my dev instance

image

ericfischereu:
@nimishavijay @SebastianKrupinski One thing that came into my mind is that it would be possible as soon as the first consent is reached, the event could be scheduled without further user interaction. So basically if everyone voted, take the first option where everyone is available and automatically schedule the meeting and notify the attendees.

But I guess we can keep that for a follow up to finish the base feature quicker.

SebastianKrupinski:
@ericfischereu @nimishavijay 

Please have a look at the updated screen captures in the PR overview at the top

lib/Migration/Version5050Date20250701000001.php
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	#[\Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('calendar_proposal_dts')) {
+			return $schema;
+		}
+
+		$table = $schema->createTable('calendar_proposal_dts');
```
st3iny, line 33: Missing a column in the migration: `uuid`

`Internal Server Error
An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'uuid' in 'INSERT INTO'`

Belonging to `ProposalDetailsEntry::$uuid` I believe.

.eslintrc.js
```diff
@@ -6,7 +6,36 @@ module.exports = {
...
+				project: [__dirname + '/tsconfig.json'],
+			},
+			node: {
+				extensions: ['.js', '.ts', '.vue'],
+			},
+		},
+	},
+	rules: {
+		// Let TS resolver handle path aliases like "@/…"
+		'n/no-missing-import': 'off',
```
st3iny:
```suggestion
```

Nope, please revert. That is too generic and we will miss valid eslint errors this way.

.eslintrc.js
```diff
@@ -6,7 +6,36 @@ module.exports = {
...
+		// Allow both extensionless and extension-based imports
+		// (the bundler/TS config handles resolution)
+		'import/extensions': 'off',
```
st3iny: This can be configured inside `tsconfig.json` via `"allowImportingTsExtensions": true` (and also `"noEmit": true` to make it work).

Ref https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions

st3iny (changes requested):
I can't create a new meeting proposal:

`POST https://nextcloud.local/ocs/v2.php/calendar/proposal/create`
```xml
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>996</statuscode>
  <message>Internal Server Error
votedYes is not a valid attribute</message>
 </meta>
 <data/>
</ocs>
```

**EDIT:** My bad, I didn't run the latest migrations.

.eslintrc.js
st3iny: Please add all missing extensions to imports afterward.

nimishavijay:
Awesome! Looks very cool! :) Only 4 minor changes:
- [x] Proposal Editor: 
  - [x] make the update button a wide button
  - [x] There seems to be some spacing issues between the avatar and the text in the list of participants, but I'm not sure since the avatars are not loaded. There should be atleast 8px space between the avatar and the text
  - [x] "Dates" --> "Selected times"
- [x] Proposal Participant Availability:
  - [x] Add the thin horizontal line here too
  - [x] The time zone picker is next to the duration now, it should be pushed all the way to the right on that row

Nothing other than that, so I'm approving to not block :)

st3iny (commented):
Tested this and found some issues:

1. Attendee emails: Date is always `Thursday, January 1, 1970`
2. Avatars are not working in the modal and on the public page (probably due to display name being used instead of user id).
3. Creating an event is not working properly (after attendees reacted).

SebastianKrupinski:
@nimishavijay @st3iny thank you, I will make the changes later today

SebastianKrupinski:
> Tested this and found some issues:
> 
>     1. Attendee emails: Date is always `Thursday, January 1, 1970`
> 
>     2. Avatars are not working in the modal and on the public page (probably due to display name being used instead of user id).
> 
>     3. Creating an event is not working properly (after attendees reacted).

Fixed. 

1. Attendee email dates now show properly with time span and no longer are bulleted
2. Avatars now show properly
3. Converting a meeting to a proposal now works properly, although the is a issue, that will need to be addresses in a followup
4. Fixed timezone selector on public view , it now changes the times properly
5. Fixed selected time slot drag and drop

st3iny (changes requested):
Creating the event is not working still.

I can see network requests and the proposal being deleted. However, there is no calendar event being created in the organizer's calendar.

SebastianKrupinski:
> Creating the event is still not working.
> 
> I can see network requests and the proposal being deleted. However, there is no calendar event being created in the organizer's calendar.
> 
> The rest seems to be fixed now.

Yeah that was the glitch I mentioned in the last comment. It resolved.

st3iny:
Let's drop support for Nextcloud < 32 and merge this to circumvent the iMip API incompatibilties on Nextcloud < 32 (and < 31 respectively).

SebastianKrupinski:
Temporarily removed the time blocking part of this feature (iMip Sending) due to missing changes in server.

Will be added in:

https://github.com/nextcloud/calendar/pull/7315

PR #7079, opened 2025-07-02
perf: don't load ContactsMenuScript when not logged in

joshtrichards:
Not sure how this interacts with guests, but I don't think #6502 currently works with guests anyhow.

Cuts at at least ~700 kB from the initial JS assets that are loaded when visiting `/login`

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -78,7 +78,11 @@ class Application extends App implements IBootstrap {
 		// TODO: drop condition once we only support Nextcloud >= 31
 		if ($serverVersion->getMajorVersion() >= 31) {
 			// The contacts menu/avatar is potentially shown everywhere so an event based loading
-			// mechanism doesn't make sense here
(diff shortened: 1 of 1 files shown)


lib/AppInfo/Application.php
```diff
@@ -78,7 +78,11 @@ private function addContactsMenuScript(ContainerInterface $container): void {
 		// TODO: drop condition once we only support Nextcloud >= 31
 		if ($serverVersion->getMajorVersion() >= 31) {
 			// The contacts menu/avatar is potentially shown everywhere so an event based loading
-			// mechanism doesn't make sense here
+			// mechanism doesn't make sense here - well, maybe not when not logged in yet :-)
+			$userSession = $this->getContainer()->get(IUserSession::class);
```
tcitworld: `IUserSession` needs to have a `use` statement for the correct namespace

st3iny (approved):
Thank you very much :)

This is a clear oversight.

My change: Use the local container in the method instead of the one from the class.

PR #7089, opened 2025-07-07
Fix readonly event formatting in full page calendar modal

GVodyanov:
| Before | After  |
| - | - | 
| image | swappy-20250707_174048 |


SebastianKrupinski:
Tested

GVodyanov:
I also deleted the `NoAttendeesView` component because at this point it wasn't useful. 

Now if you are an editor, you just see the search bar and a list.

If you are an invitee, you see a list if there are attendees, and nothing if there aren't any.

PR #7132, opened 2025-07-18
fix: show time zone selector

SebastianKrupinski:
## Issue
When 10 people review a feature and no one thinks of changing the time zone of a new event. 

## Resolution
- Modified logic to make time zone selector show properly when event is NOT all day.

### Before
image

### After
image


st3iny (approved):
Tested and works.

PR #7208, opened 2025-08-11
Make transparent events be background color instead of transparent to prevent overlapping issues

GVodyanov:
| Before | After |
| - | - |
| swappy-20250811_142523 | swappy-20250811_142331 |


SebastianKrupinski (approved):

hamza221 (approved):
LGTM from the screenshot, how would the gradient look tho ?

PR #7226, opened 2025-08-18
Update styling for "no email" warning while inviting attendees

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/7215

| Before | After |
| - | - |
| image | image |


SebastianKrupinski:
We should also change that plus sign! It does not look right.

SebastianKrupinski:
Tested. Looks good

Issue #7035, opened 2025-06-10
Show meeting time in invitees' timezones


ericfischereu:
### Is your feature request related to a problem? Please describe.

If planning meetings with people in different time zones, it is difficult to chose a time within usual working hours, because you need to manually calculate the desired attendees' time.

### Describe the solution you'd like

The meeting time is automatically calculated for the attendees' time zones and displayed in the attendees list. This should also show the date (e.g. meeting between Germany and US west coast - 6:30 in Germany means around 22:30 the day before at the west coast). Ideally, the date is only shown if there is a difference.

### Describe alternatives you've considered

Calculating manually as described in the problem statement.

jancborchardt:
@ericfischereu is this something which would work?
- Show the local time in the subline of participants in free/busy dialog
- Only show the local time for people who have a different one from yours
- Of course needs to account for day before too, so example values are "9:00 local time", "18:30 local time", "22:30 local time, day before", etc.
- It’s not that important to show the actual timezone, so I left it out. If we want it, it could just be added in parentheses like "9:00 local time (Eastern Time)"
- We could think about making the height of all the rows larger in general so it looks more uniform, the drawback would of course be lower information density.

FYI @nimishavijay, also to check how Teams and Google do this.

![Image](https://github.com/user-attachments/assets/3b57f62d-8768-41ea-be11-46aa2aa1a474)

ericfischereu:
@jancborchardt Thanks for the draft, looks good. But I think this would already be valuable in this view:

![Image](https://github.com/user-attachments/assets/7f5f086f-ba6c-43e9-9ae0-adbab074b3db)

The information you proposed is exactly what would be needed. I would make the addition that any day difference should be shown (e.g. if you're based in the US west cost and set up a meeting at 10PM, it is already the next day 6AM in Germany)

PR #7303, opened 2025-08-29
feat: add local time when inviting attendees to events

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/7035

image

### DIFF
--- a/src/components/Editor/AvatarParticipationStatus.vue
+++ b/src/components/Editor/AvatarParticipationStatus.vue
@@ -22,6 +22,9 @@
 				:size="20" />
 			<div class="avatar-participation-status__text">
 				{{ status.text }}
+				<span v-if="adjustedTime" class="avatar-participation-status__text__time">
+					, {{ adjustedTime }} local time
+				</span>
 			</div>
 		</template>
 	</div>
@@ -34,6 +37,9 @@ import IconCheck from 'vue-material-design-icons/CheckCircleOutline.vue'
 import IconNoResponse from 'vue-material-design-icons/HelpCircleOutline.vue'
 import IconClose from 'vue-material-design-icons/CloseCircleOutline.vue'
 import IconDelegated from 'vue-material-design-icons/ArrowRightDropCircleOutline.vue'
+
+import useCalendarObjectInstanceStore from '../../store/calendarObjectInstance.js'
+import { mapStores } from 'pinia'
 
 export default {
 	name: 'AvatarParticipationStatus',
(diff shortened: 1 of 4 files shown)


SebastianKrupinski:
Hey,

My only comment on the visuals would be that the wording makes it appear that the "Invitation will be sent at 6.00pm" maybe having the time zone under the user name would be better.

GVodyanov:
> Hey,
> 
> My only comment on the visuals would be that the wording makes it appear that the "Invitation will be sent at 6.00pm" maybe having the time zone under the user name would be better.

Sure @SebastianKrupinski, how about this?

image

(this is what the last commit looks like now)

SebastianKrupinski:
> Sure @SebastianKrupinski, how about this?

I think that is much better. I would just check what design thinks of this.

In the mean time I will do a code review

SebastianKrupinski (approved):

SebastianKrupinski:
I just had a thought how we could solve a couple of other issue with this PR.

FYI, this is not a blocker and I don't want to delay this but these are just issue we have had requests/complaints for...

A. Invitation being sent to participants are all calculated in the organizers timezone
B. Invitation being sent to participants are in the organizers language

Since this information is already available in the users profile and in contacts, and you are already pulling the users details and contacts details, We can pretty easily fix those two issues, all this would require is for the pulled information to be added to the attendee of the event.

Then when the event is saved, we can use that info to send invitation in the participants timezone and language. As added benefit we could skip pulling this info for events that already contain it. The RFC also already has parameters we can use TZID and LANGUAGE.

The modification to your PR would be minor, then I could make the changes in the backend.

Let see what the team thinks

GVodyanov:
> I just had a thought how we could solve a couple of other issue with this PR.
> 
> FYI, this is not a blocker and I don't want to delay this but these are just issue we have had requests/complaints for...
> 
> A. Invitation being sent to participants are all calculated in the organizers timezone B. Invitation being sent to participants are in the organizers language
> 
> Since this information is already available in the users profile and in contacts, and you are already pulling the users details and contacts details, We can pretty easily fix those two issues, all this would require is for the pulled information to be added to the attendee of the event.
> 
> Then when the event is saved, we can use that info to send invitation in the participants timezone and language. As added benefit we could skip pulling this info for events that already contain it. The RFC also already has parameters we can use TZID and LANGUAGE.
> 
> The modification to your PR would be minor, then I could make the changes in the backend.
> 
> Let see what the team thinks

Sounds like a cool idea to me!

About design Jan already left a comment about this https://github.com/nextcloud/calendar/issues/7035#issuecomment-2965755245 and I based myself off of it.

ChristophWurst:
Let's tackle one problem at a time

Please get the timezone information displaying in first

PR #7341, opened 2025-09-03
fix: from alignment

SebastianKrupinski:
### Summary
- Fixed "From" alignment

After
image

Before
image


SebastianKrupinski:
> Does not solve the problem fully on my machine (Firefox).

The PR was to address the lack of space between the word "from" and the input, but you are right input is slightly of to the left, Its barely even noticeable on me screen

st3iny:
Yeah, I noticed that. But we can't fix one problem just to create another one.

SebastianKrupinski:
> Yeah, I noticed that. But we can't fix one problem just to create another one.

I agree I fixed it already

image

st3iny (approved):
Tested and works.

PR #7395, opened 2025-09-10
fix: adjust tentative status icon colour

SebastianKrupinski:
### Summary

- Adjusted icon color for tentative status

image


SebastianKrupinski:
@nimishavijay

I was doing some testing on participant invitation responses and noticed that it was difficult to distinguish the accepted status from the tentative status as they are both green check marks.

What do you think?

nimishavijay:
Definitely makes sense! We should use another differentiator than just color for a11y. How about we use the question mark icon that is currently used for "Awaiting response", and for awaiting response we don't use any icon?

SebastianKrupinski:
> Definitely makes sense! We should use another differentiator than just color for a11y. How about we use the question mark icon that is currently used for "Awaiting response", and for awaiting response we don't use any icon?

image

How about this, I aligned the colours and icons with the meeting proposals, alternatively we could use a ProgressQuestion but I think that is to close to the tentative icon, alternatively there is ProgressClock

image

image

SebastianKrupinski:
@nimishavijay ^^ please

nimishavijay:
Looks good -- thought I 👍'ed Sebastian's comment but I had missed it.

PR #7398, opened 2025-09-11
fix: actually create a talk room when converting proposal

SebastianKrupinski:
### Summary
- Creates a talk room when a proposal is converted to a event.

### Testing
- Create meeting proposal, with add talk conversation selected
- Convert a meeting proposal date to a calendar appointment/event
- Then open the appointment and click on the talk link


src/services/autocompleteService.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+import { generateOcsUrl } from '@nextcloud/router'
+import type { AutocompleteEntry } from '@/types/autocomplete'
+import type { OcsEnvelope } from '@/types/ocs'
+
+export type AutocompleteResponse = OcsEnvelope<AutocompleteEntry[] | AutocompleteEntry>
+
+export type AutocompleteOptions = {
+    search: string
```
st3iny:
```suggestion
	search: string
```

src/services/talkService.ts
```diff
@@ -0,0 +1,149 @@
...
+  emoji?: string | null
+  avatarColor?: string | null
+  participants?: {
+    users?: string[]
+    federated_users?: string[]
+    groups?: string[]
+    emails?: string[]
+    phones?: string[]
+    teams?: string[]
+  }
```
st3iny, line 43:
```suggestion
	roomType: number
	roomName: string
	objectType: string
	objectId: string
	password?: string
	readOnly?: number
	listable?: number
	messageExpiration?: number
	lobbyState?: number
	lobbyTimer?: number | null
	sipEnabled?: number
	permissions?: number
	recordingConsent?: number
	mentionPermissions?: number
	description?: string
	emoji?: string | null
	avatarColor?: string | null
	participants?: {
		users?: string[]
		federated_users?: string[]
		groups?: string[]
		emails?: string[]
		phones?: string[]
		teams?: string[]
	}
```

SebastianKrupinski:
> Breaks when Talk is disabled.
> 
> You need to add a check if Talk is not installed/enabled and hide the `Add Talk conversation` checkbox.

Ah yes, that is a donkey oversight on my part.

SebastianKrupinski:
> Breaks when Talk is disabled.
> 
> You need to add a check if Talk is not installed/enabled and hide the `Add Talk conversation` checkbox.

Done. Added two checks

- Check one in the editor to hide the the talk conversation option
- Check two in the store in case the proposal was created with the option selected but talk is turned off for some reason later
- Also removed the redundant location input lock as the location field is now hidden when the "add talk conversation" is selected

src/services/talkService.ts
SebastianKrupinski: Are you saying to change the spacing? I ran lint:fix on all the files, this should be the default tab spacing

src/services/talkService.ts
```diff
@@ -0,0 +1,149 @@
...
+					users.push(matches[0].id)
+					continue
+				}
+				// Fallback to email if no good user match
+				emails.push(participant.address)
+			} catch {
+				// On errors, fallback to email
+				emails.push(participant.address)
+			}
+		} else {
```
st3iny: This does not seem to work. On my instance, a user is always added via email and thus not invited to the room.

st3iny (commented):
Tested again but found another bug.

src/services/talkService.ts
SebastianKrupinski: Hey,

So I finally had a look at this, but I am not sure, why your room is getting created with emails only,

It is discovering the users properly
image

And seems to be creating the room with users proproperly
image

SebastianKrupinski: I had another thought on this, I think maybe the issue was that you had contacts with the same email address as the user and the contacts where being returned first. 

Also made the whole process async to make it faster and fixed a bug in the backend

src/services/talkService.ts
st3iny: Yes. There are spaces in that file while we use tabs. Perhaps we should add an eslint rule for that.

src/services/talkService.ts
st3iny: Indeed, I clicked on the SAB contact instead of the actual user. Perhaps the SAB can be skipped when searching for contacts? Users are shown anyway.

(**Not required for this PR.**)

src/services/talkService.ts
st3iny: I'm gonna add a rule for this. Let's handle this in another PR.

PR #7400, opened 2025-09-11
fix: psalm error for ci

SebastianKrupinski:
### Summary
- Fixed 
ERROR: ArgumentTypeCoercion - lib/Service/Proposal/ProposalService.php:327:38 - Argument 1 of DateTimeZone::__construct expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
                        $eventTimezone = new DateTimeZone((string)$options['timezone']);

### DIFF
--- a/lib/Service/Proposal/ProposalService.php
+++ b/lib/Service/Proposal/ProposalService.php
@@ -324,7 +324,9 @@ class ProposalService {
 		// timezone option
 		$eventTimezone = null;
 		if (isset($options['timezone']) && is_string($options['timezone']) && in_array($options['timezone'], DateTimeZone::listIdentifiers(), true)) {
-			$eventTimezone = new DateTimeZone($options['timezone']);
+			if (!empty($options['timezone'])) {
+				$eventTimezone = new DateTimeZone($options['timezone']);
+			}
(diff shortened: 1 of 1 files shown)


lib/Service/Proposal/ProposalService.php
```diff
@@ -324,7 +324,9 @@ public function convertProposal(IUser $user, int $proposalId, int $dateId, array
 		// timezone option
 		$eventTimezone = null;
 		if (isset($options['timezone']) && is_string($options['timezone']) && in_array($options['timezone'], DateTimeZone::listIdentifiers(), true)) {
```
st3iny, line 326:
```suggestion
		if (isset($options['timezone']) && is_string($options['timezone']) && !empty($options['timezone']) && in_array($options['timezone'], DateTimeZone::listIdentifiers(), true)) {
```

To prevent nested ifs.

SebastianKrupinski: I tried that first didn't work

image

image

st3iny: Ok, Psalm is weird. Fine by me then.

Issue #7451, opened 2025-09-22
Dependency conflict, running npm i causes build failure


GVodyanov:
### Steps to reproduce

1. Checkout main
2. Run `npm i` or manual link dependency like `calendar-availability-vue`
3. Run `npm run dev` and see the following errors:

```
ERROR in ./src/views/EditSimple.vue
Module build failed (from ./node_modules/vue-loader/dist/index.js):
TypeError: Cannot read properties of undefined (reading 'styles')
    at Object.loader (/home/scratchx/NextcloudDev/nextcloud-docker-dev/workspace/server/apps-extra/calendar/node_modules/vue-loader/dist/index.js:96:34)
 @ ./src/router.js 10:0-48 29:17-27 42:17-27 92:17-27 108:17-27
 @ ./src/main.js 9:0-33 39:2-8 45:19-25
```

### Expected behavior

No errors

### Actual behavior

Errors

teutat3s:
This issue is back after the `vue-loader` version bump in #7617. Reproduced with version `v6.2.0-beta.1`.

st3iny:
@teutat3s I can't reproduce the problem. Are you using the recommended node and npm versions as per the `package.json` file?

```json
"engines": {
    "node": "^24.0.0",
    "npm": "^11.3.0"
  },
```

teutat3s:
Sorry for the noise, I can no longer reproduce with Nodejs version 24.

PR #7452, opened 2025-09-22
fix: set vue loader to compatible version

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/7451


GVodyanov:
Once this gets merged we can also update `calendar-availability-vue`

PR #7496, opened 2025-09-26
Fix wrong end date calculation

### LINKED ISSUE #7317: The day Until automatically jumps to the next day
### Steps to reproduce

As soon as you create a calendar entry starting at 7:00 or 8:00, the to-date jumps to the next day. Calendar entries starting at 9:00 remain on the same day.

### Expected behavior

The to-date should also remain on the same day for these times

### Actual behavior

The to-date jumps to the next day

### Calendar app version

5.5.1

### CalDAV-clients used

_No response_

### Browser

Chrome 139.0.7258.155

### Client operating system

Win10

### Server operating system

Debian

### Web server

Apache

### Database engine version

MariaDB

### PHP engine version

PHP 8.2

### Nextcloud version

Nextcloud Hub 10 (31.0.8)

Updated from an older version

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

_No response_

### LINKED ISSUE #7409: NextCloud Calendar on version 31.0.7 Date/Time Bug for Event Creation
### Steps to reproduce

Hello!

I have been using the NextCloud calendar app for a long time now and we have recently upgraded to the latest version.

I have encountered a bug pertaining to event creations. If I want to make a calendar event that was (for example) this upcoming Wednesday at 3PM, my typical process would be as follows:

Click on the day to open the new event creation screen
Uncheck “all day” since it is not an all day event
Manually set the time of the hour the event starts
What now happens:

I click on the day and the new event creation screen comes up
I uncheck “all day”
When I manually set the time to 3PM, it also automatically adjusts the end day to a day further than it should. In other words, my Wednesday @3PM appointment would automatically adjust to go on until Thursday.

Video example of issue and how to reproduce it: 
https://youtu.be/gXhRJusAhl8

### Expected behavior

When I make a calendar event and uncheck all day, then set a particular time I am looking for the following to happen:

Nextcloud automatically adjusts the end time to be an hour after the start time I set AND it properly keeps the same date. If I make an event on Nov. 14th and set the start time to 1PM on the 14th, I want it to be from 1PM-2PM on the 14th only.

### Actual behavior

However, right now it is automatically pushing the end date ahead by one day so the short meeting event that is only supposed to span an hour looks like it is going from Nov. 14th at 1PM to Nov. 15th 2PM.

### Calendar app version

31.0.7

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

Samba

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

31.0.7

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

_No response_

max65482:
Fixes #7317
Fixes #7409

### DIFF
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -1565,25 +1565,26 @@ export default defineStore('calendarObjectInstance', {
 			onlyTime = false,
 			changeEndDate = true,
 		}) {
+			// Calculate current duration between start and end
+			const oldDuration = calendarObjectInstance.endDate.getTime() - calendarObjectInstance.startDate.getTime()
+
 			if (onlyTime) {
 				startDate.setFullYear(calendarObjectInstance.startDate.getFullYear(), calendarObjectInstance.startDate.getMonth(), calendarObjectInstance.startDate.getDate())
-			}
-
-			// Changing the end date first is needed to not have the start date be after the end date
-			if (changeEndDate) {
-				const difference = startDate.getTime() - calendarObjectInstance.startDate.getTime()
-				const endDate = new Date(calendarObjectInstance.endDate.getTime() + difference)
-
-				this.changeEndDateMutation({
-					calendarObjectInstance,
-					endDate,
-				})
 			}
 
 			this.changeStartDateMutation({
 				calendarObjectInstance,
 				startDate,
 			})
+
+			// When changing time, preserve the original duration
+			if (changeEndDate) {
+				const newEndDate = new Date(startDate.getTime() + oldDuration)
+				this.changeEndDateMutation({
+					calendarObjectInstance,
+					endDate: newEndDate,
+				})
+			}
 		},
 
 		/**


SebastianKrupinski:
Hi @max65482 

I tested this PR, but I don't seem to notice any difference in functionality, what am I missing?

max65482:
Can you reproduce #7317? Note the exact hour range that triggers the problem.

SebastianKrupinski:
> Can you reproduce #7317? Note the exact hour range that triggers the problem.

Yes, I can reproduce the issue. But after applying the patch the issue was still present

image

image

max65482:
I know this isn't helping but: Works on my machine.
Even after a rebase.
The only difference is the date/time format (which I didn't manage to change).

grafik

The underlying server version shouldn't matter, right?

SebastianKrupinski:
@max65482 

The server version does not matter, as long as you are on the dev version of the calendar app (6.1 dev).

But I think the issue is related to timezones, when the date changes past the UTC boundry is flips to another day

max65482:
My understanding is that the issue was caused by the fact that the end date was changed first.
The issue only happened when setting the start date before 9.00 (default start time 10.00 - default duration 1h).

SebastianKrupinski:
When I tested this, I did not touch the end date, just changed the start time.

max65482:
I also didn't touch the end date.
I was referring to the implementation that causes the issue:
https://github.com/nextcloud/calendar/blob/bb5ebb220b2ac0c5d90a8b6dd2d524bb159cc181/src/store/calendarObjectInstance.js#L1577-L1586
Sorry for the confusion.

This logic causes the end date to be before the start date and triggers these lines (which add one day):
https://github.com/nextcloud/calendar/blob/bb5ebb220b2ac0c5d90a8b6dd2d524bb159cc181/src/store/calendarObjectInstance.js#L224-L225

SebastianKrupinski (approved):

SebastianKrupinski:
@max65482 

Okay, so yes this solution works, for some reason, the first time I tested this, it did not compile correctly. Idk. But after manually coping the code and replacing the function, it works.

PR #7537, opened 2025-10-08
fix(appointments): unify IDs

ChristophWurst:
Mixing two IDs makes reading logs more complex than necessary. Let's be consistent.


ChristophWurst:
CI fix at https://github.com/nextcloud/calendar/pull/7532

st3iny (approved):

ChristophWurst:
Tests are not passing

PR #7608, opened 2025-10-21
fix(global.css): z-index style leaking

GVodyanov:


ChristophWurst:
is backport to stable6.0 fine?

GVodyanov:
> is backport to stable6.0 fine?

Not needed, the styling isn't present there

PR #7609, opened 2025-10-21
fix(fullcalendar.scss): calendar grid headers overlapping

GVodyanov:
| Before | After |
| - | - |
| image | image |


GVodyanov:
Gotta rebase it first hol' up

PR #7610, opened 2025-10-21
fix(InviteesList): change attendee summary logic

GVodyanov:
| Before | After V1 | After V2 | After V3 |
| - | - | - | - |
| swappy-20251021_163840 | swappy-20251021_163710 | image| image |

I am right in assuming that an organizer cannot deny the event right?

Also for reference, in both screenshots the event had 1 organizer and 4 invitees

### DIFF
--- a/src/components/Editor/Invitees/InviteesList.vue
+++ b/src/components/Editor/Invitees/InviteesList.vue
@@ -8,8 +8,8 @@
 		<div v-if="showHeader" class="invitees-list__header">
 			<div class="invitees-list__header__title">
 				<AccountMultipleIcon :size="20" />
-				{{ t('calendar', 'Attendees') }}
-				{{ statusHeader }}
+				<span class="invitees-list__header__title__text">{{ t('calendar', 'Attendees') }}</span>
+				<NcCounterBubble :count="invitees.length + 1" />
 			</div>
 
 			<div v-if="!hideButtons" class="invitees-list-button-group">
@@ -37,6 +37,10 @@
 			</div>
 		</div>
 
+		<div class="invitees-list__subtitle">
+			{{ statusHeader }}
+		</div>
+
 		<InviteesListSearch
 			v-if="!isReadOnly && hasUserEmailAddress"
 			:already-invited-emails="alreadyInvitedEmails"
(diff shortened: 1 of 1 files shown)


ChristophWurst:
I still think that "Attendees 5 invited" sounds like invalid grammar, no?

Big :+1: on the 1 vs 2 more attendees line at the bottom

ChristophWurst:
> I am right in assuming that an organizer cannot deny the event right?

Something @SebastianKrupinski or an RFC study can answer :)

GVodyanov:
image

Looking like this now.

GVodyanov:
@nimishavijay Hey! This change is something me and Christoph discussed, we decided it made more sense to count the organizer as an attendee, do you think that is something that could work or do you have any better ideas? Thanks!

ChristophWurst:
> I still think that "Attendees 5 invited" sounds like invalid grammar, no?

@nimishavijay that is the other part where I'd appreciate input :)

SebastianKrupinski:
> Looking like this now.

That looks much better!

As per RFC for ORGANIZER:

```
(";" cnparam) / (";" dirparam) / (";" sentbyparam) / (";" languageparam) 
```
The PARTSTATUS parameter is not allowed

As per RFC for ATTENDEE:
```
(";" cutypeparam) / (";" memberparam) /
(";" roleparam) / (";" partstatparam) /
(";" rsvpparam) / (";" deltoparam) /
(";" delfromparam) / (";" sentbyparam) /
(";" cnparam) / (";" dirparam) /
(";" languageparam) /
```
The PARTSTATUS is allowed.

So to answer your RFC question, correct the organizer can not decline a event.

Which brings me to the question, why do we set the partstatus on the organizer and handle it on the front end

SebastianKrupinski:
> @nimishavijay Hey! This change is something me and Christoph discussed, we decided it made more sense to count the organizer as an attendee, do you think that is something that could work or do you have any better ideas? Thanks!

This is a matter of human logic vs language.

Technically the organizer is also an attendee as they attend the meeting (language)

But in human thinking terms this might be confusing, as an organizer I think of attendees as who else besides me is attending (human logic)

nimishavijay:
Definitely agreed on the grammar part, I think in the orginal specs the "Attendees" part was supposed to be bold (like a heading) and the rest was suppose to be regular font and text-maxcontrast, so it shouldn't look like one big heading. 

As for the wording change, I agree with @SebastianKrupinski that "4 invited" makes sense to the organizer here, but not to anyone else. Maybe the issue is that it says "invited" (because the organizer wouldn't invite themselves). Gcal says "5 guests, 1 yes, 4 awaiting" over 2 lines and this seems pretty clear to me. So maybe we can do 

```
Attendees (5)
1 confirmed, 4 awaiting response 
``` 

The number of attendees would be in a counter bubble, and the second line with the response details would be in text-maxcontrast.

GVodyanov:
Here ya go:

image

SebastianKrupinski (approved):

nimishavijay:
Looking great! Small alignment stuff:
- There can be a little less gap between the "Attendees" and the counter bubble
- The "1 confirmed, 2 awaiting response" can be aligned with the "Attendees" text

GVodyanov:
image

Here it is now!

PR #7622, opened 2025-10-23
Make the wording "Selected slot" translatable

Jerome-Herbinet:
In the following screenshot, French UI is used, but "Selected slot" remains in English. 

If I'm not mistaking, this PR should fix this.

Testing from reviewers needed.


SebastianKrupinski:
@Jerome-Herbinet Thank you!

PR #7644, opened 2025-10-30
fix: week day view issue with tentative status

SebastianKrupinski:
### Summary
- Added safe guards to prevent UI from crashing when elements are missing in the dom

### DIFF
--- a/src/fullcalendar/rendering/eventDidMount.js
+++ b/src/fullcalendar/rendering/eventDidMount.js
@@ -12,154 +12,170 @@
  * @param {Node} data.el The HTML element
  */
 export default function({ event, el }) {
-	if (el.classList.contains('fc-event-nc-alarms')) {
-		const notificationIcon = document.createElement('span')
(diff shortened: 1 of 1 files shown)


src/fullcalendar/rendering/eventDidMount.js
```diff
@@ -12,154 +12,171 @@
...
-			dotElement.style.minWidth = '10px'
-			dotElement.style.minHeight = '10px'
-		} else {
-			el.style.background = backgroundStyling
+			el.title = t('calendar', 'Your participation is tentative')
 		}
 
-		el.title = t('calendar', 'Your participation is tentative')
+	} catch (error) {
+		console.error('Error in eventDidMount:', error)
```
ChristophWurst: Log `event` too so the event to error association is clear

SebastianKrupinski: Done

Issue #7558, opened 2025-10-10
Calendar settings reorganization


kra-mo:
Reorganized settings using the new form components. See https://github.com/nextcloud/server/issues/55667

Clicking on "Attachments folder" should bring up a folder picker.

<img width="500" height="1218" alt="Image" src="https://github.com/user-attachments/assets/113d2596-bd3a-4782-beb4-0eb9e08b5afe" />

SebastianKrupinski:
Hi @kra-mo 

The design looks great!

I do have some recommendations:

- I would rename the CalDav section to General and move it to the top (this is probably the most used and should be first)
- I would move the import to the calendar setting modal, this would remove extra steps of selecting the calendar after selecting the file to import.

kra-mo:
> * I would rename the CalDav section to General and move it to the top (this is probably the most used and should be first)

Fair, it doesn't need to be inherently technical as a lot of clients call it something generic as well. I was going off of Files' "WebDAV" section, I think I will try to revise both to be a bit more generic.

> I would move the import to the calendar setting modal, this would remove extra steps of selecting the calendar after selecting the file to import.

You mean to move it into the "Edit calendar" dialog of each individual calendar?

SebastianKrupinski:
> You mean to move it into the "Edit calendar" dialog of each individual calendar?

yes

kra-mo:
I moved the CalDAV settings to General, although I still gave it its own subsection to group it and add some explanation.

ChristophWurst:
@GVodyanov fyi just clarified with @jancborchardt that this can be started despite the "draft" state. Components will be ready very soon.

ChristophWurst:
Adjusting status because work has been started

PR #7682, opened 2025-11-09
feat(Settings): redesign to new vue components

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/7558

The only difference from the design spec is the NcSelect not being grey and not having the label "inside" the border, but those require nextcloud vue changes.

Updated screenshots:
Screenshot From 2025-11-14 14-09-43
Screenshot From 2025-11-14 14-09-55
Screenshot From 2025-11-14 14-10-00

### DIFF
--- a/src/components/AppNavigation/Settings.vue
+++ b/src/components/AppNavigation/Settings.vue
@@ -17,92 +17,71 @@
 				id="app-settings-modal"
 				class="app-settings-modal"
 				:name="t('mail', 'Calendar settings')"
+				:legacy="false"
 				:show-navigation="true"
 				:additional-trap-elements="[]"
 				:open="showSettingsModal"
 				@update:open="(val) => showSettingsModal = val">
 				<NcAppSettingsSection
 					id="settings-modal-general"
-					class="calendar-settings-modal-section"
 					:name="t('calendar', 'General')">
-					<SettingsImportSection
-						class="settings-fieldset-interior-item"
+					<SettingsTimezoneSelect
 						:is-disabled="loadingCalendars" />
-					<NcButton
-						variant="tertiary"
-						target="_blank"
-						:href="availabilitySettingsUrl">
-						<template #icon>
-							<CalendarClockIcon :size="20" decorative />
-						</template>
-						<span class="button-content-with-icon">
+					<NcFormBox>
+						<NcFormBoxButton
+							target="_blank"
+							:href="availabilitySettingsUrl">
 							{{ $t('calendar', 'Availability settings') }}
-							<OpenInNewIcon :size="20" decorative />
-						</span>
-					</NcButton>
-					<h4>{{ t('calendar', 'CalDAV') }}</h4>
-					<span>{{ t('calendar', 'Access Nextcloud calendars from other apps and devices') }}</span>
-					<NcButton
-						variant="tertiary"
-						@click.prevent.stop="copyPrimaryCalDAV">
-						<template #icon>
-							<ContentCopy :size="20" decorative />
-						</template>
-						{{ $t('calendar', 'CalDAV URL') }}
-					</NcButton>
-					<NcButton
-						variant="tertiary"
-						@click.prevent.stop="copyAppleCalDAV">
-						<template #icon>
-							<ContentCopy :size="20" decorative />
-						</template>
-						{{ $t('calendar', 'Copy iOS/macOS CalDAV address') }}
-					</NcButton>
+						</NcFormBoxButton>
+					</NcFormBox>
+					<SettingsImportSection
+						:is-disabled="loadingCalendars" />
+					<NcFormGroup :label="t('calendar', 'CalDAV')" :description="t('calendar', 'Access Nextcloud calendars from other apps and devices')">
+						<NcFormBox>
+							<NcFormBoxCopyButton :label="t('calendar', 'CalDAV URL')" :value="primaryCalDAV" />
+							<NcFormBoxCopyButton :label="t('calendar', 'Server Address for iOS and macOS')" :value="appleCalDAV" />
+						</NcFormBox>
+					</NcFormGroup>
 				</NcAppSettingsSection>
 				<NcAppSettingsSection
-					id="app-settings-modal-view"
-					class="calendar-settings-modal-section"
+					id="settings-modal-appearance"
 					:name="t('calendar', 'Appearance')">
-					<SettingsTimezoneSelect
-						class="settings-fieldset-interior-item"
-						:is-disabled="loadingCalendars" />
-					<NcCheckboxRadioSwitch
-						class="settings-fieldset-interior-item"
-						:model-value="hasBirthdayCalendar"
-						:disabled="isBirthdayCalendarDisabled"
-						@update:model-value="toggleBirthdayEnabled">
-						{{ $t('calendar', 'Birthday calendar') }}
-					</NcCheckboxRadioSwitch>
-					<NcCheckboxRadioSwitch
-						class="settings-fieldset-interior-item"
-						:model-value="showTasks"
-						:disabled="savingTasks"
-						@update:model-value="toggleTasksEnabled">
-						{{ $t('calendar', 'Tasks in calendar') }}
-					</NcCheckboxRadioSwitch>
-					<NcCheckboxRadioSwitch
-						class="settings-fieldset-interior-item"
-						:model-value="showWeekends"
-						:disabled="savingWeekend"
-						@update:model-value="toggleWeekendsEnabled">
-						{{ $t('calendar', 'Weekends') }}
-					</NcCheckboxRadioSwitch>
-					<NcCheckboxRadioSwitch
-						class="settings-fieldset-interior-item"
-						:model-value="showWeekNumbers"
-						:disabled="savingWeekNumber"
-						@update:model-value="toggleWeekNumberEnabled">
-						{{ $t('calendar', 'Week numbers') }}
-					</NcCheckboxRadioSwitch>
-					<NcCheckboxRadioSwitch
-						class="settings-fieldset-interior-item"
-						:model-value="eventLimit"
-						:disabled="savingEventLimit"
-						@update:model-value="toggleEventLimitEnabled">
-						<span class="no-wrap-label">
+					<NcFormBox>
+						<NcFormBoxSwitch
+							v-model="hasBirthdayCalendarBinding"
+							:disabled="isBirthdayCalendarDisabled"
+							@update:modelValue="toggleBirthdayEnabled">
+							{{ $t('calendar', 'Birthday calendar') }}
+						</NcFormBoxSwitch>
+						<NcFormBoxSwitch
+							v-model="showTasksBinding"
+							:disabled="savingTasks"
+							@update:modelValue="toggleTasksEnabled">
+							{{ $t('calendar', 'Tasks in calendar') }}
+						</NcFormBoxSwitch>
+						<NcFormBoxSwitch
+							v-model="showWeekendsBinding"
+							:disabled="savingWeekend"
+							@update:modelValue="toggleWeekendsEnabled">
+							{{ $t('calendar', 'Weekends') }}
+						</NcFormBoxSwitch>
+						<NcFormBoxSwitch
+							v-model="showWeekNumbersBinding"
+							:disabled="savingWeekNumber"
+							@update:modelValue="toggleWeekNumberEnabled">
+							{{ $t('calendar', 'Week numbers') }}
+						</NcFormBoxSwitch>
+					</NcFormBox>
+
+					<NcFormBox>
+						<NcFormBoxSwitch
+							v-model="eventLimitBinding"
+							:disabled="savingEventLimit"
+							@update:modelValue="toggleEventLimitEnabled">
 							{{ $t('calendar', 'Limit number of events shown in Month view') }}
-						</span>
-					</NcCheckboxRadioSwitch>
+						</NcFormBoxSwitch>
+					</NcFormBox>
+
 					<NcSelect
 						:id="slotDuration"
 						:options="slotDurationOptions"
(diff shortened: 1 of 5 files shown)


SebastianKrupinski:
Looks great! Is this ready for testing?

Can you also rebase to the latest calendar version, it will fix some of the CI errors

kra-mo:
Just a small note before reviewing the whole thing: I updated the mockup in the issue to use the (future) [NcFormBoxSelectNative](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7838) component for the default reminder and density.

GVodyanov:
> Looks great! Is this ready for testing?
> 
> Can you also rebase to the latest calendar version, it will fix some of the CI errors

Yep, it should work... CI tho...

SebastianKrupinski:
Looking good so far!

But there are a couple glitches

A, Is everything supposed to be highlighted in dark mode?
B. Switch states are not changing when switched (look at Birthday Calendar and Task Calendar)

image

kra-mo:
> A, Is everything supposed to be highlighted in dark mode?

This is the styling we went with for the components themselves, but indeed, dark mode was definitely not thought out very much and should probably be adjusted.

kra-mo:
I opened https://github.com/nextcloud-libraries/nextcloud-vue/issues/7853

kra-mo:
> B. Switch states are not changing when switched (look at Birthday Calendar and Task Calendar)

And I think this one should be fixed by https://github.com/nextcloud-libraries/nextcloud-vue/pull/7851

src/components/AppNavigation/Settings.vue
```diff
@@ -23,120 +23,98 @@
 				@update:open="(val) => showSettingsModal = val">
 				<NcAppSettingsSection
 					id="settings-modal-general"
-					class="calendar-settings-modal-section"
 					:name="t('calendar', 'General')">
```
ShGKme, line 27: `:legacy="false"` is required to apply the new design, including:
- Heading padding
- Gap between elements inside the section

ShGKme: P.S. Applied by default in `v9` (Vue 3)

src/components/AppNavigation/Settings.vue
```diff
@@ -23,120 +23,98 @@
...
-					</NcButton>
-					<NcButton
-						variant="tertiary"
-						@click.prevent.stop="copyAppleCalDAV">
-						<template #icon>
-							<ContentCopy :size="20" decorative />
-						</template>
-						{{ $t('calendar', 'Copy iOS/macOS CalDAV address') }}
-					</NcButton>
+					<div class="calendar-settings-modal-section">
```
ShGKme: With the new design applied, wrapping groups/boxes into additional element breaks default `gap`

src/components/AppNavigation/Settings.vue
```diff
@@ -23,120 +23,98 @@
...
-						:id="slotDuration"
-						:options="slotDurationOptions"
-						:value="selectedDurationOption"
-						:disabled="savingSlotDuration"
-						:clearable="false"
-						:input-label="$t('calendar', 'Density in Day and Week View')"
-						input-id="value"
-						label="label"
-						@option:selected="changeSlotDuration" />
+						</NcFormBoxSwitch>
```
ShGKme: It technically works, but per documentation as well as per component name, `NcFormBoxSwitch` is supposed to be used inside `NcFormBox` as it a part of the box - a box item. Even though it doesn't currently change anything for a single item boxes.

src/components/AppNavigation/Settings.vue
ShGKme: Sorry, wrong plaсе for the comment. I meant, `<NcAppSettingsDialog>` component, not every single section...

GVodyanov:
> A, Is everything supposed to be highlighted in dark mode? 

[This](https://github.com/nextcloud-libraries/nextcloud-vue/issues/7853) should fix it, it's a Nextcloud vue thing

> B. Switch states are not changing when switched (look at Birthday Calendar and Task Calendar)

Fixed!, will push in a second

GVodyanov:
@kra-mo 

> The keyboard shortcuts should use the new component as well. It's fine if it will be in a followup PR, but it should be done too.

These are using `NcKbd`, isn't that the new component?

src/components/AppNavigation/Settings.vue
GVodyanov: I see, how would you recommend adding gaps then?

For example, there is supposed to be a gap between the "Week numbers" and "Limit number of events". Each of the sections are wrapped in an `NcFormBox`, and I assume the idea is that the two together should be part of an `NcFormGroup`, but if I use an `NcFormGroup`, it requires adding a title. This would result in having two titles, seeing as I already have the `NcAppSettingsSection` I am supposed to use for the scrolling to work.

ShGKme:
> These are using `NcKbd`, isn't that the new component?

See: https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppContent/NcAppSettings?id=ncappsettingsshortcutssection

Inside there is `<NcKbd>`, but new components include complete hotkey list with even a special AppSettings dialog section

src/components/AppNavigation/Settings.vue
ShGKme: > For example, there is supposed to be a gap between the "Week numbers" and "Limit number of events". Each of the sections are wrapped in an NcFormBox, and I assume the idea is that the two together should be part of an NcFormGroup

No, they only need to be wrapped into `NcFormGroup` if it is a labelled group.

Otherwise you can just have to `NcFormBox` inside the `<NcAppSettingsSection>`. The app settings section has the gap.

ShGKme: So in your case it is:
```
<NcAppSettingsSection name="Appearence">
  <SettingsTimezoneSelect />
  
  <!-- gap -->
  
  <NcFormbox>
    Birthday calendar
    ...
    WeekNumbers
  </NcFormbox>

  <!-- gap -->

  <NcFormBox>
    Limit number
  </NcFormBox>
  
  <!-- gap -->
  
  <NcSelect />
</NcAppSettingsSection>
```

Each `<!-- gap -->` is already added by the `NcAppSettingsSection` if you disabled the legacy mode.

src/components/AppNavigation/Settings/ShortcutOverview.vue
```diff
@@ -8,17 +8,17 @@
...
+			</h6>
 			<div
 				v-for="(shortcut, index) in category.shortcuts"
 				:key="`${category.categoryId}-${index}`"
 				class="shortcut-section-item">
 				<span class="shortcut-section-item__keys">
 					<template v-for="(keyCombination, index2) of shortcut.keys">
 						<template v-for="(key, index3) in keyCombination">
-							<kbd :key="`${category.categoryId}-${index}-${index2}-${index3}`">{{ key }}</kbd>
+							<NcKbd :key="`${category.categoryId}-${index}-${index2}-${index3}`">{{ key }}</NcKbd>
```
ShGKme: In addition to using `NcAppSettingsShortcurtsSection`, `<NcKbd>` has the slot only for special cases (advanced usage). But the preferred way is to use the `symbol` prop.

It has a lot of special handling that otherwise must be handled by the app:
- Translating key names like `Ctrl` to `Strg` in German
- Using macOS names like `Option` instead of `Alt`
- Using `Command` instead of `Ctrl` on macOS
- Showing symbols from `↑` for ArrowUp to `⌥` for Alt (Option) on macOS

https://nextcloud-vue-components.netlify.app/#/Components/NcKbd

src/store/settings.js
```diff
@@ -84,6 +84,7 @@ export default defineStore('settings', {
 		 * @return {Promise<void>}
 		 */
 		async toggleBirthdayCalendarEnabled() {
+			console.log('we doing stuff')
```
ShGKme: *me: doing stuff*

src/components/AppNavigation/Settings/ShortcutOverview.vue
GVodyanov: I see, I completely redid the shortcut sections now with `NcAppSettingsShortcurtsSection`, `NcHotkeyList`, `NcHotKey` and `NcKbd` when needed.

src/components/AppNavigation/Settings.vue
```diff
@@ -17,92 +17,76 @@
...
+							:disabled="savingWeekend"
+							@enable="toggleWeekendsEnabled"
+							@disable="toggleWeekendsEnabled">
+							{{ $t('calendar', 'Weekends') }}
+						</NcFormBoxSwitch>
+						<NcFormBoxSwitch
+							v-model="showWeekNumbersBinding"
+							:disabled="savingWeekNumber"
+							@enable="toggleWeekNumberEnabled"
+							@disable="toggleWeekNumberEnabled">
```
ShGKme: If you handle `enable` and `disable` the same way, you probably need a general `@update:model-value`.

Or, since you have `v-model` already, maybe even just `watch`/`watchEffect`.

src/components/AppNavigation/Settings.vue
```diff
@@ -116,18 +100,28 @@
...
+					<NcSelect
+						:options="defaultReminderOptions"
+						:value="selectedDefaultReminderOption"
+						:disabled="savingDefaultReminder"
+						:clearable="false"
+						:input-label="$t('calendar', 'Default reminder')"
+						input-id="value"
+						label="label"
+						@option:selected="changeDefaultReminder" />
+					<NcFormBoxSwitch
```
ShGKme: This one is also not in `<NcFormBox>` 👀

src/components/AppNavigation/Settings.vue
```diff
@@ -372,6 +355,23 @@ export default {
...
+		},
+	},
+
+	async mounted() {
+		this.hasBirthdayCalendarBinding = this.hasBirthdayCalendar
+		this.showTasksBinding = this.showTasks
+		this.showWeekendsBinding = this.showWeekends
+		this.showWeekNumbersBinding = this.showWeekNumbers
+		this.eventLimitBinding = this.eventLimit
+		this.simpleEventEditorBinding = !this.skipPopover
```
ShGKme: A general note: you most likely don't need `mounted` here. `Mounted` lifecycle event happens when the component is already created, initialised, rendered and mounted to the DOM with the initial state.

Changing the state in `mounted` means that right after the component was rendered with the wrong state, we change the state to the correct one and render again.

If wouldn't make a difference if there were some async operation, like getting the state from the backend.

For setting the initial state, you can do it directly in the `data()` on init, or in in `created`/`beforeMount` if it's something more complex.

ShGKme: Though, most of the assignments seem redundant as both values are `false` 👀

src/components/AppNavigation/Settings.vue
```diff
@@ -575,4 +545,13 @@ export default {
...
 }
+
+.app-settings__content {
+	margin: auto;
+}
+
+.app-settings-modal .app-settings__content {
+	box-sizing: border-box;
+	max-height: 100%;
+}
```
ShGKme: **issue(blocking)**: please, always use `scoped` styles and ever override global styles. Otherwise `calendar` app changes the settings dialog for all the other apps.

ShGKme: `box-sizing` is there already: https://github.com/nextcloud-libraries/nextcloud-vue/blob/stable8/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue#L270

ShGKme: Why do you need `margin: auto`?

kra-mo:
@GVodyanov could you please update the screenshots?

src/components/AppNavigation/Settings/SettingsAttachmentsFolder.vue
```diff
@@ -43,8 +37,8 @@ export default {
 			attachmentsFolder: (store) => store.attachmentsFolder || '/',
 		}),
 
-		inputId() {
-			return `input-${randomId()}`
+		folderIcon() {
+			return mdiFolderOpenOutline
```
ShGKme: nitpik: computed only makes sense for reactive value computing over reactive variables. `mdiFolderOpenOutline` is a constant without any computing and without reactivity.

src/components/AppNavigation/Settings/ShortcutOverview.vue
```diff
@@ -3,105 +3,113 @@
...
+	},
+
 	computed: {
 		shortcuts() {
 			return [{
 				categoryId: 'navigation',
 				categoryLabel: t('calendar', 'Navigation'),
 				shortcuts: [{
-					keys: [['p'], ['k']],
+					keys: ['p', 'k'],
```
ShGKme: Not important for the component, but in the design draft it was uppercase

src/components/AppNavigation/Settings/ShortcutOverview.vue
```diff
@@ -3,105 +3,113 @@
...
-					keys: [['Ctrl+Enter']],
+					keys: ['Ctrl Enter'],
 					label: t('calendar', 'Save edited event'),
 				}, {
-					keys: [['Ctrl+Delete']],
+					keys: ['Ctrl Delete'],
 					label: t('calendar', 'Delete edited event'),
 				}, {
-					keys: [['Ctrl+d']],
+					keys: ['Ctrl d'],
```
ShGKme: Formal name if `Control`, not `Ctrl`. Otherwise it is not translated and not macOS aware (on macOS hotkey is supposed to be with `Command`)

src/components/AppNavigation/Settings.vue
ShGKme: Ahh, `border-box` is not for `*`. Let me fix.

GVodyanov:
> @GVodyanov could you please update the screenshots?

Updated!

src/components/AppNavigation/Settings.vue
GVodyanov: `@update:model-value` does not work in this case, it doesn't trigger the `toggle` method.

Sure, a `watch` could work but wouldn't that be more verbose?

src/components/AppNavigation/Settings.vue
GVodyanov: So afaik it's not possible to do it in `data` seeing as we are assigning values from computed properties. The values depend on the settings store.

Moving it to `created` is a good idea though!

GVodyanov:
Here ya go:

image

@kra-mo @ShGKme

src/components/AppNavigation/Settings.vue
ShGKme: Could you try `@update:modelValue`?

src/components/AppNavigation/Settings.vue
```diff
@@ -116,18 +100,30 @@
...
+							@enable="togglePopoverEnabled"
+							@disable="togglePopoverEnabled">
+							<template #description>
+								{{ $t('calendar', '"More details" opens the detailed editor') }}
+							</template>
+						</NcFormBoxSwitch>
+					</NcFormBox>
 					<!-- Hidden: default calendar picker -->
-					<div v-if="false" class="settings-fieldset-interior-item">
+					<div v-if="false">
```
ShGKme: Does it actually work and do anything?
With `v-if="false"` the `CalendarPicker` instance is never created.

src/components/AppNavigation/Settings/SettingsAttachmentsFolder.vue
```diff
@@ -4,48 +4,45 @@
...
 	components: {
-		NcInputField,
+		NcFormBoxButton,
+		NcIconSvgWrapper,
+	},
+
+	data: function() {
+		return {
+			folderIcon: mdiFolderOpenOutline,
+		}
```
ShGKme: **nitpick**: The icon is also not a part of the component's reactive state (the component never changes the `folderIcon` state property). And renaming it from `mdiFolderOpenOutline` to `folderIcon` makes it harder to work with the component template, because the same icon has different name in the template and originally.

Also `data: function()` is 10-year-old JS syntax :P

src/components/AppNavigation/Settings.vue
GVodyanov: Yes! That works

src/components/AppNavigation/Settings.vue
GVodyanov: Well... No. It definitely doesn't do anything. I guess I didn't want to remove it because I didn't know the reason why it was there, but oh well, guess it will be lost to git history 🫡

SebastianKrupinski: It was a setting to set a users default calendar to be used by the backend, I actual not sure why it was commented out. I believe is was @st3iny if I remember correctly. 

You can remove it.

src/components/AppNavigation/Settings/SettingsAttachmentsFolder.vue
GVodyanov: Made it normal lol, didn't realize `NcIconSvgWrapper` wasn't necessary anymore

src/components/AppNavigation/Settings/SettingsAttachmentsFolder.vue
```diff
@@ -4,48 +4,46 @@
...
 	components: {
-		NcInputField,
+		NcFormBoxButton,
+		IconFolderOpen,
+	},
+
+	data: function() {
+		return {
+			folderIcon: mdiFolderOpenOutline,
+		}
```
ShGKme: You still have data :P

GVodyanov: *Sigh*... Sorry

ShGKme (approved):
Still approved

SebastianKrupinski (approved):

SebastianKrupinski:
Look good, Merge it when you're ready

ShGKme:
🎉

PR #7683, opened 2025-11-09
fix: meeting proposal calendar blockers

SebastianKrupinski:
### Summary
- modified the blocker event to include a prefix '[Proposed]' to identify the blockers better
- modified event creation logic to use RDATE for dates
- modified organizer calendar CUD logic to properly track meeting proposal actions
- modified participant calendar CUD logic to properly track meeting proposal actions

### Testing
- create a proposal with multiple time dates
- modify a proposal by changing one or more of the dates
- delete a proposal
- verify that the time blocks reflect that proposal dats

### DIFF
--- a/lib/Service/Proposal/ProposalService.php
+++ b/lib/Service/Proposal/ProposalService.php
@@ -25,6 +25,7 @@ use OCA\Calendar\Objects\Proposal\ProposalParticipantRealm;
 use OCA\Calendar\Objects\Proposal\ProposalParticipantStatus;
 use OCA\Calendar\Objects\Proposal\ProposalResponseObject;
 use OCA\Calendar\Objects\Proposal\ProposalVoteCollection;
+use OCA\DAV\CalDAV\InvitationResponse\InvitationResponseServer;
 use OCP\Calendar\ICalendar;
 use OCP\Calendar\ICalendarIsWritable;
 use OCP\Calendar\ICreateFromString;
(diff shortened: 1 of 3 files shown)


GVodyanov (commented):
Currently testing and trying to make sense of this:

I have 4 meeting proposals 

Image

Only 2 of these are showing up in my calendar.

Image

next week:

Image

In the availability view only the proposal on Tuesday shows up:

Image

Also when I open the meetings proposal modal, change times by dragging, click update, I get a "Succesfully saved proposal" toast but when I reopen the modal it actually didn't change anything.

SebastianKrupinski:
I just tested again, it is working properly but!

> In the availability view only the proposal on Tuesday shows up:

Ugh, you found a bug that is unrelated to this PR. If I make a "proposed" meeting on 3 separate weeks, it does not show up in "Week" view but shows correctly in "month" view. It's a "RDATE" handling bug, in the time span calculation we use to select events from the database.

image

See the first and last occurrence field are the same value. This PR will fix that.

https://github.com/nextcloud/server/pull/48131

Download the event and check if your event has a RDATE with the correct selected proposal dates.

SebastianKrupinski:
Found the bug causing the events not to show properly.

3rdparty/sabre/vobject/lib/Component/VEvent.php::isInTimeRange() line 32 disregards RDATE.

image

SebastianKrupinski:
Fix for RDATES not showing properly

https://github.com/sabre-io/vobject/pull/735

SebastianKrupinski:
@GVodyanov 

Can you retest again, see the above comments on what you found last time

SebastianKrupinski:
> However I did have to actually delete and then recreate the proposal, being able to drag it would be a nice feature

Re-saving the proposal would have updated the calendar blockers.

> Also I only saw the first of the proposals as a blocking event, I'm not sure if this is a bug or a feature.

You would need to apply the sabredav patch from my comment above. then it will properly show all the blockers

PR #7685, opened 2025-11-11
fix: full calendar error logging

SebastianKrupinski:
### Summary
- Added error catching wrappers to full calendar hooks

### Reason
Full calendar catches any errors in hooks to prevent a break down in the core engine


GVodyanov (commented):
I tested it, and at it does not seem to break anything, but could you please suggest how to recreate a fullcalendar error to test the logging?

SebastianKrupinski:
> I tested it, and at it does not seem to break anything, but could you please suggest how to recreate a fullcalendar error to test the logging?

Add a "Throw error" at the begging of the eventDidMount.js

image

GVodyanov (approved):
Oke, tested and makes sense

Issue #7684, opened 2025-11-11
Migrate to Vue 3


SebastianKrupinski:
### Summary
- Migrate to Vue 3 and Nextcloud/vue 9

Issue #6788, opened 2025-03-06
Wrong date and time formats with Firefox


ChristophWurst:
### Steps to reproduce

1. Set your system and browser to language=German and locale=Germany
2. Open the Calendar app with Firefox desktop
3. Create a new event

### Expected behavior

Date shows DD-MM-YYYY and time in 24h format

### Actual behavior

Date shows MM-DD-YYYY and time in 12h format

### Calendar app version

7.x

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

With https://github.com/nextcloud/calendar/pull/6423 we replaced the custom (and buggy) datepicker with a native one. The objectives were usability and accessibility. Most platforms come with a very capable date and time picker. Firefox only has a date picker, which was considered an acceptable trade-off, because keyboard input is always possible.

This is an **upstream issue** with Firefox. I'm filing the ticket here for transparency and to have a central point for updates/discussion.

Relevant Firefox ticket(s):
* *Poor support for datetime-local (no time, and null-value if time not selected)* https://bugzilla.mozilla.org/show_bug.cgi?id=1850383

You can also test the datetime picker at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local. It works well with Chrome but not with Firefox. The problem does not appear to be specific to Firefox.

tmh-grunwald-markus:
I'm using Chromium Version 133.0.6943.126 (Official Build) built on Debian GNU/Linux 12 (bookworm) (64-bit) and have exactly the same error.

ghnp5:
Came here to report the same :)

Just set my locale to "Portuguese (Portugal)"  (pt_PT), and the date format for new events is still MM/DD/YYYY, instead of DD/MM/YYYY.

Thanks!

GK-774:
It´s not just Firefox, same effect with Edge. Setting the Locale (German (Austria) in my case) just does nothing, the date format remains at MM/DD/YYYY as already mentioned by ghnp5. Even in the profile the birth date remains at MM/DD/YYYY when changing the locale. There should be a setting for the date and time format independent of the locale anyway.

ChristophWurst:
Thank you for your feedback.

astehlik:
[...]

PR #7697, opened 2025-11-15
feat: migrate to vue 3

SebastianKrupinski:
### Summary
- Resolves: https://github.com/nextcloud/calendar/issues/7684
- Migrate to vue 3

### Testing
- checkout branch
- delete existing node_modules directory
- npm ci && npm run dev


SebastianKrupinski:
Initial testable vue 3 migration

GVodyanov:
Fixed this:
| Before | After |
| - | - |
| image | image |

GVodyanov:
I added the new datepicker, however it requires this: https://github.com/nextcloud-libraries/nextcloud-vue/pull/8009

DerDreschner:
Ohh, and the overlay for picking an existing file to attach to an event has the same issue with the "pick" button not having any borders/styling.

<details>

<summary>Example picture</summary>

Image

</details>

SebastianKrupinski:
Thanks for testing @DerDreschner !

The styling issues will be addressed in a few follow up PR's as per our team call conversation. There is also one issue that seems to be caused by a broken upstream component.

### Style issues
- When editing the first notification of an event, the menu button sticks out into the drop-down selection field:
- The input fields are not on the same line of sight here:
- Ohh, and the overlay for picking an existing file to attach to an event has the same issue with the "pick" button not having any borders/styling.

### New to vue 9
- When opening "Add public holiday calendar" and "Repeat event", there is a new text in the top middle? But only in those cases? Don't know if that's desired, it irrated me as that are the only two cases an modal overlay creates that.

### Fixed
- Creating a new event, editing an existing one or viewing a read-only one doesn't show the option to select/show the calendar the event is being created on. For that reason, the buttons at the top right side stick out into the title field when using the quick view:
- Adding a file to an event (both upload and select existing file)
- Editing an reminder (only closes when removing reminder -> no more reminders there afterwards)

### Broken upstream
- Creating a new calendar (only exception: add public holiday calendar)

wolfgang101:
@SebastianKrupinski I saw your mention in #6788 and can't wait for this to make it into the official release.

Are we allowed to give additional feedback here?

I tested this pull request and two minor things stood out:

- the two time fields use the wrong icon - previously it was the clock icon which has now changed to the calendar date icon that is also used in the two date fields
- [simple event editor only]: when clicking into the date and time input fields the date / time selectors are centered right below the input fields. the time selector container gets cut off to the right. on smaller screen sizes the date selector container gets cut off to the left and the time selector to the right. as far as I could see it's just cosmetic and does not hinder any input. But maybe there are edge cases in other locales where this could be a problem? Only tested with de_DE to see if the new selectors work - they work great BTW.

SebastianKrupinski:
> Are we allowed to give additional feedback here?

Of course! This is why we make these public.

>     * the two time fields use the wrong icon - previously it was the clock icon which has now changed to the calendar date icon that is also used in the two date fields
>     * [simple event editor only]: when clicking into the date and time input fields the date / time selectors are centered right below the input fields. the time selector container gets cut off to the right. on smaller screen sizes the date selector container gets cut off to the left and the time selector to the right. as far as I could see it's just cosmetic and does not hinder any input. But maybe there are edge cases in other locales where this could be a problem? Only tested with de_DE to see if the new selectors work - they work great BTW.

Minor styling issues will be fixed in follow up PR's, once this is merged we'll get a design review of issues to be fixed.

Thank you for testing!!!

GVodyanov:
image

Calendar settings is showing timezone "floating"

GVodyanov:
Fixing rn

GVodyanov:
image

Fixed

GVodyanov:
Hmm also `NcFormBoxCopyButton` doesn't copy anything but that sounds like a nextcloud vue issue

GVodyanov:
image

This selector in appointment settings can't be changed

GVodyanov:
image

Also doesn't work here

GVodyanov:
image

"can edit and see confidential events" checkbox doesn't work

SebastianKrupinski:
6 months and alot of grit but its done

tcitworld:
Congrats!

louwers:
Awesome work.

Issue #7711, opened 2025-11-19
Appointments: Add 20 and 25 minutes to increment select field when creating schedules


st3iny:
### Is your feature request related to a problem? Please describe.

It would be great if I could select 20 and 25 minutes as increments when creating a new appointment schedule.

<img width="290" height="333" alt="Image" src="https://github.com/user-attachments/assets/0ece3e0c-8b48-4a34-9515-2b2c68389e3b" />

### Describe the solution you'd like

Add 20 and 25 minutes to the appointment schedule interval select field.

SebastianKrupinski:
Thank you for the request :)

More selectable quick pick options is a good idea

SebastianKrupinski:
Fixed. Should be out in the next release

PR #7718, opened 2025-11-21
fix: appointment duration selection

SebastianKrupinski:
### Summary
- Resolves https://github.com/nextcloud/calendar/issues/7711
- Added more minute increments to duration selection

image


GVodyanov (approved):
Imagine getting invited to like a 25 minute date

Issue #7706, opened 2025-11-18
After upgrade to 6.10 I can no longer change the colour of events


engmex:
### Steps to reproduce

1.  Edit an existing event or create a new one.
2. click on the event colour box 
3. Select a new colour and click choose
4. event does not change to new colour
5. If the event is any other colour than the default it's possible to click the reload button to change the event to the default colour.  Once it is the default colour it's impossible to change

### Expected behavior

As stated in the steps to reproduce I'm no longer able change the colour of events after upgrading the calendar to 6.10.  Previously it worked fine.  Nextcloud Hub 25 Autumn (32.0.1)

### Actual behavior

Nothing happens

### Calendar app version

6.10

### CalDAV-clients used

 DAVx5,Thunderbird 

### Browser

vivaldi 

### Client operating system

Debian 13

### Server operating system

Debian 13

### Web server

Apache

### Database engine version

MariaDB

### PHP engine version

PHP 8.4

### Nextcloud version

Nextcloud Hub 25 Autumn (32.0.1)

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

jannikbx:
same problem. Pls fix!

SebastianKrupinski:
Confirmed

blip-unicorn:
I'm seeing the same problem on 32.0.1 and calendar version 6.1.0 using Firefox or Chromium on Debian 13. In the browser console there is a warning each time the event editor is opened "[WARN] @nextcloud/vue: You need to fill either the text or the ariaLabel props in the button component. " which might be related.

SebastianKrupinski:
Fix is merged and should be available with the next release

blip-unicorn:
Today my calendar updated to 6.1.1 and the issue is now fixed for me.

Thank you to all involved!

jannikbx:
thx for fix!

SebastianKrupinski:
You're welcome.

PR #7719, opened 2025-11-22
fix: event color picker

SebastianKrupinski:
### Summary
- Resolves https://github.com/nextcloud/calendar/issues/7706
- Updated event color picker logic to use v-modal

### DIFF
--- a/src/components/Editor/Properties/PropertyColor.vue
+++ b/src/components/Editor/Properties/PropertyColor.vue
@@ -24,23 +24,24 @@
 		<div
 			v-else
 			class="property-color__input">
-			<ColorPicker
-				:value="selectedColor"
-				:shown.sync="isColorPickerOpen"
+			<NcColorPicker
+				v-model="selectedColor"
+				:shown.sync="selectorOpen"
 				:advanced-fields="true"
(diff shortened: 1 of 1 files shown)


src/components/Editor/Properties/PropertyColor.vue
```diff
@@ -24,23 +24,23 @@
...
 					class="property-color__color-preview"
 					:style="{ 'background-color': selectedColor }" />
-			</ColorPicker>
-			<Actions v-if="showColorRevertButton">
-				<ActionButton @click.prevent.stop="deleteColor">
+			</NcColorPicker>
+			<NcButton
+				v-if="!isReadOnly && !!value"
+				:arial-label="$t('calendar', 'Remove color')"
+				@click="deleteColor">
```
GVodyanov, line 40: Not a huge fan of the button having a background color

```suggestion
				v-if="!isReadOnly && !!value"
				variant="tertiary"
				:arial-label="$t('calendar', 'Remove color')"
				@click="deleteColor">
```

SebastianKrupinski: Done

PR #7720, opened 2025-11-24
fix: settings getCurrentUserPrincipal error

SebastianKrupinski:
### Summary
- Fixes Error below. Caused by the dav stack not being initialized on initial render.

932-459-max

[ERROR] calendar: [Vue error]: Error in render: TypeError: can't access property "principalUrl", _services_caldavService_js__WEBPACK_IMPORTED_MODULE_11__.getCurrentUserPrincipal() is null 
Object { app: "calendar", uid: "user1", level: 0, error: TypeError, vm: {…}, info: "render" }
index.mjs:54:1
    log index.mjs:54
    error index.mjs:72
    errorHandler main.js:42
    VueJS 89
    <anonymous> main.js:49
    <anonymous> calendar-main.js:309625
    <anonymous> calendar-main.js:309627

### DIFF
--- a/src/components/AppNavigation/Settings.vue
+++ b/src/components/AppNavigation/Settings.vue
@@ -345,7 +345,10 @@ export default {
 		},
 
 		appleCalDAV() {
-			return new URL(getCurrentUserPrincipal().principalUrl, this.primaryCalDAV)
+			if (!this.currentUserPrincipal) {
+				return ''
+			}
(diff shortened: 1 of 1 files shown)


zipp0akazip:
Hello, it's will be great to see new version with this fix as soon as possible. We updated to v6.1.0 and have no possibility to downgrade. Thx.

PR #7721, opened 2025-11-25
fix(PropertyTitleTimePicker): improve styling for local time subtitle

GVodyanov:
| Before | After |
| - | - |
| swappy-20251125_170550 | swappy-20251125_170300 |
| swappy-20251125_170538 | swappy-20251125_170446 |

### DIFF
--- a/src/components/Editor/Properties/PropertyTitleTimePicker.vue
+++ b/src/components/Editor/Properties/PropertyTitleTimePicker.vue
@@ -94,16 +94,18 @@
 			v-if="isReadOnly"
 			class="property-title-time-picker__time-pickers property-title-time-picker__time-pickers--readonly">
 			<div class="property-title-time-picker-read-only-wrapper property-title-time-picker-read-only-wrapper--start-date">
-				<div class="property-title-time-picker-read-only-wrapper__label">
-					{{ formattedStart }}
-				</div>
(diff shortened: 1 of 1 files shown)


SebastianKrupinski:
Looks good

src/components/Editor/Properties/PropertyTitleTimePicker.vue
```diff
@@ -636,7 +641,19 @@ export default {
...
+	gap: calc(var(--default-grid-baseline) * 2);
+}
+
 .property-title-time-picker-read-only-wrapper__subtitle {
     opacity: 0.8;
 }
+
+.app-full {
+	.property-title-time-picker--readonly {
+		margin-left: calc(var(--default-grid-baseline) * 9);
```
ChristophWurst: Should it be `margin-inline-start` to also work with right-to-left @hamza221?

hamza221: Yes, stylelint should be able to fix it automatically

hamza221: >` 655:3  ✖  Unexpected "margin-left" property. Use "margin-inline-start".  csstools/use-logical`

PR #7733, opened 2025-11-30
fix(talkService): make attendees be synced with talk room members

GVodyanov:
Fixes the following bug:
- have a talk room
- add this talk room to an event, and also an internal user that is not in the room yet, user confirms event participation
- expected: user gets automatically added to the room
- what happened: user not in the room, had to add him by hand

The issue was the `/` becoming encoded as a `%2F`

### DIFF
--- a/src/services/talkService.ts
+++ b/src/services/talkService.ts
@@ -336,7 +336,7 @@ export async function updateRoomParticipantsFromEvent(eventComponent: object): P
  */
 async function transceiveGet<TRequest extends object | undefined, TResponse>(path: string, params?: TRequest): Promise<TResponse> {
 	const apiVersion = loadState('calendar', 'talk_api_version')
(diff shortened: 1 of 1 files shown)


SebastianKrupinski (approved):

SebastianKrupinski:
If the issue was the / getting encoded, did you check if the other paths are okay? Like "room/${token}/participants"??

GVodyanov:
> If the issue was the / getting encoded, did you check if the other paths are okay? Like "room/${token}/participants"??

Where is that being used? @SebastianKrupinski

SebastianKrupinski:
> > If the issue was the / getting encoded, did you check if the other paths are okay? Like "room/${token}/participants"??
> 
> Where is that being used? @SebastianKrupinski

During event create or update, in the talkService, first we create a room then we update the participants on save, but first we pull the existing participants

PR #7794, opened 2025-12-23
feat: restrict calendar invitation participants

SebastianKrupinski:
### Summary
- Added functionality to restrict inviting only system users
- Removed calls to the DAV backend the contacts controller already checks all available dav address books

### DIFF
--- a/lib/Controller/ContactController.php
+++ b/lib/Controller/ContactController.php
@@ -18,6 +18,7 @@ use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\QueryException;
 use OCP\Contacts\IManager;
+use OCP\IAppConfig;
 use OCP\IRequest;
 use OCP\IUserManager;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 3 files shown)


GVodyanov (approved):

GVodyanov:
Is this something we want added to settings too?

lib/Controller/ContactController.php
```diff
@@ -100,11 +102,16 @@ public function searchAttendee(string $search):JSONResponse {
 			return new JSONResponse();
 		}
 
-		$result = $this->contactsManager->search($search, ['FN', 'EMAIL'], ['enumeration' => false]);
+		$externalAttendeesDisabled = $this->appConfig->getValueBool('dav', 'caldav.external_attendees_disabled', false);
```
ChristophWurst:
```suggestion
		$externalAttendeesDisabled = $this->appConfig->getValueBool('dav', 'caldav_external_attendees_disabled', false);
```

SebastianKrupinski: Corrected

SebastianKrupinski:
> Is this something we want added to settings too?

This a admin setting, we could add this to dav setting for easy discovery, but design and @ChristophWurst would need to make that decision

ChristophWurst (approved):

ChristophWurst:
Documented configuration is sufficient as a first version

SebastianKrupinski:
> Documented configuration is sufficient as a first version

Documentation
https://github.com/nextcloud/documentation/pull/13988

Issue #7704, opened 2025-11-18
Full screen event editor doesn't show calendar's name when calendar is readonly


Mer0me:
### Steps to reproduce

1. Share your calendar read only to another user
2. Create an event
3. As the sharee, open the event with the full screen editor : calendar's name is missing
4.  On the simple view, the calendar's name is visible.

### Expected behavior

Full screen editor should display, at least, all the informations visible in the simple editor, even if the calendar is read only.

### Actual behavior

Simple editor : 

<img width="527" height="278" alt="Image" src="https://github.com/user-attachments/assets/d3692afd-1c89-40ba-bba2-925df9094ec6" />

Full screen editor : 

<img width="910" height="356" alt="Image" src="https://github.com/user-attachments/assets/5bd1267b-dfd5-4188-b988-90f341229b34" />

(By the way : the full screen editor is very empty and ugly in this use case, it should be redesigned)

### Calendar app version

6.1.0

### CalDAV-clients used

None here

### Browser

Firefox 145.0

### Client operating system

XUbuntu

### Server operating system

Debian

### Web server

Apache

### Database engine version

MariaDB

### PHP engine version

PHP 8.3

### Nextcloud version

32.0.1

### Updated from an older installed version or fresh install

Updated from an older version

### List of activated apps

```Text
not relevant here
```

### Nextcloud configuration

```Text
{
    "system": {
        "enable_previews": true,
        "enabledPreviewProviders": [
            "OC\\Preview\\Movie",
            "OC\\Preview\\PNG",
            "OC\\Preview\\JPEG",
            "OC\\Preview\\GIF",
            "OC\\Preview\\BMP",
            "OC\\Preview\\XBitmap",
            "OC\\Preview\\MP3",
            "OC\\Preview\\MP4",
            "OC\\Preview\\TXT",
            "OC\\Preview\\MarkDown",
            "OC\\Preview\\PDF"
        ],
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "merome.net",
            "cloud.merome.net",
            "192.168.1.100",
            "hpb.merome.net"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "overwrite.cli.url": "https:\/\/cloud.merome.net",
        "dbtype": "mysql",
        "version": "32.0.1.2",
[...]

PR #7818, opened 2025-12-30
fix: show calendar name in read-only event editor

abhinavohri:
Resolves: https://github.com/nextcloud/calendar/issues/7704

Summary: 
The calendar details section was hidden when `isReadOnly` was true, preventing users from seeing the calendar name and recurrence information.

Before:
image

After:
image

### DIFF
--- a/src/views/EditFull.vue
+++ b/src/views/EditFull.vue
@@ -111,7 +111,7 @@
 						@update-end-time="updateEndTime"
 						@update-end-timezone="updateEndTimezone" />
 
-					<div v-if="!isReadOnly" class="app-full__header__details">
+					<div class="app-full__header__details">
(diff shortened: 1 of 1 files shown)


abhinavohri:
Removing the `isReadOnly` check also shows the recurrence information. If we don't want to show that I can shift the `v-if="!isReadOnly"` check  to the inner div.

SebastianKrupinski:
Tested. Works good

abhinavohri:
Hi @SebastianKrupinski , I am interested in applying for the javascript developer opening at nextcloud and would value your advice. Is there a way we could get in touch?

SebastianKrupinski:
@abhinavohri Sure... send me a message at sebastian.krupinski (@) nextcloud.com

PR #7819, opened 2025-12-30
feat: show attachments only in view mode for simple editor

### LINKED ISSUE #7816: add attachment (with link) to calendar card

As a user of the calendar, I often attach PDF tickets to events (like shows) I'm going to.  It would be good if it was easy to see and retrieve that ticket directly from the event card interface to know that it had been attached and is the correct thing.  

Google shows it this way 

Image

- when you click there you go to the file
- rolling over reveals the filename
Something like this would be great.  

Not doing it... it's not the end of the world but it IS convenient to have this. 

### Additional context

Thank you for all the work that goes into this project.

abhinavohri:
Resolves: https://github.com/nextcloud/calendar/issues/7816
Summary:
This PR aims to show attachment list in the simple editor in viewing mode.

Before:
image

After:
image

### DIFF
--- a/src/components/Editor/Attachments/AttachmentsList.vue
+++ b/src/components/Editor/Attachments/AttachmentsList.vue
@@ -11,7 +11,7 @@
 			multiple
 			:aria-label="t('calendar', 'Upload files as attachments')"
 			@change="onLocalAttachmentSelected">
-		<div class="attachments-summary">
+		<div v-if="showSummary" class="attachments-summary">
 			<div class="attachments-summary-inner">
 				<Paperclip :size="20" />
 				<div v-if="attachments.length > 0" class="attachments-summary-inner-label">
@@ -40,7 +40,7 @@
 				</NcActionButton>
 			</NcActions>
 		</div>
-		<div v-if="attachments.length > 0">
+		<div v-if="showList">
 			<ul class="attachments-list">
 				<NcListItem
 					v-for="attachment in attachments"
@@ -123,6 +123,11 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+
+		compact: {
+			type: Boolean,
+			default: false,
+		},
 	},
 
 	data() {
@@ -143,6 +148,14 @@ export default {
 
 		attachments() {
 			return this.calendarObjectInstance.attachments
+		},
+
+		showSummary() {
+			return !this.compact || this.attachments.length > 1
+		},
+
+		showList() {
+			return this.attachments.length > 0 && (!this.compact || this.attachments.length === 1)
 		},
 	},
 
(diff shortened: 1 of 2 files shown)


abhinavohri:
Should attachments be visible during editing in simple editor, or hidden completely as they are now?
image

SebastianKrupinski:
Hi @abhinavohri 

Thank you for the PR. But this change will need approval from the design team.

SebastianKrupinski:
@nimishavijay 

What are your thoughts on this change? 

My only comment would be maybe we should make the attachments section collapsible? or list them horizontally instead of vertically to save space.

GVodyanov:
Hey! Might also ping @kra-mo, any ideas?

kra-mo:
I think as far as space concerns go, it might make sense to only show the name and icon if there is a single attachment (or maybe two/three), and without a heading. If there are more, it could just say `{n} attachments` and show the paperclip :)

abhinavohri:
@SebastianKrupinski @kra-mo hi, i have addressed the comments. Here is the latest UI:
Multiple attachements
image

Single Attachments:
image

kra-mo (dismissed):
The gap above the attachments seems too big to me and I'm not sure why the text is bold? The rest looks fine.

abhinavohri:
> The gap above the attachments seems too big to me and I'm not sure why the text is bold? The rest looks fine.

The gap above the attachments had two causes:
1. An empty location row was still being rendered, adding an extra gap. I have fixed that in this PR.
2. The remaining gap comes from the standard layout spacing in the simple editor.

The font weight and remaining spacing match the existing attachment styling in the full editor. Would you prefer me to override them specifically for the simple editor popover?

SebastianKrupinski:
Multiple Attachments:

Screenshot 2026-09-08 090119

Single Attachment:
Screenshot 2026-09-08 090152

kra-mo:
> Would you prefer me to override them specifically for the simple editor popover?

No, it was just that row that was problematic :)

kra-mo:
Hm, the row cuts off on the left? Looks good otherwise.

abhinavohri:
> Hm, the row cuts off on the left? Looks good otherwise.

The row looks cut off because the attachment list has an `-8px` horizontal margin, while the simple editor hides horizontal overflow, clipping its left edge. Should I override the inline-start margin specifically for the compact view, or leave the existing component styling as is?

kra-mo:
@abhinavohri I'm not sure what this means in practice. Do whatever makes sense for the code, it should just not cut off.

abhinavohri:
hey @kra-mo is this what you asked for?
image
image

SebastianKrupinski (approved):

SebastianKrupinski:
Made some minor visual changes

image

PR #7841, opened 2026-01-10
[stable5.5] fix: settings button icon alignment

st3iny:
This is not an issue on `main` and all other calendar branches below `stable5.5` are EOL.

| Before | After |
| --- | --- |
| image | image |

(Tested on Nextcloud `stable31`.)


SebastianKrupinski:
Thank you! @st3iny

PR #7854, opened 2026-01-14
chore: php version bump to 8.5

SebastianKrupinski:
### Summary
- added php 8.5 to all relevant files

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     },
     "require": {
-        "php": ">= 8.1 <=8.4",
+        "php": ">= 8.1 <=8.5",
         "bamarni/composer-bin-plugin": "^1.8.3"
     },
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -1,7 +1,7 @@
 {
     "config": {
         "platform": {
-            "php": "8.1"
+            "php": "8.2"
```
ChristophWurst: This means composer will now update packages even if they are incompatible with PHP8.1. Since this app still needs PHP8.1 for Nextcloud 32, it will potentially cause PHP incompatibilities.

ChristophWurst (changes requested):
PHP8.1 must not be removed

composer.json
SebastianKrupinski: Okay, corrected

PR #7857, opened 2026-01-15
chore: fix unit tests

SebastianKrupinski:
### Summary
- Updated node test workflow for due to failing test
- Modified vitest config to increase timeouts for slow CI


kesselb (approved):

vitest.config.js
DerDreschner, line 1: Default values are these:
- testTimeout: 5000
- hookTimeout: 10000

The `npm run test:unit:coverage` run on this PR took 8 seconds, on which less then one seconds contribute to tests and the longest *testsuite* (tests/javascript/unit/models/event.test.js with 15 tests) took 129ms in total.

Is this really necessary to change the default values?

SebastianKrupinski: During the CI runs today it was taking 100s+ and failing... this hopefully helps the issue. We can always change is back if it becomes a problem

PR #7859, opened 2026-01-15
chore(ci): improve php test

SebastianKrupinski:
### Summary
- Modified php test workflow to only run when a php or related file is changed, this should skip all php test for UI changes


SebastianKrupinski:
Not needed anymore

Issue #2572, opened 2020-09-09
Importer hides one existing calendar for some .ics files?


Matthias84:
### Steps to reproduce
1. Split big .ics files of a old NC export (to workaround #2570)
2. Start importing the first files, which runs fine
3. Notice that the importer doesn't list the desired calendar for the last two files anymore

I compared the files, but they look fine so far (VCALENDAR surrounding items, VEVENT and VTODO items). I also tried the first files, which looks good. Changing filename doesn't solve the issue. Reimporting the original big file, doesn't also work anymore. Logout and new session doesn't helped as well.

### Expected behaviour
Like all the other files, I wan't to import it to the same calendar so far.

### Actual behaviour
I can't pick the right (existing) calendar, to add more items to this calendar.
(no browser console messages appear, explaining why the item is removed from the list)
![Bildschirmfoto vom 2020-09-09 21-01-51](https://user-images.githubusercontent.com/217511/92644108-1a3e2600-f2e3-11ea-8f08-70a6c3b15442.png)

### Calendar app
**Calendar app version:** 2.0.4

**CalDAV-clients used:** -

### Client configuration
**Browser:** Firefox 80

**Operating system:** Linux Mint 20

### Server configuration

**Operating system**: CentOS

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** (e.g. 7.0.3)

**Nextcloud Version:** 19.0.2

**Updated from an older installed version or fresh install:** fresh

georgehrke:
> I compared the files, but they look fine so far (VCALENDAR surrounding items, VEVENT and VTODO items). 

If you check the database (oc_calendars), does the component type field for the calendar contain VTODO or just VEVENT?
It only offers calendars if all contained component types are supported by the calendar.

Matthias84:
Thanks for the hint! I used your tip from https://github.com/nextcloud/calendar/issues/2571#issuecomment-689786917 to force a detection of both types and it worked fine.

How about to explain the user this auto-behaviour? (e.g. disable list item / textfield if detection was triggered / let user choose the supported types manually before import, ...)

miaulalala:
Suggested solution

Check a file that is being imported before we start the process for the different types. If the file contains VTODOs only show calendars that support VTODO and also let the user know that this is the case so this information isn't inaccessible.

I would also like to see the supported component types somewhere, so users know what types are supported for each calendar.

PR #7875, opened 2026-01-18
feat: show supported component types in calendar settings

odzhychko:
Implements the suggestion to show supported component type from https://github.com/nextcloud/calendar/issues/2572#issuecomment-2200791216

Refs: https://github.com/nextcloud/calendar/issues/2572

### Before

Screenshot 2026-01-18 at 22-20-09 Jan 18 2026 - Calendar - Nextcloud

### After (current)

Screenshot 2026-01-21 at 16-26-47 Jan 21 2026 - Calendar - Nextcloud

### After (previous iterations)
<details>
  <summary>First iteration</summary>

  Screenshot 2026-01-18 at 22-18-16 Jan 18 2026 - Calendar - Nextcloud
</details>

### DIFF
--- a/src/components/AppNavigation/EditCalendarModal.vue
+++ b/src/components/AppNavigation/EditCalendarModal.vue
@@ -8,6 +8,9 @@
 		<div class="edit-calendar-modal">
 			<h3 class="edit-calendar-modal__header">
 				{{ $t('calendar', 'Edit calendar') }}
+				<span class="edit-calendar-modal__header_subtitle">
+					{{ description }}
+				</span>
 			</h3>
 
 			<div class="edit-calendar-modal__name-and-color">
@@ -83,6 +86,7 @@
 
 <script>
 import { showError } from '@nextcloud/dialogs'
+import { getLanguage } from '@nextcloud/l10n'
 import { NcAppNavigationSpacer, NcButton, NcCheckboxRadioSwitch, NcColorPicker, NcModal } from '@nextcloud/vue'
 import { mapStores } from 'pinia'
 import CheckIcon from 'vue-material-design-icons/Check.vue'
@@ -166,6 +170,39 @@ export default {
 		 */
 		downloadUrl() {
 			return this.calendar.url + '?export'
+		},
+
+		/**
+		 * Description about supported component types.
+		 *
+		 * @return {string}
+		 */
+		description() {
+			const supportedTypes = []
+			if (this.calendar.supportsEvents) {
+				supportedTypes.push(this.$t('calendar', 'events'))
+			}
+			if (this.calendar.supportsTasks) {
+				supportedTypes.push(this.$t('calendar', 'tasks'))
+			}
+			if (this.calendar.supportsJournals) {
+				// TRANSLATORS "notes" would be more user-friendly. "journal entries" (from RFC 5545) was used to avoid confusion with notes from the Notes app.
+				supportedTypes.push(this.$t('calendar', 'journal entries'))
+			}
+
+			if (supportedTypes.lenght === 0) {
+				return this.$t('calendar', 'This calendar supports neither events, tasks nor journal entries.')
+			}
+
+			const formatter = new Intl.ListFormat(getLanguage(), { type: 'conjunction' })
+			const localizedTypes = formatter.format(supportedTypes)
+			return this.$n(
+				'calendar',
+				'This calendar supports only {types}.',
+				'This calendar supports {types}.',
+				supportedTypes.length,
+				{ types: localizedTypes },
+			)
 		},
 	},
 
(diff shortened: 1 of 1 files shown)


SebastianKrupinski:
Hi @odzhychko 

Thank you for the PR...

These kind of changes do require designer approval...

@nimishavijay @kra-mo

nimishavijay:
This is a rather technical detail so I'm not sure if this just wouldn't be crowding the UI in many cases. At most we can show a simplified version of this in a tooltip next to the name, since it cannot be edited anyway. So next to the name there can be an `i` icon which shows "Supports events, tasks and journal entries" on hover. We can solve the issue of people trying to add unsupported types through smarter error prevention and better error messages, what do you think?

odzhychko:
> This is a rather technical detail so I'm not sure if this just wouldn't be crowding the UI in many cases.

Oh, I see now how this is an issue :sweat_smile:

> We can solve the issue of people trying to add unsupported types through smarter error prevention and better error messages, what do you think?

This is already being done. For example, when importing, creating, or editing events, only calendars that support events can be selected. Same for tasks. But being able to check supported types directly might still be useful and was suggested in https://github.com/nextcloud/calendar/issues/2572#issuecomment-2200791216

> So next to the tooltip there can be an i icon which shows "Supports events, tasks and journal entries" on hover.

I'm not sure about this regarding discoverability and touch-friendliness.
Users have to guess that they might find this information behind the info icon.
I assume hovering interactions on touchscreen devices are inconvenient in general.

@nimishavijay What are your thoughts on providing this information more compactly in one line (e.g. "Supports events, tasks and journal entries") but still always visible?
Maybe as a hint (with a light gray color/`var(--color-text-maxcontrast)`) below the title "Edit calendar" or below the name input.

---

I initially went with the unmodifiable checkboxes to be consistent with the rest of the editing elements of this edit modal and also have some visual structuring. Perhaps eventually this modal could become a properties modal where you can edit some properties (e.g., name, color, sharability) and inspect other properties (e.g., supported entries, creating time, source (e.g., owned, shared, built-in, read-only, subscribed from URL)).

ChristophWurst:
I would also like to see this as a simple one-liner like *This calendar supports events and tasks*. This doesn't take a lot of space, is not too technical.

ChristophWurst:
> @nimishavijay What are your thoughts on providing this information more compactly in one line (e.g. "Supports events, tasks and journal entries") but still always visible?
> Maybe as a hint (with a light gray color/`var(--color-text-maxcontrast)`) below the title "Edit calendar" or below the name input.

Let's go for this :)

odzhychko:
This is ready for review again :D

The screenshot in the description was updated appropriately.

SebastianKrupinski:
Tested. Works as expected.

image

SebastianKrupinski:
@ChristophWurst any things else to add?

SebastianKrupinski:
Rebased to head prep to merge

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -167,6 +170,36 @@ export default {
...
+			if (!this.calendar.supportsEvents && this.calendar.supportsTasks && this.calendar.supportsJournals) {
+				return this.$t('calendar', 'This calendar supports tasks and journal entries.')
+			}
+			if (!this.calendar.supportsEvents && this.calendar.supportsTasks && !this.calendar.supportsJournals) {
+				return this.$t('calendar', 'This calendar supports only tasks.')
+			}
+			if (!this.calendar.supportsEvents && !this.calendar.supportsTasks && this.calendar.supportsJournals) {
+				return this.$t('calendar', 'This calendar supports only journal entries.')
+			}
+			return this.$t('calendar', 'This calendar supports neither events, tasks nor journal entries.')
```
ChristophWurst: You can use Intl for the list. Here's an example: https://github.com/nextcloud/mail/blob/f699d91eae959b9045c9bb4cabbf5f35f0fdd62f/src/components/OutboxMessageListItem.vue#L113-L114

Build an array with the supported types as translated strings and run it through the conjunction formatter.

```js
$ const formatter = new Intl.ListFormat('de', { type: 'conjunction' })
$ formatter.format(['events', 'tasks'])
"events und tasks" 
```

Then use $n and singular/plural string like `This calendar supports only {types} entries.`/`This calendar supports {types}` 

Keep a separate handler for when none of the three components are supported

nimishavijay:
Looks much better! :) thank you @odzhychko !

src/components/AppNavigation/EditCalendarModal.vue
odzhychko: I noticed that some languages [that are well supported in Nextcloud](https://explore.transifex.com/nextcloud/nextcloud) are not yet supported by `Intl.ListFormat` even by the recent Chrome version 144 (tested on Fedora 43). This includes, for example, Irish `ga` and Galician `gl`.

I'm not sure how relevant this is. Let me know if I still can use can `Intl.ListFormat` here.

ChristophWurst: That is unfortunate. But I also don't see a good alternative right now, so this can be an accepted trade-off.

The alternative would be to have three translation strings with one, two or three placeholders. However, there could be issues with casing/grammar for some languages if we chop it up this way.

Let's proceed with Intl and accept the risk

odzhychko: Done

PR #7876, opened 2026-01-18
feat(importing): show all writable calendars in picker

### LINKED ISSUE #2572: Importer hides one existing calendar for some .ics files?

odzhychko:
Mark unsupported calendars as non-selectable.
Previously they were fully hidden.

Show a message about why some calendars are not selectable.

Also show calendars that do not support events (and only support tasks and/or journal entries).
Previously, calendars that did not support events were hidden altogether.

Resolves:  #2572

### Before

Screenshot From 2026-01-18 23-18-03

### After

image

### Test data

[SampleIcsFiles.zip](https://github.com/user-attachments/files/24699054/SampleIcsFiles.zip)

### After (previous iterations)
<details>
  <summary>First iteration</summary>

Screenshot From 2026-01-18 23-15-30
</details>

<details>
  <summary>Second iteraton (vertical margin 10px)</summary>
  Screenshot From 2026-01-25 19-38-36
  </details>

### DIFF
--- a/css/import.scss
+++ b/css/import.scss
@@ -21,7 +21,8 @@
 
 		.import-modal-file-item {
 			display: flex;
-			padding-top: 10px;
+			margin-top: calc(var(--default-grid-baseline) * 2);
+			margin-bottom: calc(var(--default-grid-baseline) * 2);
 
 			&--header {
 				font-weight: bold;
@@ -33,6 +34,10 @@
 
 			&__calendar-select {
 				flex: 1 1 0;
+			}
+			
+			&__calendar-disabled-hint {
+				color: var(--color-text-maxcontrast);
 			}
 		}
 	}
--- a/src/components/AppNavigation/Settings/ImportScreenRow.vue
+++ b/src/components/AppNavigation/Settings/ImportScreenRow.vue
@@ -6,17 +6,24 @@
 <template>
 	<li class="import-modal-file-item">
 		<div class="import-modal-file-item__filename">
-			{{ file.name }}
+			<div>{{ file.name }}</div>
+			<div
+				v-if="disabledHint"
+				class="import-modal-file-item__calendar-disabled-hint">
+				{{ disabledHint }}
+			</div>
 		</div>
 		<CalendarPicker
 			class="import-modal-file-item__calendar-select"
 			:value="calendar"
 			:calendars="calendars"
+			:isCalendarSelectable="isCalendarSelectable"
 			@selectCalendar="selectCalendar" />
 	</li>
 </template>
 
 <script>
+import { getLanguage } from '@nextcloud/l10n'
 import { mapStores } from 'pinia'
 import CalendarPicker from '../../Shared/CalendarPicker.vue'
 import useCalendarsStore from '../../../store/calendars.js'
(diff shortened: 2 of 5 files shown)


src/components/AppNavigation/Settings/ImportScreenRow.vue
```diff
@@ -39,59 +55,76 @@ export default {
...
+				return this.newCalendar
 			}
-
 			return this.calendarsStore.getCalendarById(calendarId)
 		},
 
 		calendars() {
-			// TODO: remove once the false positive is fixed upstream
+			// `this.calendarsStore.sortedCalendars` cannot be used because it already filters out calendars that do not support events.
+			const calendars = [...this.calendarsStore.calendars]
```
odzhychko:
```js
[...this.calendarsStore.calendars]
				.filter((calendar) => !calendar.readOnly)
				.sort((a, b) => a.order - b.order)
```

This should be encapsulated in a getter in the `this.calendarsStore`.
I will do it once we decide [whether this is even the expected behavior. ](https://github.com/nextcloud/calendar/pull/7876/files#r2702791853)

src/components/AppNavigation/Settings/ImportScreenRow.vue
```diff
@@ -6,12 +6,28 @@
 <template>
 	<li class="import-modal-file-item">
 		<div class="import-modal-file-item__filename">
-			{{ file.name }}
+			<div>{{ file.name }}</div>
+			<div
```
odzhychko, line 10: Instead of up to 3 messages, I now think a message with a list of entry types would be better.

src/store/calendars.js
```diff
@@ -90,6 +90,21 @@ export default defineStore('calendars', {
 				.sort((a, b) => a.order - b.order)
 		},
 
+		/**
+		 * List of sorted writable calendars.
+		 *
+		 * Even including ones without support for events.
```
odzhychko, line 96: This is an interesting case. And I'm not sure if the old behavior was intended or if, respectively, the new behavior is more appropriate.

Because [sortedCalendars](https://github.com/nextcloud/calendar/blob/26fa154eb596a5f4f08f98786f739f0346611cf9/src/store/calendars.js#L87https://github.com/nextcloud/calendar/blob/26fa154eb596a5f4f08f98786f739f0346611cf9/src/store/calendars.js#L87) (and [sortedCalendarFilteredByComponents](https://github.com/nextcloud/calendar/blob/26fa154eb596a5f4f08f98786f739f0346611cf9/src/store/calendars.js#L227)) only returned calendars that at least supported VEVENT users could never import into calendars that only support VTODO or/and VJOURNAL.

src/components/AppNavigation/Settings/ImportScreenRow.vue
```diff
@@ -39,59 +45,94 @@ export default {
...
+		 * @return {string|undefined} A message, or undefined if all calendars can be selected.
+		 */
+		disabledHint() {
+			const disalbedBecauseOfEvents = this.file.parser.containsVEvents() && this.calendars.some((calendar) => !calendar.supportsEvents)
+			const disalbedBecauseOfTasks = this.file.parser.containsVTodos() && this.calendars.some((calendar) => !calendar.supportsTasks)
+			const disalbedBecauseOfJournalEntries = this.file.parser.containsVJournals() && this.calendars.some((calendar) => !calendar.supportsJournals)
 
-			return calendars
+			if (disalbedBecauseOfEvents && disalbedBecauseOfTasks && disalbedBecauseOfJournalEntries) {
+				return this.$t('calendar', 'Some calendars are disabled because this file contains events, tasks and journal entries.')
```
odzhychko: Maybe `Intl.ListFormat` can be used.

See and wait for the outcome of https://github.com/nextcloud/calendar/pull/7875#discussion_r2713213665

src/store/calendars.js
ChristophWurst: I'm fine with lifting this restriction. If someone has a dedicated collection only for tasks they should be able to import more tasks.

src/components/AppNavigation/Settings/ImportScreenRow.vue
odzhychko: Done

src/store/calendars.js
```diff
@@ -90,6 +90,21 @@ export default defineStore('calendars', {
...
+		 * Those are usually excluded by all other getters.
+		 *
+		 * @param {object} state the store data
+		 * @return {Array}
+		 */
+		sortedWritableCalendarsEvenWithoutSupportForEvents(state) {
```
odzhychko, line 102: I will add a test for this function.

odzhychko: Done

odzhychko:
This code is now rebased and works with the recent update to Vue 3.

css/import.scss
```diff
@@ -21,7 +21,8 @@
 
 		.import-modal-file-item {
 			display: flex;
-			padding-top: 10px;
+			margin-top: 10px;
```
GVodyanov: Hey! Thanks a lot for the PR!

Could you please use css variables for these changes?

https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/css.html
```suggestion
			margin-top: calc(var(--default-grid-baseline) * 2);
```

odzhychko: Done

Cool that the UI gets more consistent bit by bit :smile:

hamza221 (approved):

GVodyanov (approved):
Code looks good! Tested it and works

PR #7878, opened 2026-01-19
fix(Repeat): add button to submit

GVodyanov:
Adds this "Add" button:
image

### DIFF
--- a/src/components/Editor/Repeat/Repeat.vue
+++ b/src/components/Editor/Repeat/Repeat.vue
@@ -78,12 +78,19 @@
 				<RepeatUnsupportedWarning v-if="recurrenceRule.isUnsupported && !isRecurrenceException" />
 				<RepeatExceptionWarning v-if="isRecurrenceException" />
 			</div>
+			<div
+				v-if="!isRecurrenceException && !isReadOnly"
+				class="property-repeat__options__footer">
+				<NcButton variant="primary" @click="saveAndClose">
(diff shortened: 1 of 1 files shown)


src/components/Editor/Repeat/Repeat.vue
```diff
@@ -78,12 +78,19 @@
 				<RepeatUnsupportedWarning v-if="recurrenceRule.isUnsupported && !isRecurrenceException" />
 				<RepeatExceptionWarning v-if="isRecurrenceException" />
 			</div>
+			<div
+				v-if="!isRecurrenceException && !isReadOnly"
+				class="property-repeat__options-footer">
```
ChristophWurst:
```suggestion
				class="property-repeat__options__footer">
```

footer is an element, not a modifier :)

src/components/Editor/Repeat/Repeat.vue
```diff
@@ -507,4 +522,10 @@ export default {
 	flex-direction: column;
 	gap: calc(var(--default-grid-baseline) * 2);
 }
+
+.property-repeat__options-footer {
```
ChristophWurst: nit: rearrange by BEM

* .property-repeat
  * &__options
    * &__footer

ChristophWurst: I mean for the styles to be nested like the elements

PR #7885, opened 2026-01-20
Workflow updates

kesselb:


renovate.json
```diff
@@ -34,6 +34,26 @@
...
+		".github/workflows/lint-stylelint.yml",
+		".github/workflows/node-test.yml",
+		".github/workflows/node.yml",
+		".github/workflows/npm-audit-fix.yml",
+		".github/workflows/npm-audit-fix-selective.yml",
+		".github/workflows/openapi.yml",
+		".github/workflows/pr-feedback.yml",
+		".github/workflows/psalm-matrix.yml",
+		".github/workflows/reuse.yml"
+	],
```
SebastianKrupinski: Why do we want to ignore the workflows? I use renovate to updated these all the time

.github/workflows/node-test.yml
```diff
@@ -102,7 +105,7 @@ jobs:
 
     if: always()
 
-    name: npm-test-summary
+    name: test-summary
```
SebastianKrupinski: This breaks the summary

PR #7892, opened 2026-01-22
fix: cancel confirmation dialog

GVodyanov:
Change wording, close button not working, and dirty state handling


SebastianKrupinski:
Tested works good! Words are changed.

image

PR #7894, opened 2026-01-22
fix: improve simple editor positioning logic

SebastianKrupinski:
### Summary
- Modified popup positioning logic to address unscrolling effect when opening popup near the bottom of the screen


SebastianKrupinski:
Hey @GVodyanov 

Can you give this a quick spin please

PR #7899, opened 2026-01-23
[stable6.2] chore: adjust php test matrix

SebastianKrupinski:
### Summary
- Adjusted NC version
- Adjusted PHP version


hamza221:
~~What's the logic behind Php versions? if min max shouldn't it be 8.1-8.5, according to  `info.xml`~~
Edit: missed to correlate with the supported php server versions

PR #7942, opened 2026-01-31
test: eliminate unit tests relying on order

odzhychko:
In tests, some stores were not recreated before each test. They broke when running them in a random order.
This could be achieved by running `npx vitest --run  --sequence.shuffle`.

Properly recreating stores also allowed refactoring the tests to make them more clear.

### DIFF
--- a/tests/javascript/unit/store/contacts.test.js
+++ b/tests/javascript/unit/store/contacts.test.js
@@ -7,9 +7,12 @@ import { setActivePinia, createPinia } from 'pinia'
 
 describe('store/contacts test suite', () => {
 
-	setActivePinia(createPinia())
+	let contactsStore
 
-	const contactsStore = useContactsStore()
+	beforeEach(() => {
+		setActivePinia(createPinia())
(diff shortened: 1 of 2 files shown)


SebastianKrupinski:
Hi @odzhychko 

Thank you for the PR, what issue is this meant to address? Is there a ticket?

odzhychko:
Hi @SebastianKrupinski , thanks for looking at it.

The issue these changes adress: 

- The modified tests are now more robust because they do not rely on the execution order of test cases.
  - Adding new test cases or modifying existing ones will be easier and less error-prone.
  - Each test case can be understood without looking at the other test cases.
- They are more consistent with the other tests for the store.
  - The others already recreated Pinia before each test case.
  
If you do not find this useful (at the moment), feel free to close it. I didn't create a ticket because it's only a refactoring of tests and explaining the issue with already proposed code seemed the best approach. Let me know if I should create a ticket for this. I wasn't sure anyway how to best approach such contributions.

PR #7951, opened 2026-02-03
feat: read calendar ACL properly

SebastianKrupinski:
### Summary
- modified ACL's for calendar federation

### DIFF
--- a/vitest.config.js
+++ b/vitest.config.js
@@ -25,7 +25,7 @@ export default defineConfig({
 		// Required for transforming CSS files
 		pool: 'vmForks',
 		// Increase timeouts for slow CI environments
-		testTimeout: 120000, // 2 minutes per test
-		hookTimeout: 30000,  // 30 seconds for hooks
+		testTimeout: 300000, // 2 minutes per test
+		hookTimeout: 60000,  // 60 seconds for hooks
 	},
 });


src/models/calendar.js
```diff
@@ -86,6 +92,9 @@ function mapDavCollectionToCalendar(calendar, currentUserPrincipal) {
 	const readOnly = !calendar.isWriteable()
 	const canBeShared = calendar.isShareable()
 	const canBePublished = calendar.isPublishable()
+	const canCreateObject = calendar.currentUserPrivilegeSet.includes('{DAV:}bind') || calendar.currentUserPrivilegeSet.includes('{DAV:}write') || calendar.currentUserPrivilegeSet.includes('{DAV:}all') === true
+	const canModifyObject = calendar.currentUserPrivilegeSet.includes('{DAV:}write-content') || calendar.currentUserPrivilegeSet.includes('{DAV:}write') || calendar.currentUserPrivilegeSet.includes('{DAV:}all') === true
+	const canDeleteObject = calendar.currentUserPrivilegeSet.includes('{DAV:}unbind') || calendar.currentUserPrivilegeSet.includes('{DAV:}write') || calendar.currentUserPrivilegeSet.includes('{DAV:}all') === true
```
tcitworld, line 97: Would it be worth it to add these in cdav library instead?

SebastianKrupinski: Hello! Yes that makes sense! I will do that, but at the moment this is so we can get a feature out... Then I will modify the dav-library

PR #7970, opened 2026-02-09
chore: Release v6.2.0

SebastianKrupinski:
### Summary
- Changed version to stable release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 6.2.0
+- No changes since RC.
+
 ## 6.2.0-rc.1
 ### Fixed
 - appointment booking room creation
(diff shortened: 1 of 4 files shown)


hamza221 (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 6.2.0
+- No changes since RC.
```
hamza221, line 4: I'm against this, I would rather rename `## 6.2.0-rc.1`  to `## 6.2.0` instead

hamza221: In case someone is checking only the latest changleog say in store ... But your call

SebastianKrupinski: I'm going to leave it for today... I am going to look in to using the same script we use in the workflows to updated the change log in the repo... its a pain doing this by hand every time...

PR #8003, opened 2026-02-23
fix: put default calendar selector back

SebastianKrupinski:
### Summary
- added the calendar picker component back to the settings modal

image


SebastianKrupinski:
> ESlint apart looks good

🦃

Issue #6485, opened 2024-11-11
Issue with Event Creation Using Circles - "Could not find organizer" Error


dev-marcoC:
### Steps to reproduce

- Create a Circle (team) in Nextcloud.
- Create a new calendar event and invite the Circle as participants.

### Expected behavior

The event should be created successfully with the members of the Circle invited as participants, and the organizer field should be correctly assigned.

### Actual behaviour

The event fails to create with the error message mentioned above, and the log shows the "Could not find organizer" exception. It seems that there is an issue with the synchronization of Circle members or the identification of the event organizer.
![Image](https://github.com/user-attachments/assets/53cf0438-09b1-4547-b6f3-8f6c903bb276)

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

When trying to create an event in Nextcloud and invite a team (Circle) I created, I encounter two issues:

The message:
"Note that members of circles get invited but are not synced yet."

The following error in the logs:
"ServiceException: Could not find organizer"

Additional Information:
I’ve verified that the Circle members are correctly configured and present in Nextcloud.
I’ve checked permissions and made sure my account has the necessary rights to create events.
I have tried creating events without using Circles, and it works fine, which suggests the issue is related specifically to the use of Circles.

endoflevelboss:
Probably a duplicate of [nextcloud/circles #1690](https://github.com/nextcloud/circles/issues/1690)
The error message slightly differs, but the issue is identical.

MrAn0nym:
@dev-marcoC Do you by any chance still know, if the Circle had any of the following as a member: group, another circle, email, external contact?

Issue #6023, opened 2024-05-27
Calendar - error when creating event with an invited circle containing a group


tartard:
### Steps to reproduce

1. Create a circle which contains a group
2. Create an event in the calendar
3. Try to invite the previously created circle to the event

### Expected behavior

Members of the group which is member of the circle should be invited to the event 

### Actual behaviour

When clicking the circle found in the invite search, a server error occurs (error 500) during the following request :   
GET https://nextcloud.rochefort-numerique.fr/index.php/apps/calendar/v1/circles/getmembers?format=json&circleId=BaSgPX5ecWIZotULqBSiuw4YzO5BaLY

 with the following server log : 

{"reqId":"RJzeOpx2dzk9BLRWMgqJ","level":3,"time":"2024-05-27T13:05:49+00:00","remoteAddr":"86.214.180.2","user":"****","app":"index","method":"GET","url":"/index.php/apps/calendar/v1/circles/getmembers?format=json&circleId=BaSgPX5ecWIZotULqBSiuw4YzO5BaLY","message":"Could not find organizer","userAgent":"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:123.0) Gecko/20100101 Firefox/123.0","version":"27.1.3.2","exception":{"Exception":"OCA\\Calendar\\Service\\ServiceException","Message":"Could not find organizer","Code":0,"Trace":[{"file":"/var/www/html/lib/private/AppFramework/Http/Dispatcher.php","line":230,"function":"getCircleMembers","class":"OCA\\Calendar\\Controller\\ContactController","type":"->"},{"file":"/var/www/html/lib/private/AppFramework/Http/Dispatcher.php","line":137,"function":"executeController","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->"},{"file":"/var/www/html/lib/private/AppFramework/App.php","line":183,"function":"dispatch","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->"},{"file":"/var/www/html/lib/private/Route/Router.php","line":315,"function":"main","class":"OC\\AppFramework\\App","type":"::"},{"file":"/var/www/html/lib/base.php","line":1068,"function":"match","class":"OC\\Route\\Router","type":"->"},{"file":"/var/www/html/index.php","line":36,"function":"handleRequest","class":"OC","type":"::"}],"File":"/var/www/html/apps/calendar/lib/Controller/ContactController.php","Line":233,"CustomMessage":"--"}}

[...]

PR #8008, opened 2026-02-24
fix: adding circles to events (#6023)

MrAn0nym:
[ContactController::getCircleMembers](https://github.com/nextcloud/calendar/blob/main/lib/Controller/ContactController.php) is used to fetch all members of a circle, when adding that circle to an event.

Currently the function assumes members to be a nextcloud user:
https://github.com/nextcloud/calendar/blob/04e8e493108bfce34d3800b74be19ba2dea22480/lib/Controller/ContactController.php#L227-L229

This results in empty return values, which in turn causes an error and results in no member of the circle being added
https://github.com/nextcloud/calendar/blob/0a0dc2382fbc6eb0ac871b7273adee15316c3ae4/lib/Controller/ContactController.php#L231-L233

The PR changes the code to handle all types separately. Whilst this is not perfect and problems like federation or auto-updating memberlists are not solved by this it at least provides a short term fix for #6023 #6485 and one aspect of https://github.com/nextcloud/circles/issues/1690 if I'm not mistaken.

### DIFF
--- a/lib/Controller/ContactController.php
+++ b/lib/Controller/ContactController.php
@@ -11,6 +11,7 @@ use Exception;
 use OCA\Calendar\Service\ContactsService;
 use OCA\Calendar\Service\ServiceException;
 use OCA\Circles\Exceptions\CircleNotFoundException;
+use OCA\Circles\Model\Member;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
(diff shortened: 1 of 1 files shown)


MrAn0nym:
Didn't manage to look into how one would write tests yet and I believe `php-cs-fixer` will at least replace some quotes as well. If anyone else has the time to fix those last steps feel free to do so, otherwise I'll do so when I get around

SebastianKrupinski:
Hi @MrAn0nym 

Thank you for the PR. 

FYI, you might want to wait a couple weeks for this, as there are some pending changes that will be applied to the contacts controller which might effect your work.

odzhychko:
Relates to https://github.com/nextcloud/calendar/pull/7554

PR #8009, opened 2026-02-25
fix(capability): Add a capability if calendar webui is there

nickvergessen:


lib/AppInfo/Capabilities.php
```diff
@@ -0,0 +1,21 @@
...
+namespace OCA\Calendar\AppInfo;
+
+use OCP\Capabilities\IPublicCapability;
+
+class Capabilities implements IPublicCapability {
+	#[\Override]
+	public function getCapabilities(): array {
+		return [
+			'calendar' => [
+				'webui' => true,
```
Antreesy, line 17: not `enabled`, cause CalDAV is always on?

Antreesy (approved):
Works good enough for Talk

ChristophWurst:
This feels like a competing logic of https://github.com/nextcloud/server/pull/46510 :thinking:

nickvergessen:
> This feels like a competing logic of https://github.com/nextcloud/server/pull/46510 🤔

Well, we need this information in capabilities so mobile clients and desktop clients can consume the info as well.
Felt wrong to put it into Talks capabilities though

ChristophWurst:
That makes sense. I would just wonder if we should have a capability by the dav app that exposes the result of `\OCP\App\IAppManager::isBackendRequired(`\OCP\App\IAppManager::BACKEND_CALDAV`)

nickvergessen:
We don't need the "backend" but the "frontend" aka "is it okay to link events with href to `apps/calendar`"
But I guess yeah the dav app could have a "backend-required" and "frontend-available" capability.

PR #8011, opened 2026-02-25
fix: show proper dialogs for delete and convert

SebastianKrupinski:
### Summary
- Modified the proposal delete and convert to show a proper nc dialog popup

### Now
image

### Before
image


tcitworld:
The type: error style isn't applied in the screenshot ?

SebastianKrupinski:
> The type: error style isn't applied in the screenshot ?

Fixed. Thank you for spotting it!

Issue #7929, opened 2026-01-29
Contrast issues (Accessibility)


GVodyanov:
### Steps to reproduce

Issues:
- Past events are translucent making them not have enough contrast:

<img width="411" height="46" alt="Image" src="https://github.com/user-attachments/assets/908cc9e4-6268-4eff-ad03-64149b9e2e0a" />

- Tentative icon is yellow on white and doesn't have enough contrast:

<img width="157" height="53" alt="Image" src="https://github.com/user-attachments/assets/6f13110b-16da-4367-8cb0-3806e48a1397" />

There may be other instances of bad contrast

### Expected behavior

Contrast respecting WCAG guidelines

### Actual behavior

Not enough contrast in some parts of the app

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

PR #8019, opened 2026-02-27
fix: opacity of past events and next/past month dates

barbaraborges:
Related to [Issue 7929](https://github.com/nextcloud/calendar/issues/7929)

The default colors on the calendar color-pick already have enough contrast but an opacity was set, reducing it. Similar to the date number on past/next calendar dates.

This PR:
- Removes opacity for event past, event task completed, event cancelled and event tentative.
- Bumps up opacity for date number on past/next calendar dates for the minimum accepted for level AAA on WCAG (tested with [WebAIM contract tool](https://webaim.org/resources/contrastchecker/))


SebastianKrupinski:
Hi @barbaraborges 

Thank you for the PR!

SebastianKrupinski:
Tested. Looks good to me.

image

image

SebastianKrupinski:
@GVodyanov 

Any objections?

GVodyanov (approved):
Thanks for the PR! Approving this because yeah, what we have now absolutely isn't accessible.

Having said that maybe we could think of some other way of visually distinguishing past events, seeing as I know some users liked the feature, and it's a bit of a shame to see it go. What we have now though definitely has to go.

@kra-mo @nimishavijay Do you happen to have any ideas for some accessible ways to do this? I'll create a new issue to address this.

EDIT https://github.com/nextcloud/calendar/issues/8022

Issue #629, opened 2017-10-11
Default reminders configuration [$113]


laurentdinclaux:
It is tedious to set reminders each time I add an event. If I forget to do it, I miss appointments.

Please add a setting to set default reminders per calendar.

For reference: https://help.nextcloud.com/t/default-reminder-in-calendar/3213

pixelipo:
I would add to a bounty for this. Killer feature.

tcitworld:
Bounty URL : https://www.bountysource.com/issues/50190368-default-reminders-configuration

thigg:
Also, please add the possibility to change the default choices of reminder times. Currently they are pretty useless.

georgehrke:
> Currently they are pretty useless.

Can you elaborate? Why are they useless? What reminder times would be better? ...

jospoortvliet:
Better defaults would make sense - making it configurable seems a bit over the top if you ask me?!? Then again, people have different needs/wants. Perhaps the calendar should remember what the user did last time? I also always use the same setup of 4 reminders - 1 day, 1 hour, 5 minutes and on the time. It is indeed a bit tedious to add them.

msharp1:
We **need** this feature, 100% agree with the OP.
In a perfect world I would like to see every basic things that is found useful in Google Agenda to be ported in NextCloud Calendar :)  At the moment I'm having a hard time convincing my social circle to take the step out from Google because of "details" like this.
And I'm a bit scared to see that this feature is planned for 1.8.0 , du by October 2018 ! This is too far.
I'm ready to back for 15$ , are you still with us @pixelipo  ?

gessel:
I am agree, the two most useful features of google's calendar are well-thought-out reminders and proper use of time zone specifications.

georgehrke:
> and proper use of time zone specifications.

@gessel Please don't hijack this issue but instead open a new ticket if you have a feature request. (and please be a bit more precise about what's the current and what's the expected behavior.)

georgehrke:
> Edit and answer to the comment below: @georgehrke
Usually I want to be reminded 2 days in advance and 2 hours in advance. When the appointment is further into the future maybe 1 week in advance.

> The default times which are selectable via dropdown offer only the range from 5 minutes up to two hours. Therefore everytime when I add an appointment via the calendar app, I have to manually create the 2 days reminder.

> This is pretty frustrating, especially when you never create reminders which are closer to the appointment then two hours.

[...]

PR #8031, opened 2026-03-02
feat(EditCalendarModal): add setting for default reminder per calendar

GVodyanov:
Resolves: https://github.com/nextcloud/calendar/issues/629
Requires: https://github.com/nextcloud/server/pull/58679 and https://github.com/nextcloud/cdav-library/pull/995

### DIFF
--- a/src/components/AppNavigation/EditCalendarModal.vue
+++ b/src/components/AppNavigation/EditCalendarModal.vue
@@ -43,6 +43,24 @@
 					{{ $t('calendar', 'Never show me as busy (set this calendar to transparent)') }}
 				</NcCheckboxRadioSwitch>
 			</template>
+			<template v-if="!calendar.isSharedWithMe && isAfterVersion">
+				<div class="edit-calendar-modal__default-alarm">
+					<label for="default-alarm-select" class="edit-calendar-modal__default-alarm__label">
+						{{ $t('calendar', 'Default reminder') }}
+					</label>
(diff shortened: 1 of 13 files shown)


src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -222,6 +280,16 @@ export default {
 			this.calendarNameChanged = false
 			this.calendarColorChanged = false
 			this.isTransparent = calendar.transparency === 'transparent'
+
+			// Initialize default alarm
+			if (calendar.defaultAlarm === null || calendar.defaultAlarm === 'none') {
```
SebastianKrupinski: Since the default value in the database is an INT we should not be checking for a 'none'

src/components/AppNavigation/EditCalendarModal.vue
```diff
@@ -287,6 +355,25 @@ export default {
 			}
 		},
 
+		/**
+		 * Save the calendar default alarm.
+		 */
+		async saveDefaultAlarm() {
+			try {
+				const defaultAlarmValue = this.selectedDefaultAlarm ? this.selectedDefaultAlarm.value : 'none'
```
SebastianKrupinski: Same here.

GVodyanov:
Well this PR ballooned but it should be ready now, however we have to wait for cdav to update

Issue #8037, opened 2026-03-03
When calendar DisplayName is set to null, activity log shows error "Rich subject or a parameter for ... is malformed"


Mer0me:
### Steps to reproduce

1. Edit your calendar DisplayName and set it to a blank string
2. Add an event in this calendar
3. Check your activities and see the error : Rich subject or a parameter for "Vous avez créé l’évènement {event} dans l’agenda {calendar}" is malformed

### Expected behavior

We shouldn't be able to set a blank string on the calendar Display Name

### Actual behavior

Error on the activity log

### Calendar app version

6.2.1

### CalDAV-clients used

none

### Browser

Firefox 147.0.4

### Client operating system

Ubuntu

### Server operating system

Debian

### Web server

Apache

### Database engine version

MariaDB

### PHP engine version

PHP 8.3

### Nextcloud version

33.0.0

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text
not relevant
```

### Nextcloud configuration

```Text
not relevant
```

### Web server error log

```Text
not relevant
```

### Log file

```Text
not relevant
```

### Browser log

```Text
not relevant
```

### Additional info

2 users (among ~600 here) had set their calendar's DisplayName to null when they tried to share their calendar : the DisplayName field is focused when the Calendar sharing window is displayed, they probably hit the Delete key and then shared their calendar normally.
The behavior of this window should probably be better.

SebastianKrupinski:
Hi @Mer0me 

Thanks for letting us know about this issue. I've made a patch to prevent a blank calendar name.

PR #8038, opened 2026-03-04
fix: calendar name input

SebastianKrupinski:
### Summary
- resolves: https://github.com/nextcloud/calendar/issues/8037
- added logic to prevent calendar being created with a blank name
- added logic to prevent calendar name from being set to blank
- added logic to prevent calendar subscription from being submitted with invalid url

### DIFF
--- a/src/components/AppNavigation/CalendarList/CalendarListNew.vue
+++ b/src/components/AppNavigation/CalendarList/CalendarListNew.vue
@@ -262,10 +262,15 @@ export default {
 		 * @param {Event} event The submit event
 		 */
 		async createNewCalendar(event) {
+			const displayName = event.target.querySelector('input[type=text]').value
+
+			if (!displayName.trim()) {
+				showError(this.$t('calendar', 'Calendar name can not be blank'))
+				return
+			}
+
 			this.showCreateCalendarInput = false
 			this.showCreateCalendarSaving = true
-
-			const displayName = event.target.querySelector('input[type=text]').value
 
 			try {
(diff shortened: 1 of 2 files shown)


src/components/AppNavigation/CalendarList/CalendarListNew.vue
tcitworld, line 334: We should trim the value of `link` as well even if it doesn't matter for the validation with `new URL`, as it will be sent to server with the extra characters (not sure if server does the trimming).

src/components/AppNavigation/CalendarList/CalendarListNew.vue
```diff
@@ -262,11 +262,16 @@ export default {
 		 * @param {Event} event The submit event
 		 */
 		async createNewCalendar(event) {
+			const displayName = event.target.querySelector('input[type=text]').value
```
tcitworld, line 265: Not directly linked, but we really should use a `v-model` on `ActionInput` and use the value instead of fetching it from the DOM like this (same for calendar with tasks).

src/components/AppNavigation/CalendarList/CalendarListNew.vue
SebastianKrupinski: You make a good point! Ty

src/components/AppNavigation/CalendarList/CalendarListNew.vue
SebastianKrupinski: I agree, we'll need to fix this in a followup

PR #8048, opened 2026-03-08
feat: add event reference provider and widget

### LINKED ISSUE #7104: Reference provider for appointments

It would be great to reference/link a particular appointment using the smart picker, e.g. in a page in collectives.

Currently it's only possible to reference a full calendar by pasting the URL to the calendar (not via smart picker either, see https://github.com/nextcloud/calendar/issues/7103).

We had a Nextcloud Collectives user meetup recently and several people reported that it would be helpful for them to be able to reference appointments in a Collectives page.

_No response_

_No response_

### Additional context

_No response_

mejo-:
Fixes: #7104

### Screenshot

image

### DIFF
--- a/lib/Reference/EventReferenceProvider.php
+++ b/lib/Reference/EventReferenceProvider.php
@@ -1,0 +1,188 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Calendar\Reference;
+
+use OCA\Calendar\AppInfo\Application;
+use OCA\Calendar\Service\ObjectResolverService;
+use OCP\Calendar\ICalendar;
+use OCP\Calendar\IManager;
+use OCP\Collaboration\Reference\ADiscoverableReferenceProvider;
+use OCP\Collaboration\Reference\IReference;
+use OCP\Collaboration\Reference\Reference;
+use OCP\IDateTimeFormatter;
+use OCP\IL10N;
+use OCP\IURLGenerator;
+
+class EventReferenceProvider extends ADiscoverableReferenceProvider {
+	public function __construct(
+		private readonly IL10N $l10n,
+		private readonly IURLGenerator $urlGenerator,
+		private readonly IManager $calendarManager,
+		private readonly ObjectResolverService $objectResolverService,
+		private readonly IDateTimeFormatter $dateTimeFormatter,
+		private readonly ?string $userId,
+	) {
+	}
+
+	#[\Override]
+	public function getId(): string {
+		return 'calendar-event';
+	}
+
+	#[\Override]
+	public function getTitle(): string {
+		return $this->l10n->t('Calendar event');
+	}
+
+	#[\Override]
+	public function getOrder(): int {
+		return 21;
+	}
+
+	#[\Override]
+	public function getIconUrl(): string {
+		return $this->urlGenerator->getAbsoluteURL(
+			$this->urlGenerator->imagePath(Application::APP_ID, 'calendar-dark.svg')
+		);
+	}
+
+	#[\Override]
+	public function matchReference(string $referenceText): bool {
+		$start = $this->urlGenerator->getAbsoluteURL('/apps/' . Application::APP_ID);
+		$startIndex = $this->urlGenerator->getAbsoluteURL('/index.php/apps/' . Application::APP_ID);
+
+		foreach ([$start, $startIndex] as $base) {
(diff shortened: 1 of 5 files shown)


lib/Reference/EventReferenceProvider.php
```diff
@@ -0,0 +1,207 @@
...
+use OCP\IL10N;
+use OCP\IURLGenerator;
+use Sabre\VObject\Reader;
+
+class EventReferenceProvider extends ADiscoverableReferenceProvider {
+	public function __construct(
+		private readonly IL10N $l10n,
+		private readonly IURLGenerator $urlGenerator,
+		private readonly IManager $calendarManager,
+		private CalDavBackend $calDavBackend,
```
tcitworld: This should be made backend agnostic, and use `OCP\ICalendar\IManager::getCalendarsForPrincipal` in the `getCalendar` method, and `OCP\Calendar\ICalendar\search` in the `getEventData` method.

tcitworld (changes requested):

src/views/EventReferenceWidget.vue
```diff
@@ -0,0 +1,163 @@
...
+	methods: {
+		t,
+	},
+})
+</script>
+
+<style lang="scss" scoped>
+.event {
+	width: 100%;
+	padding-inline: 12px;
```
GVodyanov:
```suggestion
	padding-inline: calc(var(--default-grid-baseline) * 3);
```

GVodyanov (commented):
Thanks for the contribution!

Please use our CSS variables instead of custom values for padding/margin/gap/etc. 

You can check them out here https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/css.html

src/views/EventReferenceWidget.vue
mejo-: Thanks, changed now.

lib/Reference/EventReferenceProvider.php
mejo-: Can you help me to use `ICalendar::search()` for finding the event? It seems like it doesn't allow to filter by event URI (e.g. `45C2F160-BA74-4910-B302-79356C24A458.ics`). So my best bet would be to search for all events in the calendar and then iterate through them to get the one with the matching URI:

```php
$event = null;
foreach ($calendar->search('') as $result) {
	if (($result['uri'] ?? null) === $eventFile) {
		$event = $result;
		break;
	}
}
```

This works but doesn't seem a good idea performance-wise. Or is it fine to do it that way?

mejo-:
I realized another problem: calendar event URLs (e.g. `https://nextcloud.local/index.php/apps/calendar/dayGridMonth/now/edit/popover/L3JlbW90ZS5waHAvZGF2L2NhbGVuZGFycy9qYW5lL3BlcnNvbmFsLzQ1QzJGMTYwLUJBNzQtNDkxMC1CMzAyLTc5MzU2QzI0QTQ1OC5pY3M=/1772787600`)  seem to "expire" in some way. A few hours after I copied them into a document, the preview still renders correctly, but opening the URL just opens the calendar in question with a modal "Event does not exist".

Is this a bug, or are these URLs meant to be ephemeral? Is there something like a persistent URL to an event yet at all?

lib/Reference/EventReferenceProvider.php
tcitworld: I mistook the object uri and the object uid, sorry about that.

You can call `search('', [], ['uid' =>  $uid]);`, but the object uri is not always `uid.ics`, so this won't work.

I think we could add a modification in the calendar backend to allow to also do search by object uri. I don't know of any implementation using `ICalendar::search` (or interfaces that extend it) at the moment, so it shouldn't be a problem.
https://github.com/nextcloud/server/blob/7cdbb38d526edc2c32240fffb64b035d5dc0cb35/apps/dav/lib/CalDAV/CalDavBackend.php#L2061-L2063

tcitworld:
> Is this a bug, or are these URLs meant to be ephemeral? Is there something like a persistent URL to an event yet at all?

It might be related to the event occurrence added at the end (`/1772787600`). Did you try without it?

lib/Reference/EventReferenceProvider.php
mejo-: Good idea, I did this now: https://github.com/nextcloud/server/pull/59142

Tested with this PR (adjusted the `search()` parameters) and works well.

mejo-:
> It might be related to the event occurrence added at the end (/1772787600). Did you try without it?

There's no route for URLs without the event occurrence part at the end, so unfortunately this leads to "Page not found" :disappointed:

mejo-:
> There's no route for URLs without the event occurrence part at the end, so unfortunately this leads to "Page not found" 😞

This is now fixed in #8123 thanks to @hamza221 :rocket:

PR #8063, opened 2026-03-12
feat: try to detect mailing lists as event attendees

ChristophWurst:
The backend doesn't differenciate between individual email and a list. It will just send the invite. For a list, the invite is bogus and can't really be answered. We should warn the user.

image

### DIFF
--- a/src/components/Editor/Invitees/InviteesListSearch.vue
+++ b/src/components/Editor/Invitees/InviteesListSearch.vue
@@ -46,6 +46,10 @@
 					<div v-if="option.type === 'circle' || option.type === 'contactsgroup'">
 						{{ option.subtitle }}
 					</div>
+					<div v-else-if="option.looksLikeMailingList" class="invitees-search-list-item__warning">
+						<AlertCircleOutline :size="14" />
+						{{ $t('calendar', 'This might be a mailing list. Invitations will not work.') }}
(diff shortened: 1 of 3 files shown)


tcitworld:
Maybe also check for the `KIND:group` vCard parameter as possibly given by the ContactsManager ?

If the mail app knows that an e-mail address is a mailing-list through e-mail headers, it could also save it when adding the contact to the recently contacted contacts.

miaulalala (approved):

kra-mo (approved):

kra-mo:
Good, but from a design side, the clipping text is obviously not ideal. Could the warning wrap and have sufficient space on the bottom?

ChristophWurst:
I did not find an option for that in https://vue-select.org/guide/options.html#options-prop, unfortunately.

kra-mo:
Can you add a hard line break, then? If not, the string should be shortened for now and this fixed upstream later.

For a shorter string, something like "List subscribers cannot accept invites."

Obviously not great for localization and a11y, but better than it being broken in default English.

odzhychko:
> Good, but from a design side, the clipping text is obviously not ideal. Could the warning wrap and have sufficient space on the bottom?

This is possible. I will update it the PR and mark it ready for review when done.

PR #8065, opened 2026-03-12
fix: default calendar alarm application

GVodyanov:
Continuation of https://github.com/nextcloud/calendar/pull/8031

### DIFF
--- a/src/components/Editor/Alarm/AlarmList.vue
+++ b/src/components/Editor/Alarm/AlarmList.vue
@@ -52,7 +52,11 @@ export default {
 		...mapStores(useCalendarObjectInstanceStore),
 		...mapState(useSettingsStore, ['forceEventAlarmType']),
 		alarms() {
-			return this.calendarObjectInstance.alarms
+			return this.calendarObjectInstance.alarms.slice().sort((a, b) => {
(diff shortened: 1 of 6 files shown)


SebastianKrupinski:
@GVodyanov 

Can you fix the DCO

src/components/Editor/Alarm/AlarmList.vue
```diff
@@ -52,7 +52,11 @@ export default {
 		...mapStores(useCalendarObjectInstanceStore),
 		...mapState(useSettingsStore, ['forceEventAlarmType']),
 		alarms() {
-			return this.calendarObjectInstance.alarms
+			return this.calendarObjectInstance.alarms.slice().sort((a, b) => {
+				const aIsDefault = a.alarmComponent.getFirstPropertyFirstValue('X-DEFAULT') ? 1 : 0
```
tcitworld: 1) We should add the vendor ID in the x-name for clarity, like `X-NC-DEFAULT` (see existing uses such as `X-NC-APPOINTMENT` in calendar or `X-NC-SCOPE` in vcard).

2) For even more clarity and even if we're already in an alarm component, I'd set it to `X-NC-DEFAULT-ALARM`

SebastianKrupinski: You make a good point sir!

Issue #8050, opened 2026-03-09
Cannot edit calendar event with external attendees (outside organization) — e.g. adding Talk conference fails


GretaD:
### Steps to reproduce

1. Open Nextcloud Calendar and create an event.
2. Add at least one external attendee (email from outside the organization, e.g. [user@gmail.com](mailto:user@gmail.com)).
3. Save the event.
4. Reopen the event and try to edit it — e.g. click "Add Talk conference" or change any other field.
5. Observe: error appears, changes are not saved.

### Expected behavior

i am able to save the event

### Actual behavior

an error is shown instead

### Calendar app version

6.2.0

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

chandrika1993:
Hey @GretaD , I ran into this bug too and had a dig around the code.
The issue is that when an event has external attendees, the CalDAV server tries to look them up as local users when saving and rejects the whole request when it can't find them. I also noticed that the calendar grid was not updating instantly after saving because the server does not return an ETag after a save, which causes an extra request before the UI refreshes. 
I will be happy to provide a fix if no one is yet working on it.

GretaD:
Hello @chandrika1993 a fix would be very much appriciated. Happy to review your PR :)

GretaD:
closing this ticket as not reproducible starting from the latest version of stable32

PR #8069, opened 2026-03-16
fix: cannot edit calendar event with external attendees (#8050)

chandrika1993:
fixes: #8050

@GretaD 
When an event has external attendees (e.g. user@gmail.com), the CalDAV server rejects the save because it tries to resolve those email addresses as local users and fails.

The fix adds SCHEDULE-AGENT=CLIENT to external attendee properties so the server knows not to handle scheduling for addresses it cannot resolve locally (RFC 6638 Section 7.1).

Also fixed the calendar grid not updating instantly after saving. The server does not return an ETag on 204 responses, which caused an unnecessary extra request before the UI could refresh. A background PROPFIND now restores the ETag immediately after saving, so the grid updates instantly.

### DIFF
--- a/src/components/Editor/Invitees/InviteesListSearch.vue
+++ b/src/components/Editor/Invitees/InviteesListSearch.vue
@@ -145,6 +145,7 @@ export default {
 							timezoneId: null,
 							hasMultipleEMails: false,
 							dropdownName: query,
+							scheduleAgent: 'CLIENT',
 						})
 					}
 				}
--- a/src/models/attendee.js
+++ b/src/models/attendee.js
@@ -31,6 +31,8 @@ function getDefaultAttendeeObject(props = {}) {
 		uri: null,
 		// Member address of the attendee
 		member: null,
+		// The SCHEDULE-AGENT parameter value for the attendee ('CLIENT' for external attendees)
+		scheduleAgent: null,
 		...props,
 	}
 }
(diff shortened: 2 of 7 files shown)


chandrika1993:
The build failure looks like a JavaScript heap out of memory error on the CI runner rather than an issue with the code changes. Could a maintainer take a look and re-run the failing jobs if needed? Happy to investigate further if it turns out to be related to the PR.

SebastianKrupinski:
Hi @chandrika1993 

Thank you for the PR. I will look through it soon.

chandrika1993:
> This looks wrong. The server should send invitations regardless if e-mails are matching local users or not. The fix should be on the server's side.

Thanks for the feedback! Agreed that a server-side fix would be the right long-term solution.

The reason for the client-side workaround is that the server is currently rejecting the entire PUT request when it can't resolve an external attendee, so users are completely blocked from editing those events. SCHEDULE-AGENT=CLIENT is intended as a temporary fix to unblock users until the server behaviour is corrected.

Happy to hear if there is a better approach!

chandrika1993:
Hi @GretaD @SebastianKrupinski  @tcitworld ,
I just wanted to check in on the status of this PR and see if there is any further input or testing needed from my side. I am happy to make any necessary adjustments to get this ready for the merge. Please let me know if I can help with anything else.

GretaD:
Hi @chandrika1993 thanks a lot for the PR and for taking the time to dig into this. We really appreciate your effort. 

We discussed this internally, and decided not to go forward with this approach. The main reason is that this is something we’d want to fix on the server side rather than introducing a client-side workaround, even if it helps in the short term.

Totally get your reasoning about unblocking users, but we’d prefer to avoid adding logic on the client that we’d likely need to remove again later once the server behavior is fixed.

Also, we were trying to reproduce the issue, and after the latest releases of the server and calendar, we cannot reproduce it anymore.

PR #8083, opened 2026-03-21
Fix Introduce a timezoneOffsetService for proposal and fix DST problem 

### LINKED ISSUE #8040: Summer time is not taken into account for meeting proposal
### Steps to reproduce

Well it's that time of year 🐝 when clocks turns and the calendar app is getting confused. 😕 
1. Create a meeting
2.  propose a date before summer change (29 th of march)
3.  See the correct time meeting
4. Create another meeting after the time change
5. see the one hour delay
6.

### Expected behavior

Meeting should be adjusted by the time change

### Actual behavior

Summer time support seems missing

Image

### Calendar app version

6.2.1

### CalDAV-clients used

_No response_

### Browser

Firefox

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

32.0.6

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

Sorry not an admin cannot do full report

taiebot:
Hello all,

Thank you all for this amazing project. I did another pull request which works as well but i think it might have been not that easy to review.

Trying to fix https://github.com/nextcloud/calendar/issues/8040
Be aware that this is my 1st time contributing to a project like Nextcloud and i am not a developer. This work is entirely done with an AI with vibe coding over weekends.  Test it carefully to see if it does not break anything else. I tried to be as close as the original code. 
I  introduce a timezoneOffsetService which allows to reuse it in the two parts of the code.

I managed to test this with https://github.com/szaimen/nextcloud-easy-test and what i could see and looks like it is fixed. :exploding_head: 

See original problem in #8040 

With this PR

Screenshot 2026-03-20 at 23-20-05 March 2026 - Agenda - Nextcloud
Screenshot 2026-03-20 at 23-19-43 March 2026 - Agenda - Nextcloud

On main
Screenshot 2026-03-20 at 23-45-58 March 2026 - Agenda - Nextcloud

Screenshot 2026-03-20 at 23-46-34 March 2026 - Agenda - Nextcloud

Ai comparison old vs new

The old code derives the timezone offset by converting dates to locale‑formatted strings and back, which is fragile and can mis-handle daylight‑saving transitions because the string round‑trip doesn’t reliably preserve DST rules. It also depends on the system’s locale parsing behavior, which varies across environments and can introduce subtle bugs. The new code avoids these pitfalls by using Intl.DateTimeFormat with formatToParts, which is explicitly designed to respect each timezone’s DST rules. Instead of relying on string parsing, it reconstructs the exact local time components and computes the UTC equivalent using Date.UTC, producing a precise offset even during DST boundaries. The new version caches formatters for performance, while the old version recreates them on every call. Error handling also improves: the old code returns 0 on failure, masking errors, whereas the new code returns null, making issues detectable. Overall, the new implementation is more robust, DST‑safe, locale‑independent, and efficient.

### DIFF
--- a/src/components/Proposal/ProposalResponseMatrix.vue
+++ b/src/components/Proposal/ProposalResponseMatrix.vue
@@ -137,6 +137,7 @@ import NcAvatar from '@nextcloud/vue/components/NcAvatar'
 import NcButton from '@nextcloud/vue/components/NcButton'
 import NcCheckboxRadioSwitch from '@nextcloud/vue/components/NcCheckboxRadioSwitch'
 import { Proposal, ProposalResponse } from '@/models/proposals/proposals'
+import { getTimezoneOffset } from '@/services/timezoneOffsetService'
 import { ProposalDateVote } from '@/types/proposals/proposalEnums'
 
 export default {
@@ -182,7 +183,6 @@ export default {
 	data() {
 		return {
 			ProposalDateVote,
-			timezoneOffset: 0,
 		}
 	},
 
@@ -201,17 +201,21 @@ export default {
 			const groups = {}
 			dates.forEach((d) => {
 				// Apply timezone offset for grouping by day
-				const key = d.date ? moment(d.date).utcOffset(this.timezoneOffset).format('yyyy-MM-dd') : 'invalid'
+				const offset = getTimezoneOffset(d.date, this.timezoneId)
+				const key = moment(d.date).utcOffset(offset).format('YYYY-MM-DD')
 				if (!groups[key]) {
 					groups[key] = []
 				}
 				groups[key].push(d)
 			})
-			return Object.entries(groups).map(([key, grp]: [string, ProposalDate[]]) => ({
-				key,
-				label: moment(grp[0].date).utcOffset(this.timezoneOffset).format('dddd, MMMM Do'),
-				dates: grp,
-			}))
+			return Object.entries(groups).map(([key, grp]: [string, ProposalDate[]]) => {
+				const offset = getTimezoneOffset(grp[0].date, this.timezoneId)
+				return {
+					key,
+					label: moment(grp[0].date).utcOffset(offset).format('dddd, MMMM Do'),
+					dates: grp,
+				}
+			})
 		},
 
 		columnCount() {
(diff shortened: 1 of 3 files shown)


GVodyanov (commented):
Hello! We really appreciate your PR for such an annoying issue, good job!

I see that you are having some struggles with ESLint, if you like I could fix that for you seeing as from what I can see vibe coding wasn't able to do that.

Just wanted to mention that we actually have a timezone library which might be useful here: https://github.com/nextcloud-libraries/timezones

SebastianKrupinski:
Hi @taiebot 

Thank you for the PR, I will review it when I get a few spare minutes

Aveyron-RetD:
It would be good if this could land before beginning of September to make sure that we do not encounter this bug again when sending proposals for after October 25th  !!!

src/services/timezoneOffsetService.js
SebastianKrupinski, line 1: This file should be in typescript.

src/components/Proposal/ProposalResponseMatrix.vue
```diff
@@ -222,39 +226,13 @@ export default {
...
-			} catch (e) {
-				// Fallback to UTC if timezone is invalid
-				return 0
-			}
-		},
-
 		dateTimeSpan(date) {
-			const startDate = moment(date).utcOffset(this.timezoneOffset)
-			const endDate = moment(date).utcOffset(this.timezoneOffset).add(this.proposal.duration, 'minutes')
+			const offset = getTimezoneOffset(date.toISOString(), this.timezoneId)
```
SebastianKrupinski: Instead of converting all the dates to strings, just make the getTimezoneOffset function accept a date variable. that way its is properly typed.

SebastianKrupinski (commented):
Once again thanks for the PR. Sorry it took so long to review, but we've been really busy with work for the new release.

src/components/Proposal/ProposalResponseMatrix.vue
taiebot: Done

SebastianKrupinski (approved):

SebastianKrupinski:
Fixed lint errors, rebased, squashed commits and adjusted commit message

PR #8078, opened 2026-03-20
Fix: Calendar embedding not works after migration to Vue 3

Koc:
Hey mates!

It seems like you had migrated to Vue 3 recently and now calendar embedding completely broken :see_no_evil: . This PR aims to fix that

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

PR #8086, opened 2026-03-23
fix(reference): migrate to Vue 3

mejo-:
The reference provider widget registration was not ported to Vue 3 yet.


tcitworld:
It seems to be the same as https://github.com/nextcloud/calendar/pull/8078 :)

mejo-:
> It seems to be the same as https://github.com/nextcloud/calendar/pull/8078 :)

Hehe uppsie, seems like @Koc was faster :wink:

Feel free to pick/close whichever you prefer

tcitworld:
The other one was merged

Issue #8049, opened 2026-03-09
Bug: Cannot confirm/create meeting from accepted Talk Meeting Proposal — error on confirmation


GretaD:
### Steps to reproduce

1. Open Nextcloud Calendar → create a Meeting Proposal (Appointment slots / poll).
2. Share the proposal link with participants.
3. Participants vote for one or more available time slots.
4. Organizer opens the proposal results and tries to confirm a time slot / create the meeting.
5. Observe: an error is shown and the meeting is not created.

### Expected behavior

Meeting is created

### Actual behavior

I get an error instead

### Calendar app version

6.2.0

### Browser

all of them

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text
[...]

PR #8092, opened 2026-03-25
fix(ProposalService): proposal creation

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/8049

I don't know much backend so if someone who knows this stuff better could review it, it would be appreciated.


SebastianKrupinski:
I'll look at it next week

SebastianKrupinski:
@GVodyanov

Right approach, but you can just reuse the existing functions (pushed the change).

PR #8123, opened 2026-04-01
fix: add event link

hamza221:
add a way to share a permanent link to an event 
image

ps: Set as a fix intentionally so it can be backported

👇🏼 needs https://github.com/nextcloud/calendar/pull/8048
image
image

### DIFF
--- a/lib/Controller/ObjectController.php
+++ b/lib/Controller/ObjectController.php
@@ -1,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\Calendar\Controller;
+
+use OCA\Calendar\Service\ObjectResolverService;
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http\Attribute\FrontpageRoute;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
+use OCP\AppFramework\Http\RedirectResponse;
+use OCP\IRequest;
+use OCP\IURLGenerator;
+
+class ObjectController extends Controller {
+
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		private ObjectResolverService $objectResolverService,
+		private IURLGenerator $urlGenerator,
+		private ?string $userId,
+	) {
+		parent::__construct($appName, $request);
+	}
+
+	/**
+	 * Resolve a permanent calendar object deep link.
+	 *
+	 * @param string $uid The iCalendar UID of the object
+	 * @param string|null $recurrenceId Unix timestamp of the recurrence instance, or null for 'next'
+	 */
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	#[FrontpageRoute(verb: 'GET', url: '/object/{uid}', postfix: 'uid')]
+	#[FrontpageRoute(verb: 'GET', url: '/object/{uid}/{recurrenceId}', postfix: 'uid.recurrenceId')]
+	public function index(string $uid, ?string $recurrenceId = null): RedirectResponse {
+
+		$resolved = $this->userId !== null
+			? $this->objectResolverService->findByUid($this->userId, $uid)
+			: null;
+
+		if ($resolved !== null) {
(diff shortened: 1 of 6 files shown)


lib/Controller/EventController.php
```diff
@@ -0,0 +1,99 @@
...
+
+				$editRecurrenceId = $recurrenceId ?? 'next';
+
+				return new RedirectResponse(
+					$this->urlGenerator->linkToRoute('calendar.view.indexdirect.edit.recurrenceId', [
+						'objectId' => $objectId,
+						'recurrenceId' => $editRecurrenceId,
+					])
+				);
+			}
```
hamza221, line 85: should be maybe moved to a service, so it can be reused by https://github.com/nextcloud/calendar/pull/8048

hamza221:
Also wondering if we should implement a method to search by event UID

appinfo/routes.php
```diff
@@ -14,6 +14,9 @@
 		['name' => 'view#index', 'url' => '/new/{isAllDay}/{dtStart}/{dtEnd}', 'verb' => 'GET', 'postfix' => 'direct.new.timerange'],
 		['name' => 'view#index', 'url' => '/edit/{objectId}', 'verb' => 'GET', 'postfix' => 'direct.edit'],
 		['name' => 'view#index', 'url' => '/edit/{objectId}/{recurrenceId}', 'verb' => 'GET', 'postfix' => 'direct.edit.recurrenceId'],
+		// Events
+		['name' => 'event#index', 'url' => '/event/{uid}', 'verb' => 'GET', 'postfix' => 'event.uid'],
+		['name' => 'event#index', 'url' => '/event/{uid}/{recurrenceId}', 'verb' => 'GET', 'postfix' => 'event.uid.recurrenceId'],
```
SebastianKrupinski, line 19: I would make this a generic route "/object/{uid}" or something like that, this way this can be used for both Events and Tasks

lib/Controller/EventController.php
```diff
@@ -0,0 +1,97 @@
...
+
+	/**
+	 * Resolve a permanent event deep link.
+	 *
+	 * Searches all calendars for an event with the given UID and redirects
+	 * to the standard edit route. Falls back to serving the SPA if the
+	 * event cannot be resolved server-side.
+	 *
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
SebastianKrupinski, line 48: Please use the new route attributes

appinfo/routes.php
SebastianKrupinski: Also can we use the new route attributes instead of the routes file?

lib/Controller/EventController.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+class EventController extends Controller {
+
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		private CalendarInitialStateService $calendarInitialStateService,
+		private IManager $calendarManager,
+		private IURLGenerator $urlGenerator,
+		private ?string $userId,
```
SebastianKrupinski, line 35: Use a userSession

lib/Controller/EventController.php
```diff
@@ -0,0 +1,97 @@
...
+	 *
+	 * @param string $uid The iCalendar UID of the event
+	 * @param string|null $recurrenceId Unix timestamp of the recurrence instance, or null for 'next'
+	 * @return Response
+	 */
+	public function index(string $uid, ?string $recurrenceId = null): Response {
+		if ($this->userId === null) {
+			$this->calendarInitialStateService->run();
+			return new TemplateResponse($this->appName, 'main');
+		}
```
SebastianKrupinski, line 58: Why?

appinfo/routes.php
hamza221: I never worked with tasks, would they also use the same logic to resolve/ redirect ?

SebastianKrupinski: Yes, VTODO (tasks) are identical for the most part to a VEVENT, they have a UID and a RECURRANCE-ID, the main differences are VTODO does not have a DTEND it uses a DUE and DURATION.

hamza221: Makes sense, I'll update it then

lib/Controller/EventController.php
mejo-: That would be helpful indeed 😊

SebastianKrupinski (approved):

hamza221:
Don't merge please, I'll try to update the pr asap

hamza221:
image
image
works for shared and owner ✅

SebastianKrupinski (approved):

PR #8157, opened 2026-04-12
fix: modifying a proposal

taiebot:
Current code is wrong because:

    $currentDates[$key] = the old date from the database

    $date = the new mutated date from the client

So when we detect a modification, we are returning the old object instead of the new one.

On main
create a proposal. (save)
exit
re-enter Try to modify it => nothing happens (fail silently)

With pull request
create a proposal (save)
exit
Modify it => proposal is modified accurately


SebastianKrupinski (approved):
Tested. Looks good! 

Not sure how I missed this during final testing, I guess I never dragged a date to a different time.

Thank you!!!

PR #8164, opened 2026-04-17
fix: add an extra confirmation step to appointment booking

hamza221:
To prevent email bots from auto-confirming 
image


SebastianKrupinski:
Tested works.

SebastianKrupinski (approved):

PR #8170, opened 2026-04-20
fix(Editors): auto focus title inputs

GVodyanov:
Auto focuses this:
image

### DIFF
--- a/src/directives/focus.js
+++ b/src/directives/focus.js
@@ -4,9 +4,10 @@
  */
 
 export default {
-	inserted(el) {
+	mounted(el) {
 		setTimeout(() => {
-			el.focus()
-		}, 500)
+			const input = el.querySelector('input') || el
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

src/directives/focus.js
```diff
@@ -4,9 +4,10 @@
  */
 
 export default {
-	inserted(el) {
+	mounted(el) {
 		setTimeout(() => {
-			el.focus()
-		}, 500)
+			const input = el.querySelector('input') || el
+			input.focus()
```
ChristophWurst, line 10: el no longer works for focus?

GVodyanov: We had to pass from pure HTML `input`s to the Nextcloud Vue ones, so it was focusing the wrapper div and not the actual input inside.

PR #8176, opened 2026-04-20
fix(talk-integration): make sure talk is enabled for user

hamza221:
Talk app can be granularly enabled for specific user groups which is not respected by the integration in calendar. 
Introducing an OCP api for talk could also be a solution here or as a followup . 
We only use talk to create talk rooms afaik hence checking both configs. 

Admin settings -> talk 
image


SebastianKrupinski (approved):

SebastianKrupinski:
The solution look correct to me.

hamza221:
I think I'll migrate this to OCP then update the PR 
Update: I'll do it as a followup, since server merge/backport/release might take a while

PR #8183, opened 2026-04-22
chore: load module mocks reliably in JavaScript unit tests

odzhychko:
This fixes failing CI runs caused by Vitest's module cache.

This should fix the issue, that "Note tests" randomly fail in CI.
e.g. https://github.com/nextcloud/calendar/actions/runs/24769753480

### DIFF
--- a/tests/javascript/unit/setup.js
+++ b/tests/javascript/unit/setup.js
@@ -3,6 +3,64 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+beforeAll(() => {
+	// Disclaimer: The following explanation might be inaccurate 
+	// because it is a result of encountering this issue for the first time. 
+	// Still, the explanation is consistent with documentation for Vitest 4
+	// and experiments to reproduce this issue.
+    // So the following explanation is likely to reflect what is actually happening.
+	//
+	// Modules are cached across test files.
+	// > Resets modules registry by clearing the cache of all modules.
+	// See https://vitest.dev/api/vi.html#vi-mock
+	//
+    // So if modules are not reset before each test file run, 
+	// the following issue can occur:
+	//   - Given
+	//     - Module X imports module Y
+	//     - Test file A
+	//       - First imports the module X and does not mock it
+	//       - Secondly imports the module Y and does not mock it
+	//     - Test file B
+	//       - First imports the module X and not mock it directly
+	//       - Secondly imports the module Y and does mock it
+	//   - When
+	//     - Test file A is executed before test file B on the same worker
+	//   - Then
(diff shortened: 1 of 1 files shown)


SebastianKrupinski:
Hi @odzhychko 

Thanks for the PR.

So this is interesting... But I am not sure that cached mocks is the cause of the failures... the "node test" seem to mostly fail when our CI is busy... In the evenings, my time (EST) the "node tests" hardly ever fail, and if they do restarting the test makes them pass. 

If the cause was cached mocks this would happen all them time. Any thoughts?

odzhychko:
@SebastianKrupinski  I'm mostly sure of my explanation.
But still, it is the first time I have encountered something like this.
So I'm glad you are double-checking.

---

> "node test" seem to mostly fail when our CI is busy.

That could make sense.

It appears that workers can be "recyled".
(Docs for [`vmMemoryLimit`](https://vitest.dev/config/vmmemorylimit.html#vmmemorylimit) mention recycling.)
Probably modules are kept in cache when a worker is reused.

A busy CI might influence the speed of workers or maybe even how workers are recycled.
Depending on how busy the CI is, tests could be executed on different workers.
So sometimes tests are executed in a way that module caching has no effect, and sometimes in a way that makes the tests fail.

One of my guesses on why it works better when the CI is not busy is that all workers run fast because of enough CPU resources and therfore tests files are distributed evenly across workers. 
On the other hand, a busy CI would cause more test files to be executed on the same worker, raising the chance for caching related issues.

---

> If the cause was cached mocks this would happen all them time.

It only happens if tests files are executed in an order in which caching makes a difference.

I created a [demo where it always fails](https://github.com/odzhychko/calendar/tree/test-javascript-load-module-mocks-reliably-demo/tests/javascript/unit). The setup is:

- Setting `maxWorkers: 1`
- Having two tests and each breaks if the other one runs before it.

---

While making the demo, I noticed that there is a difference between directly imported and transitively imported modules.
The issues seems to affect modules that are also transitively imported.
I extended the comment in the code.
I also added a disclaimer that the explanation is just an educated guess, because I'm not fully sure.

SebastianKrupinski:
Okay, lets merge this an see what happens, when its used more frequently.

I ran all the workflows this morning which is usually a busy time on the CI and the node tests didn't fail but that could just be a coincidence. 

Only one change... can you please change the commit message to "chore: {what you fixed}" instead of "test(javascript):", we have release automation that depend on certain commit message string.

SebastianKrupinski (approved):

SebastianKrupinski:
Just wanted to say thank you! 

You seems to have solved the issue with this PR.

Issue #7881, opened 2026-01-20
A11Y initiative 2026


ChristophWurst:
### Steps to reproduce

1. Use the app with screen reader, low contrast, keyboard only, etc

### Expected behavior

Usable UI

### Actual behavior

Some elements are not usable

### Web server

None

### Database engine version

None

### PHP engine version

None

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

GVodyanov:
The datepicker is also currently not usable without a keyboard, but that will be fixed with: https://github.com/nextcloud/calendar/pull/7697

GVodyanov:
Related https://github.com/nextcloud/calendar/pull/8244

PR #8244, opened 2026-05-01
fix: meeting proposal accessibility

GVodyanov:
Related to https://github.com/nextcloud/calendar/issues/7881

Improvements:
- Layout now is usable on mobile
- Tab navigation fully works on all elements
- All elements have proper focus indication
- Works with a screen reader
- You can fully operate all the components solely with a keyboard, even to select timeslots through the fullcalendar component (arguably might be a little over the top idk)

### DIFF
--- a/src/components/Proposal/ProposalDateItem.vue
+++ b/src/components/Proposal/ProposalDateItem.vue
@@ -8,13 +8,23 @@
 		<div class="proposal-date__icon">
 			<ItemIcon />
 		</div>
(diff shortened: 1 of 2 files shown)


SebastianKrupinski:
Tested. there are some glitches tho.

GVodyanov:
TODO:
- Re-order date input field under participants
- Show date input field only on mobile
- Make calendar picker be visible only on desktop by default

PR #8264, opened 2026-05-05
feat(resources): browsable room finder with availability and dropdown filters

Rikdekker:
Reworked on top of the design in #8718, against current main.

The "Show rooms" dialog listed every room in a flat table with a button per row
to check its availability. With more than a handful of rooms that is not
browsable: no way to narrow down by building, capacity or features, and no
indication of what is free.

The dialog now contains a room browser: rooms grouped per building, each card
showing capacity, room number and whether the room is free for the time range
of the event. Filters for building, floor, minimum capacity and features narrow
the list down. A room is picked by clicking its card; nothing is written to the
event until "Done" is pressed, so closing the dialog any other way discards the
choice.

### Scope, as agreed in your comment

- Only the contents of the modal changed. The quick search, the list of already
  selected rooms and resources, and the suggestions are untouched — the diff on
  `ResourceList.vue` is four lines added, five removed.
- Rooms only. The Resources tab from the design is left for a follow-up.
- "Unavailable" is the only status shown. Nothing was factored out of
  `AvatarParticipationStatus.vue`; #8589 stays as it is.

### Structure

Search and filter logic is plain TypeScript in `src/utils/roomFilter.ts`,
wrapped in a `useRoomFilter(allRooms)` composable, so it is unit tested without
mounting a component. Types live in `src/types/models/roomFilter.ts`.

```js
const { buildingOptions, selectedBuilding, minimumSeatingCapacity,
        roomsFiltered, groupedRooms, searchText, resetFilters } = useRoomFilter(allRooms)
```

### Two things derived in the frontend on purpose

Both per your feedback on #8263:

- **Building name** is the first non-postal-code segment of the building
  address. Rooms carry no building name, so this is a heuristic, deliberately
  kept out of the principal model. Skipping postal codes matters in practice: on
  our test instance 19 of 113 rooms have `1098 XG, Amsterdam` as their entire
  address, which would otherwise group them under "1098 XG".
- **LOCATION** is built here rather than taken from `roomAddress`, which joins
  room number, story and address in that order and returns an empty string
  rather than `null` when unset.

### One deviation from the design

Each card keeps a small action that opens the existing free/busy timeline
("Find a time"). The design drops it, but it is the one thing today's dialog
does that the browser would otherwise lose. Happy to remove it if you would
rather keep step one strictly to the mockup.

roompicker
rooms-popup
select-room-availability

### Naming

The dialog is titled "Rooms" rather than "Rooms and resources", to match the
rooms-only scope. When the Resources tab lands, the dialog title, the "Show
rooms" button and the "Resources" section heading above it should be aligned in
one go — that heading already covers both rooms and resources today.

### Tests

48 unit tests across `roomFilter`, `useRoomFilter` and `attendee`. `npm run
ts:check` reports the same 47 pre-existing errors as clean main — none in the
new files. eslint and stylelint clean.

Verified manually on a Nextcloud 34 instance with 113 rooms from a real import:
filtering, grouping, selection, Done, LOCATION, round-tripping the selection,
re-checking availability on a time change, and the free/busy timeline.

Split out of #7996 as suggested by @nimishavijay; the principal mapping it
depended on landed separately in #8693.

## 🤖 AI (if applicable)


SebastianKrupinski:
Hi @Rikdekker 

Thank you for the PR. I will review it as soon as I can.

osm-frasch:
Thank you for working on this improvement.

I work for a non-profit educational and social services provider in Germany. We are currently evaluating RoomVox in our managed Nextcloud Enterprise environment, and this browsable room finder would make room booking significantly more intuitive and useful for our staff.

As we cannot apply custom patches to the Calendar app in our Enterprise environment, we would highly appreciate seeing this functionality reviewed and, if accepted, included in an official Nextcloud Calendar release.

Many thanks for your work on this.

osm-frasch:
Sorry – just a quick follow-up: has there been any progress on this matter?

osm-frasch:
> Sorry – just a quick follow-up: has there been any progress on this matter?
Sorry—nothing new?

src/components/Editor/Resources/ResourceList.vue
```diff
@@ -1,243 +1,630 @@
...
+			<div
+				v-for="group in groupedRooms"
+				:key="group.name"
+				class="resource-picker__group">
+				<button
+					class="resource-picker__group-header"
+					@click="toggleGroup(group.name)">
+					<ChevronDown
+						v-if="expandedGroups[group.name]"
+						:size="18" />
```
GVodyanov:
```suggestion
						:size="20" />
```

src/components/Editor/Resources/ResourceList.vue
```diff
@@ -1,243 +1,630 @@
...
+			display: flex;
+			align-items: center;
+			gap: calc(var(--default-grid-baseline) * 3);
+		}
+	}
+
+	&__capacity {
+		display: flex;
+		align-items: center;
+		gap: 4px;
```
GVodyanov:
```suggestion
		gap: var(--default-grid-baseline);
```

GVodyanov:
Thank you for your PR, could you please use our CSS variables for styling https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/css.html

src/components/Editor/Resources/ResourceList.vue
Rikdekker: Done in bec668d0d — both chevron icons now use `:size="20"` to match the section `MapMarker`.

Rikdekker: Done in bec668d0d — both chevron icons now use `:size="20"` to match the section `MapMarker`.

src/components/Editor/Resources/ResourceList.vue
Rikdekker: Done in bec668d0d — this and every other hardcoded spacing value in both components now use `var(--default-grid-baseline)` (baseline = 4px).

Rikdekker:
Thanks for the review @GVodyanov! I've pushed `bec668d0d` addressing the CSS feedback:

- All hardcoded pixel spacing in `ResourceList.vue` and `ResourceRoomCard.vue` now uses `var(--default-grid-baseline)` (e.g. `gap: 4px` → `gap: var(--default-grid-baseline)`, `padding: 6px 10px` → `calc(var(--default-grid-baseline) * 1.5) calc(var(--default-grid-baseline) * 2.5)`, etc.).
- Dropped the hardcoded hex colour fallbacks — status colours now use the bare `--color-success-text` / `--color-error-text` server variables.
- Normalised the group-header chevrons to `:size="20"` to match the section `MapMarker` icon.

The only raw `px` values left are the 1–3px hairline/accent border widths, for which there's no grid variable and which the existing Calendar components also use as literals. Happy to convert those too if you'd prefer.

src/components/Editor/Resources/ResourceList.vue
```diff
@@ -1,243 +1,630 @@
 <!--
-  - SPDX-FileCopyrightText: 2021 Nextcloud GmbH and Nextcloud contributors
```
SebastianKrupinski, line 2: Copy rights should not be changed on existing files.

src/components/Editor/Resources/ResourceList.vue
```diff
@@ -1,243 +1,630 @@
...
-		hasUserEmailAddress() {
-			const emailAddress = this.principalsStore.getCurrentUserPrincipal?.emailAddress
-			return !!emailAddress
-		},
-
-		isViewedByOrganizer() {
-			if (!this.calendarObjectInstance.organizer) {
-				return true
-			}
+interface RoomPrincipal {
```
SebastianKrupinski: All interfaces should be separate files, so they can be reused

vitest.config.js
SebastianKrupinski, line 1: Why are we changing the vite config? These changes seem unrelated.

src/components/Editor/Resources/ResourceList.vue
odzhychko: Partially done in 3ec5a62. During a full review, I will check what interfaces can be really reused and which are only internal to a component (e.g., only used internally or used to type the component interface (props, emits, etc.)).

odzhychko:
Hi @Rikdekker, I took over reviewing this and the related PR.
Sorry for the inconvenience of adding the reviewers in the middle of a longstanding PR.
But I'm currently the one that can dedicate attention to this feature :smile: 

We very much like the improvement in browsing rooms.

I am currently collecting input on how we can integrate this feature in a way that aligns with the current Nextcloud design language and upcoming features in the editor interface.
You will hear back from me soon, hopefully with a very concrete mockup.

odzhychko:
Hi @Rikdekker 

We prepared the design #8718 as something, we would directly incorporate into Nextcloud.
Would you like to reconcile your logic with the new design and update this (or open a new PR)?

Some nice to haves, but not required, would be:

- extract the logic to build filter-options (incl. translations) into own code (plain TypeScript)
- extract the logic for search + application of filters into its own code (plain TypeScript)
- create a composable to encapsulating search/filter options + state + results from the actual Vue component
  - e.g., `const { buildingOptions, selectedBuilding, minmialSeatingCapacity, roomsFiltered, searchText, ...  } = useRoomFilter(allRooms)`

This all would make it easier test and maybe reusable in other places.


With regards to https://github.com/nextcloud/calendar/pull/8263#issuecomment-5225063231 :

- Lets use your original heuristic for building name here
- You have good points with regards to the format in LOCATION
  - feel free to ignore `roomAddress` and use the logic you had to construct something sensible
  
Look out for:
  
- In the first step lets only change the contents of the Modal
    - Keep the quick search and overview of already selected rooms and resources as is in the editor view
- In the first step only have the "Rooms" part of the modal
    - "Resources" can be cleanly added in next steps (including data mapping, search logic, select handling, status etc.)
- In the first step let's keep only showing the status to "Unavailable"
    - https://github.com/nextcloud/calendar/pull/8589 just added cool logic about checking on change and adding more useful info
       - But factoring that out from `AvatarParticipationStatus.vue` would be better left to a separate PR.
       - Also might it might needs some though because in the modal we select a room only after clicking "Done".

Rikdekker:
@odzhychko Reworked on the new design and force-pushed. Description updated.

All five "look out for" points are in: modal contents only, rooms only,
"Unavailable" only, editor view untouched, `AvatarParticipationStatus.vue` not
touched. The three nice-to-haves as well — filter options and search/filter
logic are plain TypeScript, wrapped in `useRoomFilter(allRooms)`.

Note the branch was rebuilt on current main rather than rebased, so your three
fixups from 3 August are no longer in its history — #8693 landed the model layer
they were correcting, and the UI they touched was rewritten for the new design.

Two things worth your opinion:

1. I kept a per-card action that opens the existing free/busy timeline. It is
   not in the design, but it is the only thing the current dialog does that
   would otherwise disappear. Say the word and it goes.
2. Building name: your point about guessing was right in a way I only saw with
   real data. Taking the segment before the first comma groups 19 of our 113
   rooms under "1098 XG", because that is their entire stored address. It now
   skips postal codes, so those land under "Amsterdam" — better, still a guess.
   I would like to take you up on `BUILDING_NAME` in `IRoomMetadata`; our own
   room backend holds the building as a separate field and has nowhere to put
   it. I will open that as an issue on server first so it can be discussed
   before any code.

## 🤖 AI (if applicable)

- [x] The content of this comment was partly or fully generated using AI

Rikdekker:
The failing php/psalm checks are unrelated to this PR — it changes no PHP at
all. They run against server master and fail on `getBackends()` mocks in tests
this branch does not touch; #8589 had the same checks green on 11 August.

Rikdekker:
Some context on the building-name heuristic that may be useful for reviewing it — it does not change the code here.

The heuristic stays as agreed. Worth knowing though that it is a fallback rather than the only route: room backends *can* supply a building name directly. RoomVox 1.3.0 (released today) now publishes `{http://nextcloud.com/ns}room-building-name` from its own building field instead of leaving clients to infer one from the address, and it also fixes the floor, which it previously published under a key no client requests (`room-building-floor` rather than `IRoomMetadata::BUILDING_STORY`). On our test instance 97 of 113 rooms had a floor set that no client could see.

So for rooms from a backend that fills those properties, grouping and floor filtering run on real data; for everything else the heuristic here keeps doing its job. That is why skipping postal-code segments still matters — it is exactly the case where a backend gives us nothing better to work with.

There is no `IRoomMetadata::BUILDING_NAME` constant on the server side yet, so the key is currently published as a plain string; clients that do not know it simply ignore it. Adding it upstream would be the natural next step, but that is out of scope here.

Separately, @osm-frasch above mentioned evaluating RoomVox and not being able to patch Calendar in their Enterprise environment — 1.3.0 is on the App Store now, so the metadata half of what they need no longer requires a patch. The room browser in this PR is the other half.

odzhychko:
@Rikdekker Thx for reconciling the mockup and factoring out (incl. testing) filtering.
I'm taking over cleaning (code organization, some UI stuff) to bringing into `main` in #8731

Issue #7941, opened 2026-01-30
HTTP 500 when converting Meeting Proposal to Event: SQL Unique Constraint Violation on UID


shbb-asdea:
### Steps to reproduce

    Create a new Meeting Proposal (Poll) in the Calendar web interface.

    Add participants and dates.

    Have a participant (or yourself) vote for a date.

    Open the proposal as the owner.

    Click the "Create" (or "Crea") button on the confirmed date to finalize the meeting.

### Expected behavior

The "Meeting Proposal" (draft) should be deleted/converted, and a new "Event" should be created in the calendar with the details from the proposal.

### Actual behavior

The UI shows a loading spinner that eventually stops or fails. The browser console shows a 500 Internal Server Error on the POST /apps/calendar/v1/proposal/convert (or similar endpoint). The backend logs show a SQLSTATE[23505]: Unique violation error regarding calobjects_by_uid_index.

### Calendar app version

6.1.5

### CalDAV-clients used

N/A (Issue occurs in Nextcloud Web Interface)

### Browser

Firefox 147.0 (but issue is server-side) 

### Client operating system

Ubuntu Linux (but issue is server-side)

### Server operating system

Linux (Docker Container)

### Web server

Other

### Database engine version

PostgreSQL

### PHP engine version

PHP 8.4

### Nextcloud version

32.0.3

### Updated from an older installed version or fresh install

Fresh install

### List of activated apps

```Text
- calendar: 6.1.5
- calendar_resource_management: 0.11.0
- spreed: (enabled, version unknown from provided snippet but active)
...and others standard for Docker image
```

### Nextcloud configuration

```Text
Standard Docker configuration.
```

### Web server error log

```Text
192.168.1.33 - - [30/Jan/2026:11:40:15 +0000] "POST /ocs/v2.php/calendar/proposal/convert HTTP/1.1" 500 823 "-" "Mozilla/5.0 ..."
```

### Log file

```Text
{
  "reqId": "...",
  "level": 3,
  "time": "2026-01-30T11:40:15+00:00",
  "remoteAddr": "...",
  "user": "boris.shestakov",
  "app": "index",
  "method": "POST",
  "url": "/ocs/v2.php/calendar/proposal/convert",
[...]

PR #8284, opened 2026-05-08
fix(proposal): avoid UID collision when converting a proposal without…

shbb-asdea:
… a known blocker

convertProposal assigned the proposal's UUID to the new VEVENT before deciding between "PUT over an existing blocker" and "INSERT a new calendar object". When findCalendarBlocker missed the blocker (calendar not enumerated by the principal lookup, search backend lag, etc.) the INSERT collided with oc_calendarobjects.calobjects_by_uid_index because the orphan blocker still held the same UID, surfacing as a 500 with SQLSTATE[23505] and leaving the proposal unconverted.

Defer the UID assignment until the blocker lookup result is known: when a blocker is found, keep reusing the proposal UUID so the in-place PUT preserves iCal threading; when no blocker is found, allocate a fresh UUID so the INSERT cannot clash with a stale row.

Refs: nextcloud/calendar#7941

### DIFF
--- a/lib/Service/Proposal/ProposalService.php
+++ b/lib/Service/Proposal/ProposalService.php
@@ -341,7 +341,9 @@ class ProposalService {
 		$vObject = new VCalendar();
 		/** @var \Sabre\VObject\Component\VEvent $vEvent */
 		$vEvent = $vObject->add('VEVENT', []);
-		$vEvent->UID->setValue($proposal->getUuid() ?? Uuid::v4()->toRfc4122());
+		// UID is assigned below depending on whether an existing blocker is reused (PUT in place)
+		// or a fresh calendar object is created. Reusing the proposal UUID for a fresh INSERT
+		// would collide with an orphan blocker held by oc_calendarobjects.calobjects_by_uid_index.
 		$vEvent->add('DTSTART', $eventTimezone ? $selectedDate->getDate()->setTimezone($eventTimezone) : $selectedDate->getDate());
 		$vEvent->add('DTEND', (clone $vEvent->DTSTART->getDateTime())->add(new \DateInterval("PT{$proposal->getDuration()}M")));
 		$vEvent->add('SEQUENCE', 1);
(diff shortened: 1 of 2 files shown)


SebastianKrupinski:
Hi @shbb-asdea 

Please link a proper issue ticket, the reference link currently, links to a dependency update PR

shbb-asdea:
@SebastianKrupinski apologies for the confusion — I had pasted the wrong issue number. The PR description is now updated to reference the correct issue (#7941). Could you take another look?

SebastianKrupinski:
The issue you linked should have been resolved by this PR https://github.com/nextcloud/calendar/pull/8092

The issue was the original code would delete the existing event, but due to our trash bin logic the event was not actually deleted, so when the original code created a new event there was a collision in the index.

But the above PR resolves this, by updating the blocker, is this still an issue?

shbb-asdea:
Hi @SebastianKrupinski, thanks for pointing to #8092 — I dug into its diff and the relationship with this issue is clearer now.

#8092 (which closed #8049 — same symptom: duplicate key on `calobjects_by_uid_index` during proposal convert) changed the convert flow from "delete-blocker + create-event" to "find-blocker + update-in-place via PUT". That removed the collision on the path where the blocker is found — PUT keeps the URI and just rewrites the row, so the unique index is never re-touched. 

But the same PR introduced an `else` branch ([`ProposalService.php#L369-L377` on `main`](https://github.com/nextcloud/calendar/blob/main/lib/Service/Proposal/ProposalService.php#L369-L377)):

```php
$result = $this->findCalendarBlocker($user, $proposal);
if ($result !== null) {
    $this->applyCalendarBlockersOrganizer(...);   // PUT — safe after #8092
} else {
    $userCalendar->createFromString(...);          // INSERT — still collides
}
```

When findCalendarBlocker ([L796-L805](https://github.com/nextcloud/calendar/blob/main/lib/Service/Proposal/ProposalService.php#L796-L805)) returns null but a blocker row still exists in oc_calendarobjects, the else branch INSERTs with the proposal's UID and crashes on calobjects_by_uid_index exactly the way #8049 / #7941 described.

findCalendarBlocker is built on getCalendarsForPrincipal() + $calendar->search('', [], ['uid' => …]). The unique index is on (calendarid, calendartype, uid) — deleted_at is not part of it. So a blocker that has been trash-binned (or sits in a calendar/state the principal lookup or the search backend doesn't surface) is invisible to findCalendarBlocker, but still claims the UID in the table. That's why #8049 / #7941 keep reproducing after #8092: when the trash-bin row exists, we fall through to the else branch and the same collision happens.

This is consistent with the report by @sobeaa in this thread, where the crash still reproduces on 32.0.6 / Calendar 6.2.0 — i.e. with #8092 already merged.

The regression test added in this PR (testConvertProposalUsesFreshUuidWhenBlockerNotFound) reproduces the failure deterministically by mocking getCalendarsForPrincipal as empty (the precise "blocker invisible to the lookup" condition). It fails on current main and passes after the patch.

The fix is small: defer UID assignment until the lookup result is known.

Blocker found → reuse proposal UUID (PUT in place — unchanged behaviour from #8092, iCal threading preserved).
Blocker not found → fresh UUID for the INSERT, which therefore can't collide with a stale or trash-binned row.
Happy to switch to an alternative if you'd prefer — e.g. deleting the stale row through the DAV layer before INSERT, or extending findCalendarBlocker to also surface trash-binned objects — but a fresh UUID seemed the smallest, most defensive change and preserves the happy-path behaviour #8092 set up.

SebastianKrupinski:
Closing as irrelavent

Issue #146, opened 2016-10-20
Month view: scroll down when current day is not visible 


georgehrke:

georgehrke:
This is actually quit tricky, because the calendar events are added after the actual view was already rendered. :/

What if the user scrolled before all events were rendered already?

tcitworld:
Could we have a kind of « loading » animation for the user to wait until the events are fetched and rendered ?

georgehrke:
> Could we have a kind of « loading » animation for the user to wait until the events are fetched and rendered ?

This way we would be blocking user interaction without a real need to. Blocking user interaction should be kept to a minimum, better not at all.

Anyway, there have been changes with regard to event rendering in full calendar 3.1.
Let's postpone this to 1.6, so we don't have to implement it twice.

georgehrke:
There is currently no easy way to do this with full calendar. Moving to backlog.

jancborchardt:
Proposing this for feature planning @ChristophWurst as it is a consistent issues in the last 2 weeks of every month.

szaimen:
Maybe this helps in figuring out a solution to this: https://stackoverflow.com/a/32768735

szaimen:
🎉🎉🎉🎉🎉

PR #8306, opened 2026-05-12
fix: Scroll month view to current day so it’s visible

### LINKED ISSUE #8016: Month view: auto-scroll to current week
In month view, the calendar always shows the full month starting from row 1. During the second half of the month, the current day and upcoming events are below the fold, requiring manual scrolling every time. It would be helpful if month view auto-scrolled so the current week is near the top of the view, keeping the most relevant dates visible without scrolling.

jancborchardt:
When in the month view of the current month, the current day might often be very far down or even outside of the viewport. This is an issue especially for filled calendars (like when our vacation calendar is enabled), or if the setting "Limit number of events shown in month view" is disabled, or if it’s at the end of a month.

Was thinking this could be fixable by cross-checking which element has the class "fc-day-today" and then using the id of the a element inside (for example id="fc-dom-32") to navigate to it via a simple hash link.

Fix longstanding issue https://github.com/nextcloud/calendar/issues/146 and fix https://github.com/nextcloud/calendar/issues/8016

No video/screenshot as I used an export of our internal calendars since they are so full. ;)

---

AI-assisted: Claude Code (Sonnet 4.6)

Adds a loading callback to the FullCalendar options that fires once all event sources have finished fetching – at which point row heights are final. When loading completes in dayGridMonth view and today is in the visible range, the `.fc-day-today` cell is scrolled into view using `scrollIntoView({ block: 'nearest', behavior: 'smooth' })`, which only scrolls if the element is actually outside the viewport.

### DIFF
--- a/src/components/CalendarGrid.vue
+++ b/src/components/CalendarGrid.vue
@@ -170,6 +170,7 @@ export default {
 				droppable: true,
 				eventReceive: this.handleEventReceive,
 				eventShortHeight: 38,
+				loading: this.scrollMonthViewToToday,
 			}
 		},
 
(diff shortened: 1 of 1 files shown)


SebastianKrupinski (approved):
Tested. Works!

PR #8311, opened 2026-05-12
chore: Remove unused files, duplicate settings for eslint / stylelint and blank JSDoc blocks

DerDreschner:
This PR makes 5 changes:
- Removes `.stylelintignore` as we don't have a `src/fonts` folder anymore
- Enables eslint settings that were marked as `TODO` and are now possible with the use of `eslint-plugin-jsdoc`
- Run fixer to remove blank JSDoc blocks
- Removes stylelint settings that are now being handled by the `@nextcloud/stylelint-config` package
- Use `export default` instead of `module.exports` for stylelint

~~Note: `@nextcloud/eslint-config` doesn't use and expose the `eslint-plugin-jsdoc` for anything other then documentation. That's why I had to manually install it as dev dependency here in order to use the removed `require-jsdoc` / `require-param` settings and the new `no-blank-blocks` setting.~~

Edit: The real issue was the incorrect object being passed into the rule. Now I use the `eslint-plugin-jsdoc` that's being shipped through `@nextcloud/eslint-config`.


eslint.config.mjs
```diff
@@ -3,25 +3,30 @@
...
 			'no-console': 'off',
 			'@typescript-eslint/no-unused-vars': 'off',
 			'vue/multi-word-component-names': 'off',
+		},
+	},
+	jsdoc({
+		files: filesToLint,
+		rules: {
 			// JSDocs are welcome but lint:fix should not create empty ones
 			'jsdoc/require-jsdoc': 'off',
```
DerDreschner: I guess we can set this to `error` or `warning` now, as `lint:fix` will not create empty ones unless we set `enableFixer: true` (like for `no-blank-blocks`).

SebastianKrupinski: Set it to Warn only

DerDreschner: Done

SebastianKrupinski (approved):

PR #8365, opened 2026-05-20
fix(models): markRaw calendar-js objects in Pinia state

### LINKED ISSUE #8364: Removing an attendee in the web UI silently fails to persist - Vue 3 Proxy bypasses calendar-js identity check
### Steps to reproduce

1. As an organizer, create a normal event (single-shot or recurring - both reproduce) and add at least one attendee (e.g. another local Nextcloud user). Save.
2. Reopen the event in the web UI editor.
3. Click the three-dot menu next to the attendee row → **Remove attendee**. The row disappears from the editor. The Context Menü remains open, but updated to next attendee. (Latest not the core of the issue)
4. Click **Update** (or for recurring events: **Update this occurrence** / **Update this and all future**).

### Expected behavior

- The web UI close the Context Menü after clicking "Remove attendee" (Not the core of the issue)
- The PUT body sent to CalDAV contains the modified VEVENT without the removed attendee.
- The server stores the change.
- Reopening the event shows the attendee gone.
- The removed attendee receives an iTip `CANCEL`.

### Actual behavior

- The editor closes without error and the attendee is gone from the visible list **until the next reload**.
- The PUT body the browser actually sends still contains the removed attendee in the VEVENT.
- No iTip CANCEL is sent.
- After page reload, the attendee is still present.
- The Context Menü remains open, but updated to next attendee. (not the core of the issue)

### Calendar app version

6.4.0

### CalDAV-clients used

n/a - bug is in the Nextcloud Calendar web UI itself. Reproduces without any external client.

### Browser

Chromium 148

### Client operating system

Majaro Linux

### Server operating system

Ubuntu

### Web server

Apache

### Database engine version

MySQL

### PHP engine version

PHP 8.3

### Nextcloud version

33.0.3.2

Updated from an older version

### List of activated apps

```Text
calendar 6.4.0
dav 1.36.0
(other apps not relevant)
```

### Nextcloud configuration

```Text
(default - no calendar-specific overrides)
```

### Web server error log

```Text
(none - request returns 204 No Content / success)
```

### Log file

```Text
(no nextcloud.log entries - the server cleanly stores whatever the browser sends; the bug is entirely client-side)
```

### Browser log

```Text
EditorMixin.js:474 [Editor] Loading event data... {objectId: '...', recurrenceId: '...'}
calendarObjects.js:108 PUT https://.../remote.php/dav/calendars/<user>/<calendar>/<UID>.ics  → 204 OK

Call stack of the save:
  saveThisOnly → save → saveAndLeave → saveEvent
    → prepareAccessForAttachments → saveCalendarObjectInstance
      → updateCalendarObject (store/calendarObjects.js:108) → dav.update() → PUT
```

### Additional info

**Root cause**

This is a Vue 3 / Pinia reactivity bug introduced by the Vue 3 migration (PR #7697, merged Feb 2026).

`src/models/attendee.js:44-55` (`mapAttendeePropertyToAttendeeObject`) stores the calendar-js `AttendeeProperty` directly as `attendeeProperty` in the Pinia state object. Pinia/Vue 3 then deep-wraps that into a **reactive Proxy**.

`src/store/calendarObjectInstance.js:512` calls:

```js
calendarObjectInstance.eventComponent.removeAttendee(attendee.attendeeProperty)
```

- passing the **Proxy** to calendar-js.

`@nextcloud/calendar-js` `abstractComponent.js` `deleteProperty()` (synthesized via `advertiseMultipleOccurrenceProperty`) does:

```js
const arr = this._properties.get(property.name)
const index = arr.indexOf(property)
if (index === -1) return false   // ← silent fail here
```

`Array.indexOf` does **strict object identity**. The Proxy is not the same JS object as the raw `AttendeeProperty` stored in `_properties` → `indexOf` returns `-1` → `deleteProperty` returns `false` and does **nothing**. The caller never checks the return value and proceeds with the Vue-state splice. The UI updates, the user sees "success", and the next `toICS()` serializes the unmodified VEVENT.

## Live reproduction & verification at a DevTools breakpoint

In Chromium DevTools, set a breakpoint at `src/store/calendarObjectInstance.js:512`. Trigger a remove. In the console:

```js
// Confirm we're holding a Proxy:
[!!attendee.attendeeProperty.__v_isReactive, !!attendee.attendeeProperty.__v_raw]
// → [true, true]

// Confirm the underlying raw object IS in the array, just not under that proxy identity:
attendee.attendeeProperty.__v_raw === calendarObjectInstance.eventComponent._properties.get('ATTENDEE')[0]
// → true

// Confirm strict-equality indexOf misses it:
calendarObjectInstance.eventComponent._properties.get('ATTENDEE').indexOf(attendee.attendeeProperty)
// → -1   ← THE BUG

// Workaround at the call site fixes it:
calendarObjectInstance.eventComponent.removeAttendee(attendee.attendeeProperty.__v_raw)
calendarObjectInstance.eventComponent._properties.get('ATTENDEE').length
// → 1  (mutation actually applied)
```

After F8 / Save, the PUT then contains the correctly-mutated VEVENT and the server emits the expected iTip CANCEL.

**Why this slipped through**

The Vue 3 migration applied `markRaw` to several calendar-js objects, but inconsistently:

- `calendarComponent` (VCALENDAR root): `markRaw`'d at `src/models/calendarObject.js:76, 103` and `src/store/calendarObjects.js:244, 311, 326, 367` ✅
- `eventComponent`: `markRaw`'d at `src/store/calendarObjectInstance.js:73, 95` - but **after** `mapEventComponentToEventObject` has already populated `attendees[*].attendeeProperty`, so attendee-property proxies are already locked in.
- `parser` (import flow): `markRaw`'d at `src/store/importFiles.js:35` ✅ (PR #7997 - **same root cause, same fix pattern**)
- **`attendeeProperty`** in `src/models/attendee.js:46`: ❌ **not** marked raw
- **`organizerProperty`** in `src/models/event.js:129`: ❌ **not** marked raw (very likely the same bug for organizer mutations)

**Suggested fix direction**

Apply `markRaw` consistently to all calendar-js property/component objects at the point they enter the Pinia state:

- `mapAttendeePropertyToAttendeeObject` in `src/models/attendee.js`
- `mapEventComponentToEventObject` in `src/models/event.js` - including the inline `organizer.attendeeProperty`
- `mapAlarmComponentToAlarmObject`, `mapAttachmentPropertyToAttachmentObject` etc.

This mirrors the fix in PR #7997 and is the same direction PR #8341 has been taking.

**Demo**

https://github.com/user-attachments/assets/2b95a60f-5e86-4079-8e82-9d216ecf3737

ndo84bw:
### Summary

- Fixes #8364
- Wrap calendar-js `AttendeeProperty`, `EventComponent` (including organizer), `AlarmComponent`, `AttachmentProperty` and `RecurValue` with `markRaw` when they enter Pinia state, so Vue 3 does not deep-wrap them into reactive Proxies.

### Root cause

Vue 3 / Pinia deep-wrap any object stored in state into a reactive Proxy. `@nextcloud/calendar-js` mutations route through methods like `EventComponent.removeAttendee(prop)`, which in turn call `AbstractComponent.deleteProperty(prop)`:

```js
const arr = this._properties.get(property.name)
const index = arr.indexOf(property)
if (index === -1) return false
```

`Array.indexOf` uses strict object identity. The Proxy passed in from the store is not strictly equal to the raw property kept in `_properties` → `indexOf` returns `-1` → `deleteProperty` returns `false` and does nothing. The caller never checks the return value and proceeds with its local state splice, so the UI appears to update but the next `toICS()` serializes the unchanged VEVENT.

Verified live with a DevTools breakpoint at `calendarObjectInstance.js:512`:

```js
[!!attendee.attendeeProperty.__v_isReactive, !!attendee.attendeeProperty.__v_raw]
// → [true, true]   (Proxy)
attendee.attendeeProperty.__v_raw === eventComponent._properties.get('ATTENDEE')[0]
// → true           (correct underlying object)
eventComponent._properties.get('ATTENDEE').indexOf(attendee.attendeeProperty)
// → -1             (proxy ≠ raw, silent fail)
```

### Why this pattern

`calendarComponent` (VCALENDAR root) is already wrapped with `markRaw` in `src/models/calendarObject.js:76,103` and `src/store/calendarObjects.js:244,311,326,367`. `eventComponent` got the same treatment in `src/store/calendarObjectInstance.js:73,95`, but **after** `mapEventComponentToEventObject` had already populated `attendees[*].attendeeProperty` — too late, the proxies were already locked in.

This change applies the same hardening at every mapping site, so every calendar-js property/component enters Pinia state already raw. Same pattern as #7997, which applied it to the import parser.

### What this fixes

Confirmed live on Calendar 6.4.0 / Nextcloud 33.0.3.2:

- Removing an attendee from an event — PUT carries the modified VEVENT, server stores correctly, iTip CANCEL is sent.
- Changing an attendee's role (Chair / Required / Optional / Non-participant).

Both were previously silent no-ops at the `_properties.indexOf` step.

By the same mechanism this should also harden other calendar-js mutations that go through the same identity check on `_properties` — attachment removal, recurrence-rule property mutations, and other attendee setters that route into `deleteProperty` / `indexOf`. The same `markRaw` treatment is applied to all those calendar-js objects at their respective mapping sites.

### Out of scope (separate follow-ups)

After deploying the fix locally, an unrelated UI-side symptom remained — different root cause:

- **Stale editor state after save+reopen** — for up to ~60 s after a successful save, reopening the editor occasionally shows the pre-save state until a background calendar sync settles. Server-side state is correct; this looks like a client-side store-refresh race (`appendOrUpdateCalendarObjectsMutation` in `calendarObjects.js:308` overwrites the in-memory `calendarObject` after a FullCalendar refetch).

This will be filed as a separate issue.

### Tests

- All 392 unit tests pass locally.
- Lint clean on changed files (one pre-existing import-order issue in `recurrenceRule.js` fixed in the same commit).
- Manually tested on a Nextcloud 33.0.3.2 instance with Calendar 6.4.0 (bundle built from this branch's stable6.4 counterpart, deployed to `/var/www/html/apps/calendar/js`).

### Files changed

- `src/models/attendee.js` — `mapAttendeePropertyToAttendeeObject`
- `src/models/event.js` — `mapEventComponentToEventObject` (eventComponent + organizer's attendeeProperty)
- `src/models/alarm.js` — `mapAlarmComponentToAlarmObject` (both return branches)
- `src/models/attachment.js` — `mapAttachmentPropertyToAttchmentObject`
- `src/models/recurrenceRule.js` — `getDefaultRecurrenceRuleObjectForRecurrenceValue`

### DIFF
--- a/src/models/alarm.js
+++ b/src/models/alarm.js
@@ -2,6 +2,7 @@
  * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
+import { markRaw } from 'vue'
 import {
 	getAmountAndUnitForTimedEvents,
 	getAmountHoursMinutesAndUnitForAllDayEvents,
@@ -75,7 +76,7 @@ function mapAlarmComponentToAlarmObject(alarmComponent) {
 		const relativeTrigger = alarmComponent.trigger.value.totalSeconds
 
 		return getDefaultAlarmObject({
-			alarmComponent,
+			alarmComponent: markRaw(alarmComponent),
 			type: alarmComponent.action,
 			isRelative: alarmComponent.trigger.isRelative(),
 			relativeIsBefore,
@@ -92,7 +93,7 @@ function mapAlarmComponentToAlarmObject(alarmComponent) {
 		const absoluteDate = getDateFromDateTimeValue(alarmComponent.trigger.value)
 
 		return getDefaultAlarmObject({
-			alarmComponent,
+			alarmComponent: markRaw(alarmComponent),
 			type: alarmComponent.action,
 			isRelative: alarmComponent.trigger.isRelative(),
 			absoluteDate,
--- a/src/models/attachment.js
+++ b/src/models/attachment.js
@@ -2,6 +2,7 @@
  * SPDX-FileCopyrightText: 2022 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
+import { markRaw } from 'vue'
 
 /**
  * Creates a complete attachment object based on given props
@@ -36,7 +37,7 @@ function getDefaultAttachmentObject(props = {}) {
  */
 function mapAttachmentPropertyToAttchmentObject(attachmentProperty) {
 	return getDefaultAttachmentObject({
-		attachmentProperty,
+		attachmentProperty: markRaw(attachmentProperty),
 		fileName: attachmentProperty.getParameterFirstValue('FILENAME'),
 		formatType: attachmentProperty.formatType,
 		uri: attachmentProperty.uri,
--- a/src/models/attendee.js
+++ b/src/models/attendee.js
@@ -4,6 +4,7 @@
  */
 
 import { AttendeeProperty } from '@nextcloud/calendar-js'
+import { markRaw } from 'vue'
 
 /**
  * Creates a complete attendee object based on given props
@@ -43,7 +44,7 @@ function getDefaultAttendeeObject(props = {}) {
  */
 function mapAttendeePropertyToAttendeeObject(attendeeProperty) {
 	return getDefaultAttendeeObject({
-		attendeeProperty,
+		attendeeProperty: markRaw(attendeeProperty),
 		commonName: attendeeProperty.commonName,
 		calendarUserType: attendeeProperty.userType,
 		participationStatus: attendeeProperty.participationStatus,
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -4,6 +4,7 @@
  */
 
 import { DateTimeValue, DurationValue } from '@nextcloud/calendar-js'
+import { markRaw } from 'vue'
 import { getClosestCSS3ColorNameForHex, getHexForColorName } from '../utils/color.js'
 import { getDateFromDateTimeValue } from '../utils/date.js'
 import { mapAlarmComponentToAlarmObject } from './alarm.js'
@@ -84,7 +85,7 @@ function getDefaultEventObject(props = {}) {
  */
 function mapEventComponentToEventObject(eventComponent) {
 	const eventObject = getDefaultEventObject({
-		eventComponent,
+		eventComponent: markRaw(eventComponent),
 		title: eventComponent.title,
 		isAllDay: eventComponent.isAllDay(),
 		canModifyAllDay: eventComponent.canModifyAllDay(),
--- a/src/models/recurrenceRule.js
+++ b/src/models/recurrenceRule.js
@@ -1,8 +1,9 @@
-import { getDateFromDateTimeValue } from '../utils/date.js'
 /**
  * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
+import { markRaw } from 'vue'
+import { getDateFromDateTimeValue } from '../utils/date.js'
 import { getWeekDayFromDate } from '../utils/recurrence.js'
 
 /**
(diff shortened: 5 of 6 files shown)


SebastianKrupinski (approved):
Hi!

Thank you for the PR, you where on the right track.

I made two small changes
- removed the markraw on the organizer
- added markraw when adding a new attendee

PR #8378, opened 2026-05-21
fix(store): preserve calendar-object identity on unchanged refetch

### LINKED ISSUE #8367: Editor shows stale state after save+reopen until background sync settles
### Steps to reproduce

1. Open an event that has at least one attendee.
2. Make a change that goes through the Pinia store (e.g. change an attendee's role, toggle RSVP, add a reminder).
3. Click **Update** / **Save** - the editor closes.
4. Within ~60 seconds, click the event again to re-open it in the editor.

### Expected behavior

The editor shows the just-saved state.

### Actual behavior

The editor sometimes (not deterministically) shows the **pre-save state** - the change is not visible. Without a browser reload, the previous state remains. After ~60 s and/or a full page reload, the state finally reflects the server.

This affects **every editor field** that is rebuilt from `calendarObject.calendarComponent` when the editor reopens. Observed concretely for:

- **Attendees** - just-removed attendees may briefly reappear, just-added attendees may not show up yet
- **Location** - newly entered location not visible after editor reopen, only after a delay
- **Reminders** - added reminders not visible in the editor list (note: in events with attendees there is an additional separate bug that strips reminders server-side; the stale-state effect here is observable on attendee-less single-shot events where the reminder is correctly persisted in the server ICS)
- Likely every other field too (title, description, dates, etc. - same code path)

Server-side state is correct in all cases; this has been verified by capturing the PUT request body (server-side instrumentation of `CalDavBackend::updateCalendarObject`) and by checking the resulting event via an external CalDAV client (Thunderbird). The desync is entirely in the frontend Pinia store, and it self-heals after the next background sync (~60 s without `notify_push`).

### Calendar app version

6.4.1

### CalDAV-clients used

n/a - purely web-UI behaviour

### Browser

Chromium 148

### Client operating system

Manjaro Linux

### Server operating system

Ubuntu

### Web server

Apache

### Database engine version

MySQL

### PHP engine version

PHP 8.3

### Nextcloud version

33.0.3.2

Updated from an older version

### List of activated apps

```Text
calendar 6.4.1
dav 1.36.0
```

### Nextcloud configuration

```Text
notify_push: not installed → periodic 60 s background sync used (Calendar.vue:295)
```

### Web server error log

```Text
(none - request returns 204 No Content)
```

### Log file

```Text
(none - server-side flow is clean)
```

### Browser log

```Text
EditorMixin.js:474 [Editor] Loading event data... {objectId: '...', recurrenceId: '...'}
EditorMixin.js:482 [Editor] Event loaded successfully  ← but shows pre-save state
```

### Additional info

**Suspected root cause**

When the editor closes, `closeEditor` (`EditorMixin.js:546`) calls `resetCalendarObjectInstanceObjectIdAndRecurrenceId()` on the calendar-object-instance store, which nulls the cached `calendarObject` and `calendarObjectInstance` references. The next time the editor opens, `getCalendarObjectInstanceByObjectIdAndRecurrenceId` cache-misses and pulls the `calendarObject` from `calendarObjectsStore` via `getEventByObjectId` (`calendars.js:782`).

Meanwhile, after the PUT in `updateCalendarObject` (`calendarObjects.js:102-117`) we increment `modificationCount`. That triggers the FullCalendar `modificationCount` watcher in `CalendarGrid.vue:224` → `calendarApi.refetchEvents()`. If the previously-cached `fetchedTimeRange` no longer covers the relevant range (e.g. cleared by an interleaved `syncCalendar` triggered by the 60 s polling in `Calendar.vue:295`), the event source re-fetches via `calendarsStore.getEventsFromCalendarInTimeRange` (`calendars.js:724-773`), which ends in `calendarObjectsStore.appendOrUpdateCalendarObjectsMutation` (`calendarObjects.js:308-315`):

```js
appendOrUpdateCalendarObjectsMutation({ calendarObjects = [] }) {
    for (const calendarObject of calendarObjects) {
        if (calendarObject.calendarComponent) {
            calendarObject.calendarComponent = markRaw(calendarObject.calendarComponent)
        }
        this.calendarObjects[calendarObject.id] = calendarObject   // ← replaces existing in-memory state
    }
}
```

This unconditionally replaces the existing `calendarObject` in the store with a freshly-parsed one from the most recent server response. If the iTip plugin is still processing a follow-up write at that moment (e.g. an attendee's `sabredav-*.ics` schedule-inbox write hasn't completed), the refresh can race and show a transient state. Even if not transient, the replaced `calendarObject` is a different object with different `calendarComponent` identity than what the editor still holds via `calendarObjectInstanceStore.calendarObject`.

Possible fix directions (not verified):
- Compare ETag / SEQUENCE / `LAST-MODIFIED` before replacing the in-memory `calendarObject` and prefer the local copy if newer.
- Don't re-fetch the just-PUT'd event for `N` seconds after `updateCalendarObject` returns.
- Only update *fields* of an existing `calendarObject` in place, instead of swapping the whole object reference.

**Related**

- #8364 / #8365 (this issue surfaced *during* the testing of the `markRaw` identity-mismatch fix - the silent failures had previously masked it)
- The 60 s "self-heals after waiting" behaviour suggests the next background sync settles the state once iTip processing finishes server-side.

ndo84bw:
### Summary

- Fixes #8367
- `appendOrUpdateCalendarObjectsMutation` no longer replaces an existing `calendarObjects[id]` with a fresh instance when the ETag is unchanged. The existing instance is kept, so references to it (notably the open editor's `calendarObject`) stay attached.

### Why

The editor shows stale state after save+reopen because of an object-identity detachment:

1. The editor opens an event; `this.calendarObject` points to the store instance `calendarObjects[id]`.
2. While the editor is open, a background refetch (`getEventsFromCalendarInTimeRange`) runs and `appendOrUpdateCalendarObjectsMutation` does `calendarObjects[id] = newInstance` - even when the refetched object is byte-identical (same ETag). The store entry is now a different instance; the editor still holds the old one.
3. On save, `updateCalendarObject` mutates and PUTs the editor's (now detached) object. The server is updated correctly, but the store entry is untouched.
4. On reopen, `getEventByObjectId` returns the store entry (the un-mutated instance) -> the editor shows the pre-save state.
5. The next background sync re-parses the server and replaces the store entry, so it self-heals after ~60s (longer without `notify_push`). The intermittency depends on whether a refetch happens to land during the open editor session.

The fix skips the replacement when the ETag is unchanged (same ETag means byte-identical content, so nothing is lost), preserving instance identity. Genuinely changed objects (different ETag) are still replaced.

### Alternatives considered

- **Update the existing instance in place** (copy `dav` / `calendarComponent` / ... onto the stored object) instead of skipping. This also preserves identity, but it would overwrite the editor's in-progress, unsaved working copy when that copy shares the `calendarComponent`, and it needs field-by-field copying that is easy to get wrong. The ETag-skip avoids touching anything precisely when there is nothing new to apply.
- **Verify freshness in `getEventByObjectId`** (its existing TODO: send a HEAD and compare ETags on editor open). That would address the reopen read but not the underlying instance churn, and it adds a request on every open.
- **Out of scope:** a genuine multi-client conflict, i.e. a refetch with a *different* ETag landing while the editor is open. That legitimately replaces the instance and is a separate conflict-resolution concern, not this stale-state bug.

We chose the ETag-skip as the smallest change that fixes the observed cause without risking the editor's unsaved edits.

### Tested

Root cause confirmed on the live instance (NC 33.0.3.2 / Calendar 6.4.1) via temporary instrumentation: logging every overwrite, save and reopen-read showed that stale reopens coincide exactly with the editor's `calendarObject` no longer being identical to the store entry, and that the detaching overwrite always carried the same ETag as the store entry.

Fix verified live on the same instance: changing the location 20 times in a row (edit -> save -> reopen each time) showed the just-saved value every time, with no stale reopen. Reproduced clean in both the full editor and the popup editor (both go through the same `EditorMixin` -> store path).

Unit tests (added in this PR, `npm run test:unit`):
- An unchanged refetch (same ETag) keeps the existing instance (identity preserved).
- A changed refetch (different ETag) replaces the instance.
- A new object (not in the store) is appended.
- Confirmed the unchanged-ETag test fails against the current code.

### DIFF
--- a/src/store/calendarObjects.js
+++ b/src/store/calendarObjects.js
@@ -307,6 +307,13 @@ export default defineStore('calendarObjects', {
 		 */
 		appendOrUpdateCalendarObjectsMutation({ calendarObjects = [] }) {
 			for (const calendarObject of calendarObjects) {
+				const existing = this.calendarObjects[calendarObject.id]
+				// Keep the existing instance when the object is unchanged (same ETag).
+				// Replacing it with a fresh instance would detach references held elsewhere
+				if (existing && existing.dav?.etag && calendarObject.dav?.etag
+					&& existing.dav.etag === calendarObject.dav.etag) {
+					continue
+				}
 				if (calendarObject.calendarComponent) {
 					calendarObject.calendarComponent = markRaw(calendarObject.calendarComponent)
 				}
--- a/tests/javascript/unit/store/calendarObjects.test.js
+++ b/tests/javascript/unit/store/calendarObjects.test.js
@@ -2,11 +2,59 @@
  * SPDX-FileCopyrightText: 2019 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
+import { setActivePinia, createPinia } from 'pinia'
+import { toRaw } from 'vue'
+import useCalendarObjectsStore from '../../../../src/store/calendarObjects.js'
 
 describe('store/calendarObjects test suite', () => {
 
-	it('should be true', () => {
-		expect(true).toEqual(true)
+	beforeEach(() => {
+		setActivePinia(createPinia())
+	})
+
+	describe('appendOrUpdateCalendarObjectsMutation', () => {
+		const makeObject = (id, etag, tag) => ({
+			id,
+			dav: { etag },
+			calendarComponent: {},
+			_tag: tag,
+		})
+
+		it('appends a calendar-object that is not in the store yet', () => {
+			const store = useCalendarObjectsStore()
+			const obj = makeObject('a', '"v1"', 'first')
+
+			store.appendOrUpdateCalendarObjectsMutation({ calendarObjects: [obj] })
+
+			expect(store.calendarObjects.a._tag).toBe('first')
+			expect(toRaw(store.calendarObjects.a)).toBe(obj)
+		})
+
+		it('keeps the existing instance when the ETag is unchanged (no detach)', () => {
+			const store = useCalendarObjectsStore()
+			const original = makeObject('a', '"v1"', 'original')
+			store.appendOrUpdateCalendarObjectsMutation({ calendarObjects: [original] })
+
+			const incoming = makeObject('a', '"v1"', 'incoming')
+			store.appendOrUpdateCalendarObjectsMutation({ calendarObjects: [incoming] })
+
+			// Identity must be preserved: the original instance is kept so references
+			// held elsewhere (e.g. the open editor) stay attached. See #8367.
+			expect(store.calendarObjects.a._tag).toBe('original')
+			expect(toRaw(store.calendarObjects.a)).toBe(original)
+		})
+
+		it('replaces the instance when the ETag changed', () => {
+			const store = useCalendarObjectsStore()
+			const original = makeObject('a', '"v1"', 'original')
+			store.appendOrUpdateCalendarObjectsMutation({ calendarObjects: [original] })
+
+			const incoming = makeObject('a', '"v2"', 'updated')
+			store.appendOrUpdateCalendarObjectsMutation({ calendarObjects: [incoming] })
+
+			expect(store.calendarObjects.a._tag).toBe('updated')
+			expect(toRaw(store.calendarObjects.a)).toBe(incoming)
+		})
 	})
 
 })


SebastianKrupinski (changes requested):
Hi @ndo84bw 

I am going to block this change, I can not reproduce the issue that you stated in

https://github.com/nextcloud/calendar/issues/8367

Updating information on a event is properly displayed after re-opening the event

ndo84bw:
Hi @SebastianKrupinski 

Since it is a timing issue and there were not any step-by-step instructions, it was hard for me to reproduce at first. I have added another video to issue #8367. The trick is to keep the editor open until the background synchronization is complete and only then save your changes.

SebastianKrupinski (approved):
I was finally able to reproduce the issue.

Tested the fix, it work as expected.

SebastianKrupinski:
Rebase to head and trimmed the comment a bit.

PR #8386, opened 2026-05-22
fix(editor): store categories again by disabling deselectFromDropdown

### LINKED ISSUE #8337: Calendar 6.4.0 doesn't store categories
### Steps to reproduce

1. Create new event
2. select category or add new
3. save

### Expected behavior

Category is stored.

### Actual behavior

This is the first use of a (shared) calendar.

WebUI: After selecting the category the category field remains empty and also does so after saving.
Thunderbird saves categories in local calendars but not in the Nextcloud calendar.

Additionally: After syncing the calendar in Thunderbird in the WebUI the checkboxes are removed that give permission to the groups shared.

### Calendar app version

6.4.0

### CalDAV-clients used

Thunderbird

### Browser

Firefox 150.0.3, Brave 1.90.122

### Client operating system

Windows 11 25H2

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

32.0.9

None

### List of activated apps

```Text
see screenshots below

```
Image
Image
Image
Image
Image

### Nextcloud configuration

```Text
We don't host Nextcloud ourselves but use Hetzner Storageshare.
```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

ndo84bw:
### Summary

- Fixes #8337
- Disable `deselectFromDropdown` on the categories `NcSelect` in `PropertySelectMultiple`, so selecting a category is no longer instantly undone. The point is that categories can be added and stored again at all - which is currently impossible.

### Why

With the current vue-select, `deselectFromDropdown` fires an `option:deselected` immediately after `option:selecting` for a freshly selected value in a multi-select (the condition is `deselectFromDropdown && (clearable || multiple && selectedValue.length > 1)`). `PropertySelectMultiple` adds the category on `option:selecting` and then removes it again on the spurious `option:deselected`, so categories cannot be added or stored.

Removing an existing category via the chip's close button uses a separate path and is unaffected. Disabling `deselectFromDropdown` removes the spurious deselect; categories are still removed via the chip.

Trade-off: with `deselectFromDropdown` off, you can no longer deselect an entry by clicking it again in the dropdown - but removal via the chip's close button still works, and in return categories can be added and stored at all again.

For context: `deselectFromDropdown="true"` was added in #5648 ("migrate to nextcloud/vue 8"), in the same commit that added the `// budget deselectFromDropdown since the vue-select implementation doesn't work` workaround in `tag()` - so that behavior was already known to be broken back then. This change just disables it; happy to keep an in-dropdown deselect instead if you prefer (see Follow-up).

### Tested

Manually, on NC 33.0.3 and Calendar 6.4.1:
- Before: selecting a category leaves the field empty and nothing is stored. (Instrumenting the component showed `option:selecting` immediately followed by `option:deselected` for the same value.)
- After: selecting a category keeps it and stores it on save; adding several works; removing a single category via its chip removes only that one.

No unit test is added: the bug is an interaction with the third-party vue-select component (a prop-value behavior), not in our own handler logic, so there is no meaningful behavioral unit test for it. Verified manually.

### Follow-up

With `deselectFromDropdown` off, an already-selected entry can no longer be deselected by clicking it again in the dropdown (removal is via the chip). The cleaner long-term UX is to hide already-selected entries from the dropdown entirely (standard multi-select behavior), which removes the need for `deselectFromDropdown` altogether; a toggle with a selected-state affordance would be the alternative. I left this out of the bug fix on purpose, but can follow up with it - just let me know which direction you prefer.

### DIFF
--- a/src/components/Editor/Properties/PropertySelectMultiple.vue
+++ b/src/components/Editor/Properties/PropertySelectMultiple.vue
@@ -27,7 +27,7 @@
 				:multiple="true"
 				:taggable="true"
 				:noWrap="false"
-				:deselectFromDropdown="true"
+				:deselectFromDropdown="false"
 				:createOption="(label) => ({ value: label, label })"
 				inputId="label"
 				label="label"


SebastianKrupinski (approved):
Tested. Works!

Thank you!

PR #8392, opened 2026-05-23
fix(trashbin): restrict restore and delete on the frontend

hamza221:
In case of delegation and sharing (read-only), user might not have enough privilege to restore or permanently delete objects, 
currently the frontend fails, this PR hides the buttons instead .
- needs https://github.com/nextcloud/cdav-library/pull/1043 
- for showing deleted events in the trash bin https://github.com/nextcloud/server/pull/60648

### DIFF
--- a/src/components/AppNavigation/CalendarList/Trashbin.vue
+++ b/src/components/AppNavigation/CalendarList/Trashbin.vue
@@ -70,11 +70,11 @@
 									</td>
 									<td>
 										<div class="item-actions">
-											<NcButton variant="secondary" @click="restore(item)">
+											<NcButton v-if="item.canRestore" variant="secondary" @click="restore(item)">
 												{{ t('calendar', 'Restore') }}
 											</NcButton>
 
-											<NcActions :forceMenu="true">
+											<NcActions v-if="item.canDelete" :forceMenu="item.canRestore">
 												<NcActionButton @click="onDeletePermanently(item)">
 													<template #icon>
 														<IconDelete :size="20" decorative />
@@ -164,6 +164,7 @@ export default {
 				url: calendar._url,
 				deletedAt: calendar._props['{http://nextcloud.com/ns}deleted-at'],
 				color: calendar.color ?? uidToHexColor(calendar.displayname),
+				readonly: false,
 			}))
 			const formattedCalendarObjects = this.objects.map((vobject) => {
 				let eventSummary = t('calendar', 'Untitled item')
@@ -185,6 +186,8 @@ export default {
 				const color = vobject.calendarComponent.getComponentIterator().next().value?.color
 					?? vobject.calendar?.color
 					?? uidToHexColor(subline)
+				const canDelete = vobject.calendar?.canDeleteObject
+				const canRestore = vobject.calendar?.canCreateObject && vobject.calendar?.canModifyObject
 				return {
 					vobject,
 					type: 'object',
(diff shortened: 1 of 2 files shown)


src/store/calendars.js
```diff
@@ -126,17 +126,32 @@ export default defineStore('calendars', {
...
+				calendarBySource.set(`${lastSegment(calendar.owner)}|${sourceUriOf(calendar)}`, calendar)
 			})
 
-			return state.deletedCalendarObjects.map((obj) => ({
-				calendar: calendarUriMap[obj.dav._props['{http://nextcloud.com/ns}calendar-uri']],
-				...obj,
-			}))
+			return state.deletedCalendarObjects.map((obj) => {
+				const owner = obj.dav._props['{http://nextcloud.com/ns}calendar-owner-principal-uri']
+				const sourceUri = obj.dav._props['{http://nextcloud.com/ns}source-calendar-uri']
```
SebastianKrupinski: Please add access methods for these properties, instead of using the properties dav definition name

src/components/AppNavigation/CalendarList/Trashbin.vue
```diff
@@ -194,6 +197,8 @@ export default {
 					url: vobject.uri,
 					deletedAt: vobject.dav._props['{http://nextcloud.com/ns}deleted-at'],
```
SebastianKrupinski, line 198: I know this is not really part of the scope, but when you are adding the access methods can you also add them for any other missing ones

vobject.dav._props['{http://nextcloud.com/ns}deleted-at']

SebastianKrupinski (changes requested):

src/store/calendars.js
```diff
@@ -126,17 +126,32 @@ export default defineStore('calendars', {
...
+			}
+
+			// Key calendars by `${ownerUserId}|${sourceUri}` so deleted objects
+			// match their origin calendar across owned/shared/delegated views.
+			const calendarBySource = new Map()
 			state.calendars.forEach((calendar) => {
-				const withoutTrail = calendar.url.replace(/\/$/, '')
-				const uri = withoutTrail.slice(withoutTrail.lastIndexOf('/') + 1)
-				calendarUriMap[uri] = calendar
+				calendarBySource.set(`${lastSegment(calendar.owner)}|${sourceUriOf(calendar)}`, calendar)
```
hamza221: @SebastianKrupinski is it safe to assume that url/ owner always follows this ?

SebastianKrupinski: I don't know for sure but in a delegated scenario, the calendar owner is the original owner, in a shared scenario the owner is the original owner also. But in a shared scenario the uri is different than the original uri, not sure about delegated uri atm.

So I guess it depends on the intent, are you trying to match all the different calendar and object to the original owner or separate them.

GVodyanov (approved):
Tested and works

ChristophWurst (approved):

src/store/calendars.js
hamza221: where should they be defined ? as these are specific to trashbin Objects

SebastianKrupinski: The should be added to the cdav library objects. Then used in calendar

hamza221: My question is where in the cdav library?  as this is not a general prop for all VObjects

SebastianKrupinski: You can add them to:

https://github.com/nextcloud/cdav-library/blob/main/src/models/vobject.js

If you really want to do it properly there should technically be a vobject trash bin object, like the trashbin collection object.

https://github.com/nextcloud/cdav-library/blob/main/src/models/calendarTrashBin.js

SebastianKrupinski (approved):
Looks good

hamza221:
Bypassing codecov

PR #8398, opened 2026-05-25
fix(appointment): capture Attendee browser language and save it

hamza221:
Language parameter is checked here https://github.com/nextcloud/server/blob/master/apps/dav/lib/CalDAV/Schedule/IMipService.php#L928 but never set 

This is for non authenticated attendees who have their browser language set to something different than server default

### DIFF
--- a/lib/Service/Appointments/BookingCalendarWriter.php
+++ b/lib/Service/Appointments/BookingCalendarWriter.php
@@ -175,7 +175,9 @@ class BookingCalendarWriter {
 				'CUTYPE' => 'INDIVIDUAL',
 				'RSVP' => 'TRUE',
 				'ROLE' => 'REQ-PARTICIPANT',
-				'PARTSTAT' => 'ACCEPTED'
+				'PARTSTAT' => 'ACCEPTED',
+				// Read by IMipService::setL10nFromAttendee for iTip update/cancel emails.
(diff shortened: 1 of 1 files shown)


SebastianKrupinski:
Hey, did you test this? I think you need to capture it in the front end, the l10n like this will give you non consistent results

hamza221:
> Hey, did you test this? I think you need to capture it in the front end, the l10n like this will give you non consistent results

Ofc , it works. you can test it too

GVodyanov (dismissed):
I don't understand what to test for here, is it emails arriving in the correct language of the attendee?

EDIT I MISCLICKED APPROVE AND I DON'T KNOW HOW TO REMOVE IT

SebastianKrupinski:
> I don't understand what to test for here, is it emails arriving in the correct language of the attendee?

The created event ics with have a LANGUAGE parameter attached to it.

hamza221:
Repro: 
- be a non logged in user 
- have your browser set a different language eg Italian ( not server default eg English) 
- schedule an appointment with user1
- check confirmation message (will be in Italian even pre patch) 
- user1 cancels or modifies the event
- check the notification email (pre patch: English, with patch: Italian )

SebastianKrupinski (approved):

AndyScherzinger (approved):

GVodyanov (approved):

PR #8400, opened 2026-05-25
add agents.md

GVodyanov:


SebastianKrupinski (approved):

SebastianKrupinski:
Approved. But I am wondering if we should add something about Cdav and the server dav app, so that agents do now add server specific and cdav specific code to the calendar app

GVodyanov:
> Approved. But I am wondering if we should add something about Cdav and the server dav app, so that agents do now add server specific and cdav specific code to the calendar app

Good idea, will add now

ChristophWurst (approved):

SebastianKrupinski:
Just thought of another addition we should add to the instruction.

We should instruct the agent that when creating new components it should use vue composition api with setup and typescript, and any newly created stores, services, etc should be in typescript.

ChristophWurst:
@SebastianKrupinski good point!

we can get a first version in soon. the file will definitely need a few iterations and refinements over time. we won't and can't get it perfect right away. that is expected

GVodyanov:
> Just thought of another addition we should add to the instruction.
> 
> We should instruct the agent that when creating new components it should use vue composition api with setup and typescript, and any newly created stores, services, etc should be in typescript.

Added!

SebastianKrupinski:
Looks great!

SebastianKrupinski:
Bypassed the failing tests as the failures are not releated

PR #8433, opened 2026-06-03
fix(events): do not discard text input of dates and times in event edit dialog

### LINKED ISSUE #8307: Can no longer edit time
### Steps to reproduce

1. Edit event
2. Click on time
3. Try to enter time instead of picking from a list

### Expected behavior

The way it was until the 6.4.0. update. You could enter your own time instead of having to choose. Now it's impossible to enter a time that isn't rounded to 5 minute increments.

This violates the iCal standard, BTW. We are allowed to choose times that aren't rounded to 5 minutes.

### Actual behavior

Must choose time from a list instead of manually entering it like just before the latest update.

### Calendar app version

6.4.0

### CalDAV-clients used

_No response_

### Browser

Firefox 150.0.1

### Client operating system

Fedora 44

### Server operating system

Fedora 44

### Web server

Apache

### Database engine version

MySQL

### PHP engine version

Other

### Nextcloud version

Nextcloud Hub 26 Winter (33.0.3)

Updated from an older version

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

_No response_

odzhychko:
Provisional fix for #8307 until I can properly fix it in @nextcloud/vue .

The comments in the code describe how this fix works.
The underlying issue is that `NcDateTimePicker` specifies how to format dates/times but not how to parse them back.

#### Possible issues with this fix

-  `datetime-locale-patterns` might use older locale data than used by browsers for `Intl.DateTimeFormat`
   - this would result in dates/times formatted differently if the spec for a locale was recently changed
- `datetime-locale-patterns`  bundles more locale data then possibly needed for locales supported by Nextcloud

#### Before

[Screencast from 2026-06-03 16-17-41.webm](https://github.com/user-attachments/assets/f9f15340-1866-48e2-81ad-1ece1c0dc3a1)

#### After

[Screencast from 2026-06-03 16-16-14.webm](https://github.com/user-attachments/assets/f07638cb-b46a-4fdd-9673-76c347d1cfb6)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,6 +39,7 @@
         "color-string": "^2.1.4",
         "core-js": "^3.49.0",
         "css-color-names": "^1.0.1",
+        "datetime-locale-patterns": "^1.0.2",
         "debounce": "^3.0.0",
         "linkifyjs": "^4.3.3",
         "md5": "^2.3.0",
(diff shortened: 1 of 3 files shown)


SebastianKrupinski:
Hey,

I'm not against doing it this way, but..
 
I had a look at this and yes this is a way to temporarily fix this. The issue with the temporary part is that we never, end up getting time to the actually fix it properly and temporary becomes permanent.  So I am big supporter of not cutting corners and fixing the issue where it is being caused. (nextcloud/vue),

As for the extra parsing library, I don't know if this is needed, the nc/vue pr I did to fix this used Date() to parse and it worked correctly, 99% of the time. Can we just use the built in date() function instead of another dependency?

odzhychko:
@SebastianKrupinski Thanks for looking it over.

> The issue with the temporary part is that we never, end up getting time to the actually fix it properly and temporary becomes permanent.

Yeah, I know what you mean. If it helps, I do not intend to mark #8307 as finished until I fix it at the root.

> As for the extra parsing library, I don't know if this is needed, the nc/vue pr I did to fix this used Date() to parse and it worked correctly, 99% of the time.

I think using `Date()` will not work for all locales. It fails, for example, for `es`:

```js
new Date(Intl.DateTimeFormat('es', { dateStyle: 'medium' }).format(new Date('2026-12-30')))
```

But maybe I miss something.
I plan to test all locales automatically in the proper fix in nc/vue.
I can be more sure then.

> Can we just use the built in date() function instead of another dependency?

I also don't like the extra dependency. When make a fix in the `NcDateTimePicker` I will look for existing depencies that might solve this. @nextcloud/moment seems to be a good candidate for consistency.

SebastianKrupinski (approved):

PR #8441, opened 2026-06-05
fix(eventDidMount): increase past event opacity

### LINKED ISSUE #8380: display of booked slots since version 6.4.1

Thank you for the app – it’s really great! Since the last update (May 2026), booked slots are displayed with a single line on the left when using the high-contrast dark theme in Nextcloud 33.0.3.

 The old display, where the whole slot was coloured, was easier to read. Would it be possible to consider reverting to that display? Or to allow users to customise the display so they can do so if they wish?

_No response_

### Additional context

_No response_

### LINKED ISSUE #8418: Re-enable full colored events.

Colored events on the app used to look like this screenshot from https://apps.nextcloud.com/apps/calendar:

Image

After latest update of the app last night (`6.4.2`), they look like:

after

What’s going on? The point of coloring events is to ease navigation among them when there are so many. Now my calendar pages look all uniform. This left thin vertical bar is not enough to notice colored events without squinting.

I found no way to set it back.. is there a way to set it back? Colors are useful :cry:

Either revert the change or introduce a toggle switch to opt-out of the new `unreadable fainted colors`? :\

I am considering reverting to earlier version or looking for an alternate calendar app :\

### Additional context

~~I've looked through the repo without finding the rationale for this change.. maybe there is a good reason to do that that I've missed?~~
[UPDATE]: relevant PR and rationale kindly pointed out by @tcitworld : #8149

Again:

> I understand the intent and I value WCAG AA as a feature. I can even live with the new style being the new default. But [this](https://github.com/nextcloud/calendar/pull/8149#issuecomment-4444537137) is flipping the app for me. It has suddenly become a pain to parse.
> 
> Can we have an option to opt back into opacity?

GVodyanov:
Fix https://github.com/nextcloud/calendar/issues/8380
Fix https://github.com/nextcloud/calendar/issues/8418

| Before | After | 
| - | - |
| Screenshot From 2026-06-05 17-37-25 | Screenshot From 2026-06-05 17-34-12 |

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -135,7 +135,7 @@
 		border-block-start-width: 0 !important;
 		border-block-end-width: 0 !important;
 		border-inline-end-width: 0 !important;
-		border-inline-start-width: var(--default-grid-baseline) !important;
+		border-inline-start-width: calc(var(--default-grid-baseline) + var(--default-grid-basline) * 0.5) !important;
 		border-inline-start-style: solid !important;
 	}
 
@@ -184,7 +184,7 @@
 			border-block-start-width: 1px !important;
 			border-block-end-width: 1px !important;
 			border-inline-end-width: 1px !important;
-			border-inline-start-width: var(--default-grid-baseline) !important;
+			border-inline-start-width: calc(var(--default-grid-baseline) + var(--default-grid-basline) * 0.5) !important;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


GVodyanov:
Thanks for your suggestions! How does this look @kra-mo?

SebastianKrupinski (approved):

PR #8456, opened 2026-06-10
fix(store): propagate organizer/attendees to recurrence-exceptions on…

### LINKED ISSUE #8450: Adding an attendee to a recurring series with a moved occurrence - moved instance is invisible to the attendee, and shows no attendees to the organizer
### Steps to reproduce

1. As organizer, create a recurring event in the web UI: daily, 4 occurrences, no attendees. Save.
2. Open occurrence NR-2 in the editor, change its start time (e.g. push it by one hour), choose **Update this occurrence**.
3. Reopen occurrence NR-1 in the editor, **add one attendee** (another local Nextcloud user), choose **Update this and all future**.
4. As the newly added attendee, look at the series in the calendar.
5. As the organizer, reopen the moved occurrence (NR-2).

### Expected behavior

- The newly added attendee sees the series in the same state the organizer has, including the moved occurrence NR-2 at its new time.
- The organizer's moved occurrence NR-2 lists the newly added attendee - "Update this and all future" was applied to the whole series.

### Actual behavior

- **Symptom A - attendee side:** The new attendee sees the series at the original times. The moved occurrence NR-2 appears at its original time, as if step 2 had never happened.
- **Symptom B - organizer side:** When the organizer reopens the moved occurrence NR-2, the attendee list is empty for that occurrence. The series master (occurrence NR-1, NR-3, NR-4) shows the attendee correctly.

### Calendar app version

6.4.2

### CalDAV-clients used

-

### Browser

Chromium 148 and Firefox 151

### Client operating system

Manjaro Linux

### Server operating system

Ubuntu Linux

### Web server

Apache

### Database engine version

MySQL

### PHP engine version

PHP 8.3

### Nextcloud version

33.0.3.2 - also reproduced on **33.0.5** (current `latest` Docker image, "Nextcloud Hub 26 Winter").

Fresh install

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

Image

ndo84bw:
## Summary

Fixes nextcloud/calendar#8450.

When the organizer applies **"Update this and all future"** to the master occurrence (NR-1) of a recurring series that already contains a moved occurrence (RECURRENCE-ID override), calendar-js' `_overridePrimaryItem()` copies the new state onto the master but leaves the existing override VEVENT untouched. Consequence: if the user just added the first attendee (and thus the ORGANIZER), the override ends up without either property. Two visible symptoms:

- **Attendee side:** Sabre's TipBroker omits the override from the iTip REQUEST (no ATTENDEE = no recipient on that VEVENT), so the attendee's client expands the RRULE normally and shows the moved occurrence at its original time.
- **Organizer side:** the Nextcloud WebUI crashes silently inside `OrganizerListItem.vue` (`organizer.attendeeProperty.getParameterFirstValue(...)` on `undefined`), the attendee list does not render, and the override appears empty even though Thunderbird and other CalDAV clients display it correctly.

## Fix

+30 LOC in `src/store/calendarObjectInstance.js`: one new file-local helper `syncMasterParticipantsOntoExceptions(master, exceptions)` plus a four-line guarded call inside `saveCalendarObjectInstance`.

Immediately after `eventComponent.createRecurrenceException(thisAndAllFuture)` returns, if the call entered the master-in-place override path (`thisAndAllFuture && isExactForkOfPrimary && primaryItem.isMasterItem()`), iterate the recurrence-exceptions whose `RECURRENCE-ID` is at or after the master start and overwrite each exception's `ATTENDEE` list and `ORGANIZER` with clones of the master's. The `RECURRENCE-ID >= master start` filter is defensive: for any RFC-compliant series every exception already satisfies this, so in practice the filter selects all exceptions, but it guards against pathological data where an exception's RID precedes the current master start. `ORGANIZER` on the exception is only deleted when a replacement exists, to avoid introducing the same crash this PR fixes.

The fix lives in the calendar app store rather than in `calendar-js`'s `_overridePrimaryItem()` for delivery speed; the same propagation arguably belongs one layer down so every consumer of `createRecurrenceException(true)` on a master-aligned fork benefits. A follow-up against `nextcloud/calendar-js` is a reasonable next step.

## Trade-offs

- **Scope limited to `ATTENDEE` and `ORGANIZER`.** The same shape of bug almost certainly applies to any other master-level property edit (`SUMMARY`, `DESCRIPTION`, `LOCATION`, ...) when "this and all future" is applied to the master in place: the override keeps its stale value while the master gets the new one. Out of scope for this PR, which targets the reported #8450 attendee-visibility regression specifically. Worth a follow-up if anyone reproduces another property surfacing the same way.
- **Wholesale replacement of override attendees, not a delta.** The Nextcloud WebUI does not currently expose per-occurrence attendee editing, but other CalDAV clients (Thunderbird) do. Any per-occurrence attendee customisation a user has set on an override is overwritten by this propagation. Acceptable for the reported scenario; a future enhancement could diff against the pre-update master to preserve overrides whose attendees diverge intentionally.
- **Per-occurrence `PARTSTAT`, `RSVP`, `SCHEDULE-STATUS`** on an override-side attendee are similarly overwritten in the same gesture. This is the intended semantics for "Update this and all future" applied to attendees.

## Test plan

Manual repro on a live Nextcloud 33 / Calendar 6.4.2 instance, two real accounts (organizer + attendee), both Thunderbird CalDAV and the NC WebUI verified:

- [x] Create a daily/4× series with no attendees.
- [x] Move NR-2 (+1h), "Update this occurrence".
- [x] Open NR-1, add an attendee, "Update this and all future".
- [x] Reopen NR-2 as organizer in popover and full editor: ORGANIZER and ATTENDEE both rendered.
- [x] Verify the override VEVENT in the iTip REQUEST received by the attendee carries the same `ATTENDEE` and `ORGANIZER` as the master (Thunderbird as ground truth: the moved occurrence shows up with organizer and attendee on it).
- [x] Attendee's CalDAV-synced calendar (Thunderbird and NC WebUI) shows the moved NR-2 at the new time with the attendee listed.
- [ ] Unit test in `tests/javascript/unit/store/calendarObjectInstance.test.js` — file does not exist yet; happy to add one in a follow-up if maintainer prefers that here.

## Related / follow-ups

- A separate iTip REPLY-side issue surfaces with the same series afterwards: when the attendee clicks "Accept" in Thunderbird, only the master VEVENT's PARTSTAT is updated. The override stays NEEDS-ACTION even though the REQUEST and the organizer's stored copy both correctly list the attendee on it. Filed as nextcloud/server#61114. Likely a server-side TipBroker concern, adjacent to nextcloud/server#60452 / nextcloud/server#60536, not in scope here.

## 🤖 AI (if applicable)

### DIFF
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -34,6 +34,35 @@ import useCalendarsStore from './calendars.js'
 import useSettingsStore from './settings.js'
 import { updateRoomParticipantsFromEvent } from '@/services/talkService'
 
+/**
+ * Overwrites the ATTENDEE/ORGANIZER of the master's current-and-future exceptions
+ * with clones of the master's own, after an in-place "this and all future" master update.
+ *
+ * @param {object} eventComponent Forked event component being saved.
+ */
+export function syncMasterParticipantsOntoFutureExceptions(eventComponent) {
+	if (!eventComponent.isExactForkOfPrimary || !eventComponent.primaryItem.isMasterItem()) {
+		return
+	}
+	const master = eventComponent.primaryItem
+	const ref = eventComponent.getReferenceRecurrenceId()
+	const futureExceptions = master.recurrenceManager
+		.getRecurrenceExceptionList()
+		.filter((exception) => ref.compare(exception.recurrenceId) <= 0)
+	const masterAttendees = master.getAttendeeList()
+	const masterOrganizer = master.getFirstProperty('ORGANIZER')
+	for (const exception of futureExceptions) {
+		exception.deleteAllProperties('ATTENDEE')
+		for (const att of masterAttendees) {
+			exception.addProperty(att.clone())
+		}
+		if (masterOrganizer) {
+			exception.deleteAllProperties('ORGANIZER')
+			exception.addProperty(masterOrganizer.clone())
+		}
+	}
+}
+
 export default defineStore('calendarObjectInstance', {
 	state: () => {
 		return {
@@ -1518,6 +1547,11 @@ export default defineStore('calendarObjectInstance', {
 				// - eventComponent.canCreateRecurrenceExceptions() - Can we create a recurrence-exception for this item
 				if (isForkedItem && eventComponent.canCreateRecurrenceExceptions()) {
 					[original, fork] = eventComponent.createRecurrenceException(thisAndAllFuture)
+
+					// calendar-js leaves existing exceptions untouched on in-place master overrides.
+					if (thisAndAllFuture) {
+						syncMasterParticipantsOntoFutureExceptions(eventComponent)
+					}
 				}
 
 				await calendarObjectsStore.updateCalendarObject({ calendarObject })
--- a/tests/javascript/unit/store/calendarObjectInstance.test.js
+++ b/tests/javascript/unit/store/calendarObjectInstance.test.js
@@ -1,0 +1,107 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { setActivePinia, createPinia } from 'pinia'
+import { syncMasterParticipantsOntoFutureExceptions } from '../../../../src/store/calendarObjectInstance.js'
+
+describe('store/calendarObjectInstance syncMasterParticipantsOntoFutureExceptions', () => {
+
+	beforeEach(() => {
+		setActivePinia(createPinia())
+	})
+
+	const makeProperty = (tag, id) => {
+		const property = { tag, id }
+		property.clone = () => ({ ...property, cloned: true })
+		return property
+	}
+
+	const makeException = (recurrenceId) => ({
+		recurrenceId,
+		deleted: [],
+		added: [],
+		deleteAllProperties(name) {
+			this.deleted.push(name)
+		},
+		addProperty(property) {
+			this.added.push(property)
+		},
+	})
+
+	const makeMaster = ({ attendees, organizer, exceptions }) => ({
+		isMasterItem: () => true,
+		getAttendeeList: () => attendees,
+		getFirstProperty: (name) => (name === 'ORGANIZER' ? organizer : null),
+		recurrenceManager: {
+			getRecurrenceExceptionList: () => exceptions,
+		},
+	})
+
+	const makeEventComponent = (master, referenceRecurrenceId) => ({
+		isExactForkOfPrimary: true,
+		primaryItem: master,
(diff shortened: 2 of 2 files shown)


SebastianKrupinski:
Hi @ndo84bw 

Thank you for the PR, I will review/test this when I have some time.

DerDreschner (approved):
Tested successfully

SebastianKrupinski (changes requested):
Hi @ndo84bw 

I am going to have to block this PR for now. If I am reading the code correctly, there are some side effects of this change.

When this is used on the base event (master in your words) this loops through all exemptions and overrides the attendees and organizer.

The first problem with this is that a recurring exemption can have all kinds of changes, one is that you can have a different organizer for a particular day.

The other major issue is that you are completely overriding attendee responses. 

```
 exception.deleteAllProperties('ATTENDEE')
for (const att of masterAttendees) {
	exception.addProperty(att.clone())
}
```

This code deletes the property competely, and applies the cloned attendees from the base event, this means that you are applying responses from the base event to the exception. 

So if a attendee responded "declined" to a exception, this would then override that with what the status is in the base event 

Essentially the issue is what is already listed in your "Trade-offs" section.

ndo84bw:
Hi @SebastianKrupinski 

I'm not aware that it's possible to change the organizer of an event. (I needed to do that just last week) As a result, I also doubt that you can change the organizer of a single instance in the series. The [RFC 3.2.4.2](https://www.rfc-editor.org/rfc/rfc6638.html#section-3.2.4.2) says:

> All the calendar components in a scheduling object resource MUST contain the same "ORGANIZER" property value when present.

So I'm not going to pay any more attention to the organizer for now.

But what you're saying is important. I'll adjust that. I was so focused on the display in Thunderbird and Calendar that I didn’t consider that.

I noticed something else that's a real problem. Even different attendees for a single occurrence are overwritten.

I will dive back into it and submit revised code.

ndo84bw:
Hi @SebastianKrupinski ,

I looked into this again and ended up going around the basic questions about the expected behavior. 

My conclusion: yes, we still need to clarify when and how the attendees on the exceptions should be changed when the series is changed. But this is not a problem of the Calendar app, because Thunderbird with Nextcloud as the caldav server behaves exactly the same.

Here too, the exceptions do not reach a guest who was invited later. In the invitation mail, the ICS has no EXDATE and no extra VEVENT for the moved occurrence.

I cannot find an existing issue or PR that addresses this. Unless you recommend a different approach, I suggest that you close this PR and that I open a separate issue about recurring events on the server.

SebastianKrupinski:
Hi @ndo84bw 

Thanks, I'm going to close the PR for now. I will get back to this at some point and see if I can come up with something

PR #8489, opened 2026-06-11
fix(css): Shorter events overlapping

### LINKED ISSUE #8229: Short events are not shown correctly in the week view which leads to overlapping events
### Steps to reproduce

1. Create a short 15min event ending at a full hour, e.g. 9:45 - 10:00
2. Check how it is shown in the week view of the calendar

### Expected behavior

The event entry end at the full hour

### Actual behavior

Visually the entry is larger (probably to fit the description text) which leads to overlaps with the followup event

Image

Probably this doesn't happen often with manually created events but is quite common if you use the appointment booking feature and define preparation / followup time like shown in the screenshot

### Calendar app version

_No response_

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

_No response_

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

_No response_

pallavibakale:
Fixes: #8229 

Removed min-height for shorter events to avoid overlapping events and a better view.
Screenshot attached:

Before fix:

1. With 1 hour density setting:
image

2. With 10 mins density setting:
image

After fix:

1. With 1 hour density setting:
Screenshot 2026-06-11 194333

2. With 10 mins density setting:
image


SebastianKrupinski:
Hi @pallavibakale 

Thanks for the PR. I will have some one from the design team review the changes.

SebastianKrupinski:
@nimishavijay @kra-mo 

Please have a look at the proposed changes

nimishavijay:
Good idea! However, I'm not sure if this would be accessible if the min-height is completely removed. Since they are all clickable items, the minimum height should be 24px to make sure it remains accessible (especially important on mobile), and we should also try to center the text in that little container so it doesn't appear cut off as much as possible :)

pallavibakale:
@nimishavijay I agree on keeping the minimum height to 24px. The text is vertically centered, it appears to be clipped when there is no minimum height set. The padding adds the extra space, without padding the text sticks to container which is not recommended.
Could you suggest some other ways to resolve this issue?

nimishavijay:
Just quickly checked with the dev tools and using a 24px min-height doesn't cut off any text, so we can do ahead with that :) the center alignment is too negligible in this case, so we don't have to necessarily adjust it :)

pallavibakale:
> Just quickly checked with the dev tools and using a 24px min-height doesn't cut off any text, so we can do ahead with that :) the center alignment is too negligible in this case, so we don't have to necessarily adjust it :)

Thank you Nimisha for your analysis and approval

pallavibakale:
Hi @SebastianKrupinski , @nimishavijay approves of the changes. Could you take a look at it?

SebastianKrupinski (approved):

PR #8503, opened 2026-06-15
fix(federation): respect federation settings

hamza221:
Hide federated sharing option when server doesn't allow it
## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Service/CalendarInitialStateService.php
+++ b/lib/Service/CalendarInitialStateService.php
@@ -90,6 +90,11 @@ class CalendarInitialStateService {
 			'enableCalendarFederation',
 			true,
 		);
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

lib/Service/CalendarInitialStateService.php
```diff
@@ -90,6 +90,11 @@ public function run(): void {
 			'enableCalendarFederation',
 			true,
 		);
+		$remoteSharesEnabled = $this->appConfig->getValueString(
```
tcitworld: Use `getValueBool` directly instead

SebastianKrupinski (approved):

tcitworld (approved):

PR #8525, opened 2026-06-18
fix(css): Use semantic font weight for event titles

kra-mo:
Since https://github.com/nextcloud-libraries/nextcloud-vue/pull/8469, clickable elements should use this font weight instead of hardcoding bold.

| Before | After |
|--------|--------|
| Screenshot 2026-06-18 at 13 31 44 | Screenshot 2026-06-18 at 13 31 35 | 

(PS: I hope I did this correctly!)

### DIFF
--- a/css/fullcalendar.scss
+++ b/css/fullcalendar.scss
@@ -190,7 +190,7 @@
 
 	.fc-event-title {
 		text-overflow: ellipsis;
-		font-weight: 700;
+		font-weight: var(--font-weight-element);
 	}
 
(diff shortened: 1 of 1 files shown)


kra-mo:
There are some other places in the app which should be updated, such as the week button:

image

Or days, which currently use an underline when hovered instead:

image

But also, all headings should be updated to use `--font-weight-heading`.

I'll leave these to someone more knowledgeable with the app's CSS, though :)

SebastianKrupinski (approved):

PR #8556, opened 2026-06-28
fix(chore): save new event even with empty title

### LINKED ISSUE #8543: Events without title are not created on save
### Steps to reproduce

1. Open the week view
2. Create an event
3. Do not add a title
4. Click "save"

### Expected behavior

Event with name "Untitled event" is created.

### Actual behavior

No event is created. And no error message is shown.

### Calendar app version

6.5.0

### CalDAV-clients used

_No response_

### Browser

Firefox 152.0.1

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

33.0.5

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

When first adding a title and then making it blank again, then it works as expected.
It's probably a front end issue.

https://github.com/user-attachments/assets/69db31b6-550c-40bd-b4ad-dd3032113f64

max65482:
Fixes #8543

### DIFF
--- a/src/mixins/EditorMixin.js
+++ b/src/mixins/EditorMixin.js
@@ -669,6 +669,12 @@ export default {
 				thisAndAllFuture = true
 			}
 
+			// If this is a new event, mark it as dirty so that it is saved even when no changes were made
+			// (e.g. when creating an event and immediately saving it without title etc.)
+			if (this.isNew) {
+				this.calendarObjectInstance.eventComponent.markDirty()
+			}
+
 			this.isLoading = true
 			this.isSaving = true
 			try {


SebastianKrupinski:
Thank for the PR @max65482 

@odzhychko What do you think?

The fix is correct. But I am wondering if that is the correct behavior, should we allow someone to save an event with a empty title? I mean there is a minimum amount of in formation required to have a use full event displayed in the calendar. Would any one ever want a event with no title?

odzhychko:
> Thank for the PR @max65482
> 
> @odzhychko What do you think?
> 
> The fix is correct. But I am wondering if that is the correct behavior, should we allow someone to save an event with a empty title? I mean there is a minimum amount of in formation required to have a use full event displayed in the calendar. Would any one ever want a event with no title?

Hm, I  still think it is (the more) correct behavior, because an event created like this still has a date and a time.
Hypothetically, It could be useful if one wants to create unnamed blockers :shrug:

The next best behavior, I can think of, would be to disable the save button until something has changed.
Still a user could type in some title, then remove the title, save the event and have the same result.
That would be just inconsistent 😄

SebastianKrupinski:
> Hm, I still think it is (the more) correct behavior, because an event created like this still has a date and a time. Hypothetically, It could be useful if one wants to create unnamed blockers 🤷

Okay, if we let people create events with out a title we should force a generic time at creation time (In my option)

> The next best behavior, I can think of, would be to disable the save button until something has changed. Still a user could type in some title, then remove the title, save the event and have the same result. That would be just inconsistent 😄

That might the best solution, the only problem is the mark dirty trigger (this.calendarObjectInstance.eventComponent.markDirty()) is not reactive, I believe. I believe all the components and properties in eventComponent are not reactive.Would need to double check

odzhychko:
> Okay, if we let people create events with out a title we should force a generic time at creation time (In my option)

I found no way to create an event without a date. A time is automatically set, when

- using the "+ Event" button
- clicking (or dragging) inside the calendar view

max65482:
I'd go with the principle of least surprise. Thunderbird, Outlook (old) and Outlook (new) let you create events with empty title. Outlook (new) asks you whether you want to save an event with empty title.

odzhychko:
> I'd go with the principle of least surprise. Thunderbird, Outlook (old) and Outlook (new) let you create events with empty title. Outlook (new) asks you whether you want to save an event with empty title.

Sounds good. Thanks for checking and summarizing it.

max65482:
Not sure I got the essence from discussion. Should we allow unnamed events or prevent them?
I would personally vote for allowing them (people may be used to creating unnamed blockers in other calendar tools).

odzhychko:
> I'd go with the principle of least surprise. Thunderbird, Outlook (old) and Outlook (new) let you create events with empty title. Outlook (new) asks you whether you want to save an event with empty title.

Google Calendar (web) also just allows creation with an empty title.

odzhychko:
> Not sure I got the essence from discussion. Should we allow unnamed events or prevent them? I would personally vote for allowing them (people may be used to creating unnamed blockers in other calendar tools).

There was no explicit conclusion yet...

In summary these two options seem equally good in general:

- just allow to save with an empty title
- warn user when saving with an empty title (using a dialog with a button to confirm and abort)

@nimishavijay Could you please help us choose which is the most appropriate here for Nextcloud Calendar? 😄

src/mixins/EditorMixin.js
```diff
@@ -669,6 +669,12 @@ export default {
 				thisAndAllFuture = true
 			}
 
+			// If this is a new event, mark it as dirty so that it is saved even when no changes were made
+			// (e.g. when creating an event and immediately saving it without title etc.)
+			if (this.isNew) {
```
odzhychko, line 674: nit(non-blocking): I guess, it would better to mark is as "dirty"/"to be saved" on creation.

Would have to look at the code, but `markDirty` in `save` seems a bit makeshift.

SebastianKrupinski: There is no other place for it to fix this issue. Its a balance between the do you want to save your changes mechanism and save the empty title

SebastianKrupinski (approved):

Issue #7679, opened 2025-11-07
Improve description field editing UI with contenteditable while preserving linkify


RobinDev:
### Is your feature request related to a problem? Please describe.

The description field currently uses linkify to automatically convert URLs and other text into clickable links, which is valuable functionality. 

However, this comes at the cost of editing usability – the text jumps between the linkified view and the plain textarea when editing, creating a jarring user experience. This is particularly disruptive when you start editing the description and then switch to another browser tab to gather information – upon returning, the view state may have changed, making it harder to resume editing where you left off.

### Describe the solution you'd like

Implement a `contenteditable` div that synchronizes with the underlying textarea to provide the best of both worlds:

- **Preserve linkify functionality**: URLs and relevant text remain clickable and formatted
- **Improve editing experience**: Users can edit directly in the formatted view without visual jumps
- **Seamless synchronization**: Keep the contenteditable div and textarea in sync to maintain data integrity (if needed, i do not yet dig in the code)

SebastianKrupinski:
Hi @RobinDev 

Thank you for the suggestion.

RobinDev:
I dug into this and found a small, pragmatic fix that works well for my workflow (tested live on Calendar 6.5.0 / Nextcloud 33).

**Root cause.** The description uses `PropertyText.vue` + `PropertyLinksMixin.js`, which swaps between a linkified read-only `<div>` and a plain `<textarea>`. Whenever the textarea loses focus (`@blur` → `textareaHasFocus = false`), `showLinksClickable` turns true again and the field collapses back to the linkified view — that's the jump.

Interesting side note while testing: switching to another browser tab does **not** blur the textarea in Chromium (`document.activeElement` stays the textarea, no `blur` fires), so the disruption is really about focus loss *within* the page (clicking another field), not the tab switch itself.

**Minimal fix.** Keep the **description** as a plain textarea in the editor instead of collapsing to the linkified view. In `src/mixins/PropertyLinksMixin.js`:

```diff
 showLinksClickable() {
 	return this.linkifyLinks
+		&& !this.isDescription
 		&& this.textareaHasFocus === false
 		&& typeof this.value === 'string'
 		&& this.value.length > 4
 		&& findLinks(this.value).length > 0
 },
```

[...]

PR #8581, opened 2026-07-08
fix(editor): keep description editable when switching browser tab

RobinDev:
## Summary

Small, targeted improvement toward #7679.

The description/text field (`PropertyText.vue` + `PropertyLinksMixin.js`) toggles between a **linkified read-only view** and a **plain textarea**. Today, switching to another browser tab fires a `blur` on the textarea, which sets `textareaHasFocus = false` and collapses the field back to the linkified view. When the user comes back to the tab, their edit context (and caret position) is gone — exactly the pain described in #7679:

> This is particularly disruptive when you start editing the description and then switch to another browser tab to gather information – upon returning, the view state may have changed.

## Change

In `handleToggleTextareaFocus`, ignore the focus loss when it is caused by the page being hidden (`document.hidden`), i.e. a tab/window switch. Focus changes *within* the visible page keep the previous behaviour.

```js
if (hasFocus === false && document.hidden) {
    return
}
```

## Notes

- This does **not** implement the full `contenteditable` solution requested in #7679 — it's a low-risk fix for the most disruptive symptom (losing edits on tab switch). A `contenteditable` rework can still follow.
- No behaviour change when `linkifyLinks` is disabled, or for in-page focus moves.

## Checklist
- [x] Code style (tabs, no semicolons) matches surrounding code
- [x] `Signed-off-by` added (DCO)

### DIFF
--- a/src/mixins/PropertyLinksMixin.js
+++ b/src/mixins/PropertyLinksMixin.js
@@ -70,6 +70,13 @@ export default {
 				// do nothing
 				return
 			}
+			// Keep the textarea open when the user merely switches to another
+			// browser tab or window. Otherwise the blur fired while the page is
+			// hidden collapses the field back to the linkified view, and the
+			// edit state (and caret position) is lost when the user returns.
+			if (hasFocus === false && document.hidden) {
+				return
+			}
 			this.textareaHasFocus = hasFocus
 		},
 	},


RobinDev:
Closing this: live testing showed the approach doesn't address the actual trigger. On a tab switch Chrome keeps the textarea focused (no `blur` fires, `document.activeElement` stays the textarea), so a `document.hidden` guard in the blur handler never runs. I'll reopen with a fix that targets the real cause once verified. Sorry for the noise.

Issue #7772, opened 2025-12-08
Room Resources Can Be Double-Booked: No Confirmation or Conflict Prevention


The-1ST:
### Steps to reproduce

1-Create a resource (room) and ensure it's visible/available for booking via calendar sharing.
2-User A books the room for a given time slot.
3-User B, choose the room in a different time then change to the same time as user A without seeing any reservation or receiving a conflict warning, can book the same room for an overlapping time slot.
Neither user receives any indication that a conflict has occurred.

### Expected behavior

Calendar should provide feedback when a room booking is successful or rejected.
If a room is already booked for a chosen time, the UI should prevent double-booking and display a conflict message.
Users should be notified of the booking status of resources.

### Actual behavior

No immediate feedback is given upon booking a resource.
room availability status always shows availability will be check.

<img width="236" height="155" alt="Image" src="https://github.com/user-attachments/assets/80909e1f-1dc1-4a91-a853-a4aa6b6adf2a" />

Multiple users can book the same room for overlapping time slots.

<img width="527" height="24" alt="Image" src="https://github.com/user-attachments/assets/98ca8869-875b-4f7e-8c3f-0778278662d4" />

No conflict messages are shown.

### Calendar app version

6.1.1

### Browser

chrome

### Client operating system

win 11

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

32.0.2

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

nikbinmarjan:
I have this Issue too.

The-1ST:
how come newer issues get triggered and confirmed as bug  but this issue I post get no attention ?!

SebastianKrupinski:
Hi @The-1ST 

All issues get triaged and confirmed... as you only posted this 2 days ago, please be patient... we have limited staff and time to test and confirm tickets...

SebastianKrupinski:
Confirmed

The-1ST:
in addition to this Issue I just find out If using reoccurring events with room it only check room availability for the first occurrence and does not check future occurrence.

@SebastianKrupinski

SebastianKrupinski:
Hi @The-1ST 

Thanks for letting us know, I think room availability with recurring event would be tricky, to both display and handle

The-1ST:
is there any update on this issue?

[...]

PR #8588, opened 2026-07-10
fix(editor): better resource wording and availability shortcut

ChristophWurst:
The pending state for a room/resource read "Availability will be checked", which sounds like a check is in progress. The availability is only checked by the server once the event is saved, so reword it to "Will be booked if available".

Also add a "Check availability" shortcut to the resource's actions menu that opens the existing room availability timeline directly, instead of going through "Show all rooms" and finding the room in the list again.

Ellipsis was also fixed, because it overflowed, and a native title now shows the full text on hover.

Ref https://github.com/nextcloud/calendar/issues/7772

Assisted-by: ClaudeCode:claude-fable-5

## Screenshots

| Before | After |
|--------|--------|
| Bildschirmfoto vom 2026-07-10 11-28-20 | image | 

## How to test

1. Enable *Calendar Resource Management* app
2. Create a test room
3. Open Calendar
4. Create new event
5. Add room
6. Don't click *Save*

## 🤖 AI (if applicable)

### DIFF
--- a/src/views/EditFull.vue
+++ b/src/views/EditFull.vue
@@ -865,6 +865,13 @@ export default {
 		min-width: unset !important;
 	}
 
+	// The cap above is sized for the wide invitees column. In the narrow
+	// resources column long statuses would still overflow onto the actions
+	// menu, so clamp them to the column width instead.
+	.app-full-footer__right :deep(.avatar-participation-status__text) {
+		max-width: calc(var(--total-width) / 3 - var(--column-gap) / 2 - 58px - var(--default-clickable-area)) !important;
+	}
+
 	&__header {
 		display: flex;
 		flex-direction: column;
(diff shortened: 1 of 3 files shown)


src/views/EditFull.vue
```diff
@@ -865,6 +865,13 @@ export default {
 		min-width: unset !important;
 	}
 
+	// The cap above is sized for the wide invitees column. In the narrow
+	// resources column long statuses would still overflow onto the actions
+	// menu, so clamp them to the column width instead.
+	.app-full-footer__right :deep(.avatar-participation-status__text) {
+		max-width: calc(var(--total-width) / 3 - var(--column-gap) / 2 - 58px - var(--default-clickable-area)) !important;
+	}
```
ChristophWurst, line 873: ngl, this is a bodge. but the attendee/resource element is a mess. it is not a proper layout but elements put into place with position:absolute. This needs a restructuring going forward. Ideally we can use a text wrap instead of the ellipsis so the text is always fully readable.

SebastianKrupinski (approved):

PR #8593, opened 2026-07-13
refactor: remove usage of `this._uid` in Vue components

odzhychko:
Resolves the browser warning `[Vue warn]: Property "_uid" was accessed during render but is not defined on instance.` None of the changed code caused a known user-facing issue.

`this._uid` was an internal API and is `undefined` since Vue 3. The recommended way to get unique IDs is `useId()`.

---

In some places, I encountered usage of `this._uid` in some suspicious code that might hint at something forgotten during previous refactoring and maybe a bug. I don't want to look into the details there for now, so I left `XXX` comments.

### DIFF
--- a/src/components/AppNavigation/CalendarList.vue
+++ b/src/components/AppNavigation/CalendarList.vue
@@ -190,10 +190,6 @@ export default {
 			serverCalendars: 'sortedCalendarsSubscriptions',
 		}),
 
-		loadingKeyCalendars() {
-			return this._uid + '-loading-placeholder-calendars'
-		},
-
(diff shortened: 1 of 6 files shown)


SebastianKrupinski:
Hey,

"this._uid" is dead vue 2 code, and should be removed

odzhychko:
> Hey,
> 
> "this._uid" is dead vue 2 code, and should be removed

I removed the non-required usages of `this._uid` and used `useId()` where we still needed it a unique id.

SebastianKrupinski (approved):

PR #8615, opened 2026-07-17
refactor: resolve lint warnings `vue/no-unused-properties`

odzhychko:
### Before

```
✖ 227 problems (0 errors, 227 warnings)
```

### After

```
✖ 207 problems (0 errors, 207 warnings)
```

### DIFF
--- a/src/components/AppNavigation/Settings.vue
+++ b/src/components/AppNavigation/Settings.vue
@@ -25,8 +25,7 @@
 				<NcAppSettingsSection
 					id="settings-modal-general"
 					:name="t('calendar', 'General')">
-					<SettingsTimezoneSelect
-						:isDisabled="loadingCalendars" />
+					<SettingsTimezoneSelect />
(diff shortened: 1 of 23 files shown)


SebastianKrupinski (approved):
Strong!

But, I feel, between all your refactoring and mine, we are going to need to test all the calendar functionality when we are done

odzhychko:
> But, I feel, between all your refactoring and mine, we are going to need to test all the calendar functionality when we are done

Oh, good to know you have that feeling.

If you are currently actively refactoring, it makes sense for me to stop. We then have a lower risk of merged refactoring to break something. So I will stop with big refactoring/cleanups for now and will focus on quickly reviewing what you do.

Also, in the future, I can do refactoring by feature/file-groups instead of by problem types. And then test them directly or even add automated tests. Might also reduce something breaking. I'm not too keen on testing when we are done 😆

PR #8628, opened 2026-07-21
fix(editor): time input for relative reminders to all day events (with cleanups)

odzhychko:
- [ ] Requires #8630 to write new test with TypeScript

- fix(editor): time input for relative reminders to all day events
  - the main fix
- fix(editor): time input for relative reminders to all day events
  - edge case for when user clears the time input (for example using Delete/Backspace)
    This is an issue co
 - refactor: migrate `TimePicker` to Composition API
   - Refactoring cause I touch the file and tested the component anyway

#### Before

- Time is not shown initially
- Time cannot be editing

https://github.com/user-attachments/assets/86b10a8e-42ca-48d7-9793-995c69d2f646

#### After

- Time is shown initially
- Time can be edited and saved
-  Clearing the time input is handled by restoring the last valid value on changing focus
  - Clearing cannot be disabled with native inputs
  - Clearing in general makes sense to have an intermediate input state
    - But in some cases user might continue without inputting a new value

https://github.com/user-attachments/assets/01cf70ba-81ec-44c3-9b99-8c137d88af1e

#### TODO

* [x] add component test for `TimePicker`
  - Good component to get started with component testing


src/components/Shared/TimePicker.vue
```diff
@@ -2,48 +2,42 @@
...
-	emits: ['change'],
+const emit = defineEmits<{
+	change: [value: Date]
+}>()
 
-	data() {
-		return {
-		}
-	},
+const timePickerValue: Ref<Date | null> = ref(date)
```
SebastianKrupinski: Can I suggest a bidirectional computed internal value... It makes the code really clean 

https://github.com/nextcloud/calendar/blob/065b22949c61a733fe56b7dd367b73afb4f55092/src/components/AppointmentConfigModal/DurationInput.vue#L22-L33

src/components/Shared/TimePicker.vue
```diff
@@ -2,48 +2,42 @@
...
-		/**
-		 * Emits a change event for the Date
-		 *
-		 * @param {Date} date The new Date object
-		 */
-		change(date) {
-			this.$emit('change', date)
-		},
-	},
+/**
```
odzhychko, line 41: issue: bogus comment

src/components/Shared/TimePicker.vue
odzhychko: Sure, always welcome 😁 

I actually thought about it, but came to the conclusion that it does not work well here.
Here is what I came up with (and discarded) when trying it out:

```js
const savedTimePickerValue: Ref<Date | null> = ref(date)

function restoreLastValidDate() {
	savedTimePickerValue.value = date
}

const timePickerValue = computed<Date | null>({
	get() {
		return savedTimePickerValue.value
	},
	set(value) {
		savedTimePickerValue.value = value
		if (value === null) {
			return
		}
		emit('change', value)
	},
})
```

I need to track the ~last valid value~ intermediate `null`/cleared state somehow.
So even when using a bidirectional compute I cannot get rid of a real Ref for that state.

Seems bidirectional compute just makes it more complicated here.
But maybe I'm overlooking something. My brain is pretty fried for today 😆

odzhychko: Done. For c2741ef I had to rework the code and bidirectional computed became a good solution.

tests/javascript/unit/components/Shared/TimePicker.test.ts
```diff
@@ -0,0 +1,85 @@
...
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { mount } from '@vue/test-utils'
+import { describe, expect, it } from 'vitest'
+import { nextTick } from 'vue'
+import TimePicker from '@/components/Shared/TimePicker.vue'
+
+describe('TimePicker', () => {
```
odzhychko, line 11: note: What could go wrong reworking such a small component...

Well, adding tests uncovered issue with emitting redundant change events.

Image

tests/javascript/unit/components/Shared/TimePicker.test.ts
```diff
@@ -0,0 +1,85 @@
...
+	it('displays provided date', async () => {
+		const wrapper = mount(TimePicker, {
```
odzhychko, line 13: note: firs component unit test 🥳

src/components/Shared/TimePicker.vue
```diff
@@ -2,48 +2,55 @@
...
+	change: [value: Date]
+}>()
 
-	props: {
-		date: {
-			type: Date,
-			required: true,
-		},
+const cleared: Ref<boolean> = ref(false)
+const timePickerValue = computed<Date | null>({
```
SebastianKrupinski, line 20: Can we call this "internalValue" or something that explains what this value is for: 

The component is called the TimePicker and the internal component is a TimePicker so where does this value belong. 

I know in this small component it make no difference, but it good practice

SebastianKrupinski (approved):
Okay, I am approving. 

But... This pr has really gone out of scope of the title "fix". Its now a refactor to Ts and a implement Ts testing.

tests/javascript/unit/components/Shared/TimePicker.test.ts
SebastianKrupinski: Lol

odzhychko:
> But... This pr has really gone out of scope of the title "fix". Its now a refactor to Ts and a implement Ts testing.

I don't like that to. It blew up unexpectedly while I was trying to add component tests.

 I'm closing this in favor of https://github.com/nextcloud/calendar/pull/8640
 
 Sorry, for having to review it again.
 
 The new PR:
 
 - addresses https://github.com/nextcloud/calendar/pull/8628#discussion_r3630092693
 - can be back ported cleanly

PR #8630, opened 2026-07-21
chore: add support for test written in TypeScript

odzhychko:
- configure Vitest, TypeScript and ESLint
- convert `tests/javascript/unit/utils/url.test.ts` as an example

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "@types/node": "^26.1.1",
     "@vitejs/plugin-vue": "^6.0.7",
     "@vitest/coverage-v8": "^4.1.10",
+    "@vitest/eslint-plugin": "^1.6.23",
     "@vue/compiler-sfc": "^3.5.40",
     "@vue/test-utils": "^2.4.11",
     "css-loader": "^7.1.4",
--- a/src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
+++ b/src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
@@ -58,7 +58,7 @@ import AccountMultiple from 'vue-material-design-icons/AccountMultiple.vue'
 import { principalPropertySearchByDisplaynameOrEmail } from '../../../services/caldavService.js'
 import useCalendarsStore from '../../../store/calendars.js'
 import usePrincipalsStore from '../../../store/principals.js'
-import { urldecode } from '../../../utils/url.js'
+import { urldecode } from '@/utils/url.ts'
 
 export default {
 	name: 'SharingSearch',
--- a/tests/javascript/unit/utils/url.test.ts
+++ b/tests/javascript/unit/utils/url.test.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { urldecode } from '../../../../src/utils/url'
+import { urldecode } from '@/utils/url'
 
 describe('utils/url test suite', () => {
 	it('should decode urls encoded by php', () => {
--- a/tests/tsconfig.json
+++ b/tests/tsconfig.json
@@ -1,0 +1,8 @@
+{
+	"extends": "../tsconfig.json",
+    "include": ["**/*.ts", "**/*.vue"],
+	"compilerOptions": {
+		"rootDir": "..",
+		"types": ["vitest/globals"]
+	},
+}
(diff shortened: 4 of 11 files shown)


package.json
```diff
@@ -26,6 +26,8 @@
     "stats": "rspack build --env production --json js/stats.json",
     "lint": "eslint src",
     "lint:fix": "eslint src --fix",
+    "lint:tests": "eslint tests",
+    "lint:tests:fix": "eslint tests --fix",
```
odzhychko: question(non-blocking): Any preferences on how/when to apply auto-fixable problems and to clean linting problems in general for the tests?

Current state, often related to formatting:

```
✖ 1130 problems (1114 errors, 16 warnings)
  804 errors and 0 warnings potentially fixable with the `--fix` option.
```

I have no strong preference and would probably go with cleaning up files that I touch during development.

tests/javascript/unit/utils/url.test.ts
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { urldecode } from '../../../../src/utils/url'
+import { urldecode } from '@/utils/url'
```
SebastianKrupinski, line 6: Oh, nice the aliases work in tests too!

tests/tsconfig.json
SebastianKrupinski, line 1: Why do we require a different ts configuration?

package.json
SebastianKrupinski: We should just make lint/lint:fix do both, instead of having separate commands. This is how we do it on the PHP side. 

But only do lint fix as a start (unless lint check passes without errors), also do this in a separate PR.

src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
```diff
@@ -58,7 +58,7 @@ import AccountMultiple from 'vue-material-design-icons/AccountMultiple.vue'
 import { principalPropertySearchByDisplaynameOrEmail } from '../../../services/caldavService.js'
 import useCalendarsStore from '../../../store/calendars.js'
 import usePrincipalsStore from '../../../store/principals.js'
-import { urldecode } from '../../../utils/url.js'
+import { urldecode } from '../../../utils/url.ts'
```
SebastianKrupinski: use the alias @/utils/url.ts

I added it to stop the ../../../../ insanity

tests/tsconfig.json
odzhychko: We have `globals: true,` configured in vitest.config.js.
This makes `expect` and so on available in tests without importing it.

`"types": ["vitest/globals"]` in tests/tsconfig.json‎ tells TS that `expect` and so on are available in test without importing it.
But `"types": ["vitest/globals"]` should only apply to `tests` and not `src` so I need a second TSConfig file.

package.json
odzhychko: > But only do lint fix as a start (unless lint check passes without errors), also do this in a separate PR.

Noted. I will apply `eslint tests --fix` in a seperate PR. And after that merge `lint:tests` and `lint` configs.

src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
odzhychko: Done

package.json
odzhychko: I removed `lint:tests` and `lint:tests:fix` commands and will just later add the `tests` folder to the existing `lint` and `lint:fix` commands.

tests/tsconfig.json
SebastianKrupinski: Okay, understood you make to make the vitest testing function available in all tests files

SebastianKrupinski: We could also just import "vitest/globals" in the shared setup file

SebastianKrupinski (approved):
Approved.

One last change if you can ,  "tests/tsconfig.json.license" can this just be a /** */ in side the config file?

tests/tsconfig.json
odzhychko: Technically the configuration tests/tsconfig.json does not import them but only declares them to be available for type checking. I think this cannot be done in the tests/javascript/unit/setup.js file.

The actual global importing is already happening because we configured:

https://github.com/nextcloud/calendar/blob/d407662f9e7fe1ecd31c8f253be7f1d9d2921ad2/vitest.config.js#L23

PR #8656, opened 2026-07-29
fix(settings): display plus icon for  "add delegate" button

odzhychko:
### Before

Screenshot 2026-07-29 at 17 40 41

### After

image


src/components/AppNavigation/Settings/SettingsDelegationSection.vue
odzhychko, line 52: `PlusIcon` is referenced here.

SebastianKrupinski (approved):

PR #8664, opened 2026-08-03
feat(resources): browsable room finder with availability and dropdown filters

odzhychko:


odzhychko:
Just used to test why my actions in a different PR didn't run.

PR #8688, opened 2026-08-05
refactor: seperate source SCSS from generated CSS

odzhychko:
Separate source SCSS from generated CSS.

Sames as Talk does it: https://github.com/nextcloud/spreed/blob/638f591bcd8e513020bd347afc54b291613e85db/src/components/LobbyScreen.vue#L104

Also delete unused 'dashboard.css` which became unused with commit d97533d.


SebastianKrupinski:
Hey, 

Funny enough I was thinking about doing the exact same thing, when I did the css extraction PR. 

But I am wondering if we should either just put everything in "src/scss" or "src/assets/scss" (if we are going to go with an assets folder, we should organize things in it from the start) but these are also technically not really assets, they're source files just like type script or vue files.

odzhychko:
Putting styles in `src/assets` is what the official Vue starter does. So it wouldn't be too unexpected.

image

`src/styles` and `src/assets/styles` might also be candidates.

Anyway, let me know if you have a preference and I will update the commit.
I have no preference here ^^

src/styles/app-full.scss
odzhychko, line 1: Move to src/styles

SebastianKrupinski:
This looks good, as per convo, let:

1. use "src/styles"
2. remove the chunking subfolder in the rspack.config

image

src/styles/app-full.scss
odzhychko: Done

odzhychko:
@SebastianKrupinski Can be reviewed again.

SebastianKrupinski (approved):

PR #8421, opened 2026-06-02
feat(chore): allow duplication of read-only events

max65482:
Fixes #8361
This allows to duplicate read-only events into a writable calendar.

PR #8689, opened 2026-08-06
fix: allways allow duplicaiton

### LINKED ISSUE #8608: Allow to copy events

Sometimes it can be really useful to copy an event. Two examples of my day-to-day work:

1. We have a company wide vacation calendar and for better overview I also have a vacation calendar for our team. At the moment I always have to add vacations to both calendar. It would be quite nice if I could create one calendar entry and than copy it over to the second calendar
2. Sometimes a colleague gets invited to a meeting i should join on short notice. But as the invitation comes from a external person it is hard to extend the invitation. In this case it would be nice if I could simply copy the event from the colleagues calendar over to my calendar

Most convenient option would be a right-click menu, similar to what we have for files these days which offers a copy&pate option. On paste the user should be able to select the target calendar

_No response_

### Additional context

_No response_

SebastianKrupinski:
### Summary
- Resolves: https://github.com/nextcloud/calendar/issues/8608
- Removed the isReadOnly (User should be able to duplicate any event, this is no different then exporting then importing)
- Removed the canCreateRecurrenceException (This is used to control changing recurring events, so it makes no sense on duplication)
- Fixed broken copy logic in copyCalendarObjectInstanceIntoEventComponent
- Added proper view mode detection

### DIFF
--- a/src/mixins/EditorMixin.js
+++ b/src/mixins/EditorMixin.js
@@ -20,7 +20,7 @@ import { removeMailtoPrefix } from '@/utils/attendee.js'
 import { uidToHexColor } from '@/utils/color.js'
 import { dateFactory } from '@/utils/date.js'
 import logger from '@/utils/logger.js'
-import { getPrefixedRoute } from '@/utils/router.js'
+import { getPrefixedRoute, getViewMode, ViewMode } from '@/utils/router.js'
 
 /**
  * This is a mixin for the editor. It contains common Vue stuff, that is
@@ -292,6 +292,21 @@ export default {
 				return null
 			}
 			return this.principalsStore.getPrincipalByUrl(this.selectedCalendar.delegatorUrl)?.userId ?? null
+		},
+		/**
+		 * Returns the mode the editor is currently rendered in
+		 * (authenticated user, public share, embedded share, or widget).
+		 *
+		 * @return {string} One of ViewMode
+		 */
+		viewMode() {
+			return getViewMode(this.$route?.name, this.isWidget)
+		},
+		/**
+		 * @return {boolean}
+		 */
+		canDuplicate() {
+			return this.viewMode === ViewMode.USER
 		},
 		/**
 		 * Returns whether or not the user is allowed to delete this event
@@ -645,7 +660,7 @@ export default {
 		keyboardDuplicateEvent(event) {
 			if (event.key === 'd' && event.ctrlKey === true) {
 				event.preventDefault()
-				if (!this.isNew && !this.isReadOnly && !this.canCreateRecurrenceException) {
+				if (!this.isNew && this.canDuplicate) {
 					this.duplicateEvent()
 				}
 			}
@@ -701,12 +716,24 @@ export default {
 		},
 
 		/**
-		 * Duplicates a calendar-object and saves it
+		 * Duplicates the calendar-object. If the source calendar is
+		 * read-only, the duplicate is created in the first writable calendar.
 		 *
 		 * @return {Promise<void>}
 		 */
 		async duplicateEvent() {
-			await this.calendarObjectInstanceStore.duplicateCalendarObjectInstance()
+			if (!this.canDuplicate) {
+				return
+			}
+
+			const calendarId = this.isReadOnly
+				? (this.calendarsStore.sortedCalendars[0]?.id ?? null)
+				: (this.calendarObject?.calendarId ?? null)
+			await this.calendarObjectInstanceStore.duplicateCalendarObjectInstance({ calendarId })
+
+			// The editor's calendar picker is driven by this.calendarId, which is
+			// separate from the store's calendarObject.calendarId.
+			this.calendarId = this.calendarObject?.calendarId ?? null
 		},
 
 		/**
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -1425,9 +1425,11 @@ export default defineStore('calendarObjectInstance', {
 		/**
 		 * Duplicate calendar-object-instance
 		 *
+		 * @param {object} data The destructuring object
+		 * @param {string} data.calendarId The id of the calendar to duplicate the event into. Must be a writable calendar
 		 * @return {Promise<void>}
 		 */
-		async duplicateCalendarObjectInstance() {
+		async duplicateCalendarObjectInstance({ calendarId }) {
 			const calendarObjectsStore = useCalendarObjectsStore()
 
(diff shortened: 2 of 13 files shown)


src/views/EditSimple.vue
```diff
@@ -82,7 +82,7 @@
 								</template>
 								{{ $t('calendar', 'Export') }}
 							</ActionLink>
-							<ActionButton v-if="!canCreateRecurrenceException && !isReadOnly" @click="duplicateEvent()">
```
odzhychko, line 85: issue: copying events with recurrence doesn't work

It is unexpected input for `copyCalendarObjectInstanceIntoEventComponent`.
See https://github.com/nextcloud/calendar/pull/8606
```
[ERROR] calendar: [Vue error]: Error in component event handler: Error: Illegal argument: Event objects has recurrence related property RRULE. 
```

odzhychko: The reason, I didn't just implemented it in #8606 was, that I didn't know what the expected behavior should be.

Like, should we:

- Copy one occurrence?
- Copy the whole series?
- Copy the occurrence and all future instances?
- Or have some UI for the user to choose?

src/views/EditFull.vue
```diff
@@ -69,7 +69,7 @@
 									</template>
 									{{ $t('calendar', 'Export') }}
 								</NcActionLink>
-								<NcActionButton v-if="!canCreateRecurrenceException && !isReadOnly && !isNew" @click="duplicateEvent()">
+								<NcActionButton v-if="!isNew" @click="duplicateEvent()">
```
odzhychko: issue: This now makes the button also available when user is not logged in.
e.g. `apps/calendar/embed/...` or public links

odzhychko: issue: Doesn't work with shared calendars, because after duplicating the target calendar is still the original calendar and cannot be changed.

https://github.com/user-attachments/assets/51d250bb-699d-4bf3-a4f8-789040395f17

odzhychko (changes requested):

odzhychko:
Found https://github.com/nextcloud/calendar/pull/8421 that implements duplication for read-only events.

src/views/EditSimple.vue
SebastianKrupinski: Hey, yeah this wasn't ready for testing, I posed it just to add it to my board to keep track of, and didn't realize i created it as  a ready to go pr instead of a draft

SebastianKrupinski: You can test it now

src/mixins/EditorMixin.js
```diff
@@ -646,7 +646,7 @@ export default {
 		keyboardDuplicateEvent(event) {
 			if (event.key === 'd' && event.ctrlKey === true) {
```
odzhychko, line 661: note(non-blocking): Shortcuts do not work as expect under MacOS

We should use https://nextcloud-vue-components.netlify.app/#/Composables/useHotKey

Created https://github.com/nextcloud/calendar/issues/8715

src/mixins/EditorMixin.js
```diff
@@ -646,7 +646,7 @@ export default {
...
 				event.preventDefault()
-				if (!this.isNew && !this.isReadOnly && !this.canCreateRecurrenceException) {
+				if (!this.isNew && !this.canCreateRecurrenceException) {
```
odzhychko: issue: Inconsistent/outdated check for `!this.canCreateRecurrenceException`.

Overview of current checks:

- `EditorMixin.keyboardDuplicateEvent`: `!this.isNew && !this.canCreateRecurrenceException`
- `EditFull.vue`: `!isNew` (checked on action)
- `EditSimple.vue`: `!isNew` (checked on action container) 

Might worth introducing `EditorMixin.canDuplicate` to avoid accidental drift.

src/views/EditFull.vue
odzhychko: issue: Duplicate action is available in embedded view (`apps/calendar/embed/...`) and has no effect.

odzhychko: issue: Duplicate action is available in public view (`/apps/calendar/p/...`) and has no effect.

src/mixins/EditorMixin.js
```diff
@@ -702,12 +702,20 @@ export default {
...
+		 * read-only, the duplicate is created in the first writable calendar.
 		 *
 		 * @return {Promise<void>}
 		 */
 		async duplicateEvent() {
-			await this.calendarObjectInstanceStore.duplicateCalendarObjectInstance()
+			const calendarId = this.isReadOnly
+				? (this.calendarsStore.sortedCalendars[0]?.id ?? null)
+				: (this.calendarObject?.calendarId ?? null)
+			await this.calendarObjectInstanceStore.duplicateCalendarObjectInstance({ calendarId })
```
odzhychko, line 732: issue: We  should abort (with a warning), if `calendarId === null` aka. no writable target calendar exists.

Or maybe even dont show that duplicate action to the user, if no writable target calendar exists. 

Could be part of`EditorMixin.canDuplicate` (suggested in https://github.com/nextcloud/calendar/pull/8689/changes#r3754176914)

Such a check would automatically resolve:

- https://github.com/nextcloud/calendar/pull/8689/changes#r3754202561
- https://github.com/nextcloud/calendar/pull/8689/changes#r3754207345

src/store/calendarObjectInstance.js
```diff
@@ -1554,7 +1556,7 @@ export default defineStore('calendarObjectInstance', {
 				end: endDate.unixTime,
 				timezoneId: oldEventComponent.startDate.timezoneId,
 				isAllDay: oldEventComponent.isAllDay(),
-				calendarId: this.calendarObject?.calendarId ?? null,
+				calendarId: calendarId ?? this.calendarObject?.calendarId ?? null,
```
odzhychko: nit: `this.calendarObject?.calendarId` might be of a read only calendar. Let's make `calendarId` mandatory.

Relates to https://github.com/nextcloud/calendar/pull/8689/changes#r3754267389

odzhychko (changes requested):
Copying now works as expected.
And thx for fixing what I broke in #8606 🤦  

In some case the duplicate action is displayed event though duplication is not possible.

odzhychko:
@SebastianKrupinski Let me know, when this becomes reviewable/testable again

SebastianKrupinski:
@odzhychko 

Ready to re-test... this spun out of control a bit... had to add property viewMode detection to know when we are in public mode

odzhychko (approved):
Code looks good and work

PR #8690, opened 2026-08-06
fix(editor): show resource heading always together with search

odzhychko:
Fixes that the subheading "Resources" only appeared after some rooms (or resources where added). This is the transition from state A to B. 

| State | Before | After |
|------|--------|-------|
| A: No room added yet | image | image |
| B: Room added| image | image |

Also fixes some UI issues arising from showing the subheading:

- Visual alignment with the "Attendees" inputs
  -  Button now also in subheading
    - Change to shorter button name to fit better
  - Align buttons to the right
- Replace map-marker-icon with door-open-icon because it already used for "Location"
  - The duplication was previously not too apparent, because "Resources" subheading was often hidden

### DIFF
--- a/css/app-full.scss
+++ b/css/app-full.scss
@@ -130,10 +130,30 @@
 	gap: calc(var(--default-grid-baseline) * 4);
 }
 
-.invitees-list {
+.invitees-list,
+.resource-list {
 	display: flex;
 	flex-direction: column;
 	gap: calc(var(--default-grid-baseline) * 2);
+
+	&__header {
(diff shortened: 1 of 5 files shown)


odzhychko:
@nimishavijay Let me now, if you have design input for this.

odzhychko:
@claude review

GVodyanov (approved):
Always nice to see styling being cleaned up

SebastianKrupinski (approved):
Looks good, do we need to back port?

PR #8693, opened 2026-08-07
feat(principal): map CalDAV room metadata properties

odzhychko:
Make room metadata available on principle objects (e.g. returned by `@/models/principal.js`)

This implementation breaks with a few of the established patterns in `@/models` anticipating ongoing push for type-safety/ TypeScript:

- introduce `RoomPrincipal` type
  - in the future we can have something like `type Principle = RoomPrincipal | RoomPrincipal
- do type validation as early as possible
- be strict with type validation
  - non-ambiguous types prevent parsing and error handling further up 
- do not fail for unexpected types
  - always discard with a warning
  - we are not fully in control of data sources
    - apps might make different assumptions
    - and other CalDav service (integrated through https://github.com/nextcloud/integration_davc) might make different assumptions
   
This PR just adapts what was build in https://github.com/nextcloud/calendar/pull/8263 to align with how we want to extend our models in the future.

@Rikdekker Thank you very much. Here is a summary of things I decide to do differently compared to #8263:

- Keep properties `undefined` if is not a `RoomPrincipal` 
  - `null` is now reserved for when it is a `RoomPrincipal` and a value was not provided or had an unexpected value
- Parse `roomSeatingCapacity` and `roomFeatures`
- Take string values as is without trimming
- Took out logic for `roomBuildingName`
   - Seems more like a guessing of the building from a given convention
     - It is a really good heuristic, but I would rather put it outside the `RoomPrincipal` object
- Adding `roomBuildingStory`
- Taking `roomAddress` directly from the [cdav-library object  ](https://github.com/nextcloud/cdav-library/blob/ce0c5f17fe78ff04c63231accd21b739ff17e22f/src/models/principal.js#L117)
- exposing `roomBuildingRoomNumber` instead of `roomNumber`
  - seems a bit of a redudant property name but I like to stay close the properties names from the cdav-library object
   

## 🤖 AI (if applicable)

### DIFF
--- a/src/models/principal/principal.ts
+++ b/src/models/principal/principal.ts
@@ -1,0 +1,90 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { RoomPrincipalProperties } from '@/types/models/principal'
+
+import logger from '@/utils/logger'
+
+// NOTE File can be merged with @/models/principal.js after it was migrated to TypeScript.
+
+/**
+ * Extract properties specific to a room from a dav principal.
+ *
+ * @param dav cdav-library Principal object
+ * @see {@link https://github.com/nextcloud/cdav-library/blob/main/src/models/principal.js} for the available properties on {@link dav}
+ */
+function mapDavToRoomPrincipalProperties(dav: Record<string, unknown>): RoomPrincipalProperties {
+	const properties: RoomPrincipalProperties = {
+		roomType: null,
+		roomSeatingCapacity: null,
+		roomBuildingAddress: null,
+		roomBuildingStory: null,
+		roomBuildingRoomNumber: null,
+		roomFeatures: null,
+		roomAddress: null,
+	}
+
+	if (typeof dav.roomType === 'string') {
+		properties.roomType = dav.roomType
+	} else if (dav.roomType !== undefined) {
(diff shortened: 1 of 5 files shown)


src/models/principal/principal.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export interface RoomPrincipalProperties {
+	roomType: string | null
+	roomSeatingCapacity: number | null
+	roomBuildingAddress: string | null
+	roomBuildingStory: string | null
+	roomBuildingRoomNumber: string | null
+	roomFeatures: string[] | null
+	roomAddress: string | null
+}
```
SebastianKrupinski: Can we put all the interface in a types file please... "src/types/"

src/models/principal/principal.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+import logger from '@/utils/logger'
+
+// TODO File can be merged with @/models/principal after it was migrated to TypeScript.
+
+/**
+ * Properties shared between all principal types.
+ */
+export interface SharedPrincipalProperties {
```
SebastianKrupinski: Can we call this something like "BasePrincipalProperties" istead of "Shared", at first I through this had something to do with sharing until I read the doc

src/models/principal/principal.ts
odzhychko: Done

src/models/principal/principal.ts
odzhychko: Done

SebastianKrupinski (approved):

PR #8705, opened 2026-08-09
chore: fix eol workflow

SebastianKrupinski:
### Summary
- Adjusted EOL work flow to work with all branches (face palm)

### DIFF
--- a/.github/workflows/block-merge-eol.yml
+++ b/.github/workflows/block-merge-eol.yml
@@ -3,11 +3,10 @@
 
 name: Block merges for EOL branches
 
-on:
-  pull_request:
-    branches:
-      - main
-      - stable6.5
+on: pull_request
+
+env:
+  ACTIVE_BRANCHES: '["main", "stable6.5"]'
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


odzhychko (commented):
> Adjusted EOL work flow to work with all branches (face palm)

Guess, this new an approach also works.
But should the existing one not also work if we require "Block merges for EOL branches" to succeed before merging is this a repo ? (in repo settings > rulesets )

e.g. Someone opens a PR to `stable5.5.`, then "Block merges for EOL branches" does not run and the PR is not allowed to merge because  "Block merges for EOL branches" is required to succeed?

Existing one seems nicer, because it even avoid triggering a runner.

SebastianKrupinski:
> > Adjusted EOL work flow to work with all branches (face palm)
> 
> Guess, this new an approach also works. But should the existing one not also work if we require "Block merges for EOL branches" to succeed before merging is this a repo ? (in repo settings > rulesets )
> 
> e.g. Someone opens a PR to `stable5.5.`, then "Block merges for EOL branches" does not run and the PR is not allowed to merge because "Block merges for EOL branches" is required to succeed?
> 
> Existing one seems nicer, because it even avoid triggering a runner.

The previous approach did not work on branches that where on in the "run on this branch list" so when I BP it to 5.5 the job was stuck in a pending state instead of failing or succeeding

SebastianKrupinski:
Sill not working correctly... grrrr

SebastianKrupinski:
> Sill not working correctly... grrrr

I'm a donkey 🫏 used the wrong branch name!!

odzhychko (approved):

PR #8707, opened 2026-08-09
chore: fix eslint warning no-unused-vars

SebastianKrupinski:
### Summary
- fixes eslint no unsued vars warning

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/AppNavigation/CalendarList/Trashbin.vue
+++ b/src/components/AppNavigation/CalendarList/Trashbin.vue
@@ -171,7 +171,7 @@ export default {
 				let eventSummary = t('calendar', 'Untitled item')
 				try {
 					eventSummary = vobject?.calendarComponent.getComponentIterator().next().value?.title
-				} catch (e) {
+				} catch {
 					// ignore
 				}
 				let subline = vobject.calendar?.displayName || t('calendar', 'Unknown calendar')
--- a/src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
+++ b/src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
@@ -192,7 +192,7 @@ export default {
 			let results
 			try {
 				results = await principalPropertySearchByDisplaynameOrEmail(query)
-			} catch (error) {
+			} catch {
 				return []
 			}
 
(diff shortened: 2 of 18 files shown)


src/components/AppNavigation/CalendarList/Trashbin.vue
```diff
@@ -171,7 +171,7 @@ export default {
 				let eventSummary = t('calendar', 'Untitled item')
 				try {
 					eventSummary = vobject?.calendarComponent.getComponentIterator().next().value?.title
-				} catch (e) {
+				} catch {
 					// ignore
```
odzhychko, line 175: Would be nice to know, why we can ignore the error here.

src/components/AppNavigation/EditCalendarModal/SharingSearch.vue
```diff
@@ -192,7 +192,7 @@ export default {
 			let results
 			try {
 				results = await principalPropertySearchByDisplaynameOrEmail(query)
-			} catch (error) {
+			} catch {
 				return []
```
odzhychko, line 196: Why do we not log the error here? Might help find error causes faster. (Same for all other similar places.)

src/components/Editor/InvitationResponseButtons.vue
```diff
@@ -117,7 +117,7 @@ export default {
 				await this.setParticipationStatus('ACCEPTED')
 				showSuccess(this.t('calendar', 'The invitation has been accepted successfully.'))
 				this.$emit('close')
-			} catch (e) {
+			} catch {
 				showError(this.t('calendar', 'Failed to accept the invitation.'))
```
odzhychko, line 121: Why do we not log the error? Might help with finding error cause faster. (Same for all other similar occurrences.)

odzhychko (approved):
Hm, why not log all the errors? I would expect to either:

1. catch the error and log
2. or have a comment why it doesn't need to be logged
  - . e.g., "Not logging error because `aFunctinoThatCanFail` would already have logged an error
  - . e.g. "Not logging error because error is used for control flow by `aFunctinoThatCanFail`.


I'd suggest to enable https://eslint.org/docs/latest/rules/preserve-caught-error with `requireCatchParameter` `true`.

---

Anyway this refactoring just removes the already unused declarations, so approving. `preserve-caught-error` can be enabled and enforced independently of this PR.

Issue #8522, opened 2026-06-17
Advanced events editor API


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Nextcloud app developer I want to be able to extend the event editors with custom inputs and influence the ICS generation.

### Describe the solution you'd like

TBD

### Describe alternatives you've considered

TBD

### Additional context

Part of https://github.com/nextcloud/calendar/issues/812

ChristophWurst:
> **Note:** This design proposal was drafted with AI assistance (Claude Code, model `claude-opus-4-8`) and is posted for discussion. Details are subject to review and change.

# Proposed design: a framework-agnostic event editor plugin API

A concrete proposal for this ticket. The goal: let any Nextcloud app **add custom inputs to the event editor** and **read/modify the event being edited** — both the standard editor properties *and* arbitrary custom ICS properties — **without depending on Calendar's internals or its frontend framework**.

## Scope (this iteration)

- ✅ Custom editor fields (UI injected into the event editor)
- ✅ First-class access to the editor's working model (standard properties)
- ✅ An ICS serialization hook (influence the generated iCalendar data)
- ❌ Out of scope for now: event context-menu actions, calendar-grid/view plugins, settings hooks. These can reuse the same registry later (see #812).

## Design decisions

1. **Framework-agnostic UI** — plugin UI is a **Web Component (custom element)**. Calendar only renders `<your-tag>` and assigns it a context object. The plugin can be written in vanilla JS, Lit, Svelte, React, Vue — anything that compiles to a custom element. **No coupling to Calendar's Vue version.** (Same mechanism the Files app uses for sidebar tabs.)
2. **Framework-agnostic registration** — a tiny published package whose functions only write to a `window` global. Plugins never import Calendar code.
3. **The editor model is the primary surface** — plugins operate on a stable, reactive projection of the event (the same data the built-in fields edit), not the Pinia store and not (primarily) raw ICS strings.
4. **No restriction on property names** — plugins may read/write any ICS property; docs recommend an `X-<APPID>-…` prefix but it is not enforced.
5. **Load-order-proof** — a drained queue (Viewer pattern) + a live registry (Files pattern), so registration works whether the plugin script runs before or after Calendar boots.

## Public API

Working package name (TBD — see open questions): **`@nextcloud/calendar-plugin`**. Zero runtime deps; ships types + thin functions writing to `window` globals. `EventComponent` re-exported type-only from `@nextcloud/calendar-js`.

[...]

PR #8714, opened 2026-08-10
feat: add navigation action to create calendar events

susnux:
- [x] Requires https://github.com/nextcloud/server/pull/63134

This is a first implementation for the new Nextcloud navigation actions.
It allows to create new calendar events without interrupting the workflow from every other apps by adding a new action to the menu.

Changes:
- refactor code to make it possible to use the modals outside of the calendar app
- add listeners to register the actions
- implement action and action listener to spawn a dialog which allows creating new events

[Bildschirmaufnahme_20260810_222403.webm](https://github.com/user-attachments/assets/d9e49cc8-fb00-4b11-a407-851e54fefc90)

## 🤖 AI (if applicable)


lib/AppInfo/Application.php
```diff
@@ -89,4 +109,11 @@ private function addContactsMenuScript(ContainerInterface $container): void {
 		Util::addScript(self::APP_ID, 'calendar-contacts-menu');
 		Util::addStyle(self::APP_ID, 'calendar-contacts-menu');
 	}
+
+	/**
+	 * Whether the server supports actions in the app menu.
+	 */
+	private function hasAppMenuActions(): bool {
```
SebastianKrupinski: Can we change the name of this function to something more aligned with what it does, "supportsAppMenuActions"

Also add a TODO comment that the check can be removed after support for server 35 and less is dropped

src/components/Editor/Resources/ResourceList.vue
```diff
@@ -140,7 +140,8 @@ export default {
 		},
 
 		resourceBookingEnabled() {
-			return loadState('calendar', 'resource_booking_enabled')
+			// The editor can be opened on pages without the calendar initial state
+			return loadState('calendar', 'resource_booking_enabled', false)
```
SebastianKrupinski: Not sure we need to force a default here since "resource_booking_enabled" is returned by the editor initial state

SebastianKrupinski:
Hey,

The psalm errors are caused because psalm runs in isolation without the server, it gets its public server definitions from nextcloud/ocp package.

https://github.com/nextcloud/calendar/blob/1cd4ee3ee19f1b245445a9c178a380d0a3b2b1a1/.github/workflows/psalm-matrix.yml#L74

But the new interfaces do not exist in the package yet, i guess... your options are to wait or to add exemptions to the psalm file...

Since we are in a hurry for this feature, I would add exemptions

Rikdekker:
Coming here from #8522, where @odzhychko pointed at this PR — thanks, it does indeed do most of the work.

I have been prototyping a week view inside Talk that shows the events tied to your conversations. Reading them is fine; opening one for editing is where it falls apart, because the editor could not be placed outside the Calendar app. This PR removes almost all of what was in the way: `bootstrapEditor()` sets up the stores without `Calendar.vue`, `app-menu.ts` mounts the editor with its own Pinia instance, and `runForEditor()` makes the initial state available on every page. From the outside that is the hard half, and it is done.

The one thing it does not cover is opening an **existing** event, because `isDialog` currently means "new event" by definition:

```js
// EditorMixin.js:479
const isNewEvent = this.isDialog || this.$route?.name?.startsWith('New')
```

and the other branch still reads the id from the router, without the optional chaining the new-event branch got:

```js
// EditorMixin.js:517-518
const objectId = this.$route.params.object
const recurrenceId = this.$route.params.recurrenceId
```

So a dialog mounted outside the router can only create, never open. What makes this look small is that everything underneath is already router-free: `getCalendarObjectInstanceByObjectIdAndRecurrenceId({ objectId, recurrenceId })` (`store/calendarObjectInstance.js:1324`) just takes arguments, and in `EditorMixin` `$router` only appears in `cancel()` and the route guards — not in `save()`. The `isDialog` branch in `cancel()` already sidesteps the former.

Would you be open to separating the two meanings `isDialog` currently carries — "there is no router" and "this is a new event"? Concretely, something like accepting `objectId`/`recurrenceId` as props and treating `$route` as the fallback:

```js
const objectId = this.objectId ?? this.$route?.params?.object
const recurrenceId = this.recurrenceId ?? this.$route?.params?.recurrenceId
const isNewEvent = objectId === undefined
```

I am raising it here rather than in #8522 because this PR already touches exactly these lines, and the change is much smaller now than it would be as a follow-up. It is your call whether it belongs in this PR or a later one — I mainly want to avoid the structure settling in a way that makes existing events an awkward fit afterwards.


*Disclosure: I used AI assistance while investigating this PR and drafting this comment.*

odzhychko:
>  It is your call whether it belongs in this PR or a later one.

Please in an later one 😄 
Background: Routing code will be pulled out in the long run out of `EditorMixin.js`, `FullEdit.vue` and `SimpleEdit.vue`.

PR #8718, opened 2026-08-11
feat: room browser design

nimishavijay:
This is just a visual reference for PR #8264 which needs design feedback. Directly made code changes so the CSS is easier to reference rather than recreate from scratch, feel free to take over it :) Note: I am NOT a professional developer and made this using mostly AI, so take the validity of the code in this with a grain of salt, apologies for any spaghetti 🙈 

The main idea is to move the room browser into a modal so it's easier to see. As far as I could tell the resources were also not considered, so I added that too.

Brief description of the ideal design, other than the nitty-gritty visuals which are best referenced in the screenshots and code:

Dialog:
- [ ] Opens as a dialog titled "Rooms and resources"
- [ ] Always show all available rooms and resources, and remove the "Available" switch
- [ ] Use NcTextInput components wherever relevant like in the minimum capacity
- [ ] Also use verbose text like "Minimum capacity" and "6 out of 8 rooms available" for better translations
- [ ] Picking a room should not change the event straight away. Use a "Done" button in the dialog which actually applied the changes.

Room card:
- [ ] The whole card is clickable and the +/− button at the end is gone
- [ ] "Available" and "Reserved" labels can be removed, only "Unavailable" can be used.
- [ ] Checkmark appears at the end once a room is picked, picking a different room swaps your selection rather than doing nothing
- [ ] Capacity should show as a person icon with a number, instead of "12p"

https://github.com/user-attachments/assets/2149af72-4ebb-4cef-86f0-25f031f0600c

There may be some small papercuts and such later on, but this is the general idea :) 

cc @odzhychko 

## 🤖 AI (if applicable)


SebastianKrupinski:
Thanks @nimishavijay 

We'll take it from here

odzhychko:
Took over and adapted in #8731.

PR #8719, opened 2026-08-11
feat: allow attendees to add guests to an event

kesselb:
Needs https://github.com/nextcloud/server/pull/63167

Organizers get a setting to let attendees of an event invite further people. Attendees can then add guests even though the event is otherwise read-only for them, and remove them again until the event is saved.

The server reports guests off the base instance only, so the option needs the dav attendee_guests capability and is hidden for recurring events.

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/Editor/Invitees/InviteesList.vue
+++ b/src/components/Editor/Invitees/InviteesList.vue
@@ -50,7 +50,7 @@
 		</div>
 
 		<InviteesListSearch
-			v-if="!isReadOnly && hasUserEmailAddress"
+			v-if="(!isReadOnly || canAddGuests) && hasUserEmailAddress"
 			:alreadyInvitedEmails="alreadyInvitedEmails"
 			:organizer="calendarObjectInstance.organizer"
 			@addAttendee="addAttendee" />
@@ -67,6 +67,7 @@
 			:key="invitee.email"
 			:attendee="invitee"
 			:isReadOnly="isReadOnly"
+			:canRemove="isForwardedByMe(invitee)"
 			:organizerDisplayName="organizerDisplayName"
 			:members="invitee.members"
 			:isViewedByOrganizer="isViewedByOrganizer"
@@ -122,6 +123,15 @@ export default {
 			required: true,
 		},
 
+		/**
+		 * Whether the current user may add guests even though the list is
+		 * otherwise read-only for them, because they are an attendee.
+		 */
+		canAddGuests: {
+			type: Boolean,
+			default: false,
+		},
+
 		isSharedWithMe: {
 			type: Boolean,
 			required: true,
@@ -160,6 +170,9 @@ export default {
 			creatingTalkRoom: false,
 			showFreeBusyModel: false,
 			recentAttendees: [],
+			// Guests added in this session, so they can be taken off again before
+			// saving. Once saved, only the organizer can remove an attendee.
+			forwardedAttendees: [],
 		}
 	},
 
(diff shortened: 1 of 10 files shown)


ChristophWurst:
Not sure about the approach tbh. Why do we need an additional "setting"/property and can not leverage the forwarding allowed condition?

src/components/Editor/Invitees/InviteesList.vue
SebastianKrupinski, line 74: Wasn't able to test this, the ability to add a guest is not working:

Image

```
BEGIN:VEVENT
CREATED:20260817T025600Z
DTSTAMP:20260817T025715Z
LAST-MODIFIED:20260817T025715Z
SEQUENCE:3
UID:0c502dd7-9a7c-4488-8536-c4ba1de5db7e
DTSTART;TZID=Asia/Tashkent:20260820T100000
DTEND;TZID=Asia/Tashkent:20260820T110000
TRANSP:OPAQUE
STATUS:CONFIRMED
X-NC-INVITATION-FORWARDING:TRUE
SUMMARY:This is a test event
ATTENDEE;CN=user2@;CUTYPE=INDIVIDUAL;PARTSTAT=NEEDS-ACTION;ROLE=
 REQ-PARTICIPANT;RSVP=TRUE:mailto:user2@
ORGANIZER;CN=User One:mailto:user1@
END:VEVENT
```

src/store/calendarObjectInstance.js
```diff
@@ -397,6 +397,18 @@ export default defineStore('calendarObjectInstance', {
...
+		/**
+		 * Change whether attendees of an event may invite guests
+		 *
+		 * @param {object} data The destructuring object
+		 * @param {object} data.calendarObjectInstance The calendarObjectInstance object
+		 * @param {string} data.allowAttendeeGuests Allow attendee guests value
+		 */
+		changeAllowAttendeeGuests({ calendarObjectInstance, allowAttendeeGuests }) {
+			calendarObjectInstance.eventComponent.updatePropertyWithValue('X-NC-ALLOW-ATTENDEE-GUESTS', allowAttendeeGuests)
+			calendarObjectInstance.allowAttendeeGuests = allowAttendeeGuests
```
SebastianKrupinski, line 385: I think this should be... "X-NC-INVITATION-FORWARDING"

src/mixins/EditorMixin.js
```diff
@@ -241,6 +242,41 @@ export default {
...
+		 * X-NC-ALLOW-ATTENDEE-GUESTS has to be turned on explicitly, so it is also
+		 * absent on invitations we could not report guests back to.
+		 *
+		 * @return {boolean}
+		 */
+		isAddingGuestsAllowed() {
+			// Our own server has to generate the reply carrying the guests
+			if (loadState('core', 'capabilities', {})?.dav?.attendee_guests !== true) {
+				return false
+			}
```
SebastianKrupinski, line 257: Why?

kesselb: You prefer to check against the server version?

src/components/Editor/Invitees/InviteesList.vue
kesselb: The option to add guests as attendee is only shown when the event contains `X-NC-ALLOW-ATTENDEE-GUESTS: TRUE` (which is part of https://github.com/nextcloud/server/pull/63167).

src/mixins/EditorMixin.js
SebastianKrupinski: Oh, this is a check to see if the server supports the guests feature. Okay, now i understand. 

Ummm I am in different to it, I don't think we use this anywhere else... so we can but I would change the flag to something self exploratory... supports_attendee_guests and the comment should reflect what we are doing

src/components/Editor/Invitees/InviteesList.vue
SebastianKrupinski: I created the event with the server PR active, as user 1, then set the forwarding to yes, and logged in as user 2 to add an attendee, but was not able to do so.

I think those are the correct steps??

kesselb: There's a new option below forwarding:

image

Should probably switch the order (guests before invitation forwarding). The original plan was to connect it to the forwarding toggle. I'd switched to a own flag, because the default for forwarding is true (to keep the behavior unchanged for 3rdparty clients) and the flag for add guests as attendee should be off by default.

SebastianKrupinski: I don't think this is showing for me...

I am pretty sure I have the correct branches

image

image

image

PR #8720, opened 2026-08-12
fix(editor): support Cmd key for editor shortcuts on macOS

### LINKED ISSUE #8715: `Ctrl` is used instead of `Cmd` in editor shortcuts under MacOS
### Steps to reproduce

1. Be on MacOS
1. Create an event
2. Open the event
3. Press `Cmd` + `d`

### Expected behavior

Duplicate event.

### Actual behavior

Executes a browser shortcut action. Usually `Cmd` + `d` adds a bookmark.

### Calendar app version

6.6.0

### CalDAV-clients used

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

_No response_

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

We should use https://nextcloud-vue-components.netlify.app/#/Composables/useHotKey

It takes care checking `Cmd` instead of `Ctrl` under MacOS.

Searching for `event.ctrlKey` I found affect shortcuts only in `EditorMixin.js`.

xhon-pelushi:
## Summary
- Register calendar editor shortcuts via `@nextcloud/vue` `useHotKey` so macOS uses Cmd and other platforms use Ctrl (not both).
- Call `useHotKey` directly from `EditFull` and `EditSimple` `setup()`.

Fixes #8715

## Test plan
- [ ] On macOS: open an event editor and confirm Cmd+Enter / Cmd+Delete / Cmd+D behave as expected; Escape still cancels
- [ ] On Windows/Linux: confirm Ctrl shortcuts still work; plain Ctrl must not also fire the Cmd bindings on macOS

### DIFF
--- a/src/mixins/EditorMixin.js
+++ b/src/mixins/EditorMixin.js
@@ -6,6 +6,7 @@
 import { showError, showSuccess } from '@nextcloud/dialogs'
 import { translate as t } from '@nextcloud/l10n'
 import { generateUrl } from '@nextcloud/router'
+import { useHotKey } from '@nextcloud/vue/composables/useHotKey'
 import { mapState, mapStores } from 'pinia'
 import { getRFCProperties } from '@/models/rfcProps.js'
 import { containsRoomUrl } from '@/services/talkService.ts'
@@ -59,6 +60,8 @@ export default {
 			isRecurrenceException: false,
 			// Whether or not the Talk modal is open
 			isTalkModalOpen: false,
+			// Cleanup functions for hotkeys
+			hotKeysRegister: [],
 		}
 	},
 	computed: {
@@ -518,6 +521,19 @@ export default {
 		}
 	},
 
+	mounted() {
+		this.hotKeysRegister = [
+			useHotKey('Escape', () => this.keyboardCloseEditor(), { allowInModal: true }),
+			useHotKey('Enter', () => this.keyboardSaveEvent(), { ctrl: true, allowInModal: true }),
+			useHotKey('Delete', () => this.keyboardDeleteEvent(), { ctrl: true, allowInModal: true }),
+			useHotKey('d', () => this.keyboardDuplicateEvent(), { ctrl: true, prevent: true, allowInModal: true }),
+		]
+	},
+
+	beforeUnmount() {
+		this.hotKeysRegister.forEach((stop) => stop())
+	},
+
 	methods: {
 		/**
 		 * Opens the Talk modal for selecting or creating a Talk room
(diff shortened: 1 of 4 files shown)


src/mixins/EditorMixin.js
```diff
@@ -634,17 +634,17 @@ export default {
 			}
 		},
 		keyboardSaveEvent(event) {
-			if (event.key === 'Enter' && event.ctrlKey === true && !this.isReadOnly && !this.canCreateRecurrenceException) {
+			if (event.key === 'Enter' && (event.ctrlKey || event.metaKey) && !this.isReadOnly && !this.canCreateRecurrenceException) {
```
odzhychko: Hi @xhon-pelushi, thanks for the contribution!

Could you please change it to use https://nextcloud-vue-components.netlify.app/#/Composables/useHotKey ?


You proposed code solves, that `metaKey` + `<other>` is now accepted on MacOS.
But `ctrlKey` + `<other>` is still accepted which is not ideal.

odzhychko (changes requested):

src/mixins/EditorMixin.js
SebastianKrupinski: Also, if we are going to refactor to using "useHotKey" we should do it everywhere on just in the mixin's.

xhon-pelushi:
Thanks for the review — pushed a follow-up that switches the editor shortcuts to `useHotKey` (via a small `useEditorHotKeys` composable used from both `EditSimple` and `EditFull`).

- Escape / Ctrl|Cmd+Enter / Ctrl|Cmd+Delete / Ctrl|Cmd+D are registered through `@nextcloud/vue/composables/useHotKey`
- On macOS that means Cmd only (not Ctrl||Meta)
- The mixin window `keydown` listeners and handlers are removed

Happy to adjust further if needed.

src/composables/useEditorHotKeys.js
SebastianKrupinski, line 1: The editor mixin is already a common library between both editors, why do we now need to extract and abstract this in to a composable? useHotKey is already a composable, so we are wrapping a @nextcloud composable in a local composable?

xhon-pelushi:
Thanks @odzhychko and @SebastianKrupinski — updated to call `useHotKey` directly from `EditFull` / `EditSimple` `setup()` (no local wrapper composable). macOS gets Cmd via `useHotKey`; Ctrl is not also accepted for those bindings.

src/composables/useEditorHotKeys.js
xhon-pelushi: Agreed — the local wrapper is gone. `EditFull` / `EditSimple` now call `@nextcloud/vue` `useHotKey` directly from `setup()`; `useEditorHotKeys.js` was removed.

src/views/EditFull.vue
```diff
@@ -437,6 +439,35 @@ export default {
 		EditorMixin,
 	],
 
+
+	setup() {
```
odzhychko: This doesn't work.

- It has no effect (check out the `allowInModal` option)
- `...proxy` is accessed wrongly
  - throws `Uncaught TypeError: Cannot read properties of null (reading 'proxy')`

odzhychko: No need to duplicate. Can be put it in `setup` in `EditorMixin`.

odzhychko (changes requested):

src/views/EditFull.vue
xhon-pelushi: Good catch, thanks. Both problems were real:

1. `getCurrentInstance()` was being called *inside* the hotkey callbacks, which run later on `keydown`, outside Vue's active-instance context — hence the null `.proxy`. Now resolved once, synchronously, in `useEditorHotKeys()` and closed over.
2. `useHotKey`'s `allowInModal` defaults to `false`, and the editor is rendered as a modal, so none of the shortcuts fired at all — added `allowInModal: true` to all four.

Pushed in 58f4089.

src/mixins/EditorMixin.js
odzhychko: `EditorMixin` was the only place `useHotKey` wasn't use.

odzhychko (dismissed):
@xhon-pelushi https://github.com/nextcloud/calendar/pull/8720#discussion_r3780401003 is still open.

Just inline the logic into `setup` in `EditorMixin` (or is there a reason why it doesn't work?). Also no need for the new code comments.

SebastianKrupinski (approved):

SebastianKrupinski:
> @xhon-pelushi [#8720 (comment)](https://github.com/nextcloud/calendar/pull/8720#discussion_r3780401003) is still open.
> 
> Just inline the logic into `setup` in `EditorMixin` (or is there a reason why it doesn't work?). Also no need for the new code comments.

I made the changes.

SebastianKrupinski:
Note: 

Shortcut keys work, but not so well for recurring events

PR #8731, opened 2026-08-15
Feature/visual room browser v4

### LINKED ISSUE #8724: [Enhancement] Allow Rooms and Resources dropdown to display all options on click/focus without requiring text input

Yes. In corporate or organized environments with a fixed set of meeting rooms or resources, users frequently do not remember the exact name or prefix of a room. Currently, the resource multi-select dropdown field requires a user to type at least one character to trigger the search logic. Clicking an empty field leaves the dropdown completely blank. This creates unnecessary friction for users trying to browse available rooms.

We would like the dropdown for Rooms and Resources inside the event detail edit window to display all available/visible resources instantly upon receiving focus or a mouse click—even if the text search input box is completely empty (query.length === 0). Ideally, this could be handled by:Setting the underlying Vue multi-select/combobox component to open and trigger an unfiltered fetch on focus (open-on-focus="true" or matching component framework equivalent). Allowing the frontend search API payload wrapper to dispatch a lookup when the query string is blank, returning the primary list of resources.

We attempted to evaluate patching the compiled distribution bundle (calendar-main.js) directly to bypass the length checks, but modifying minified logic bounds breaks code integrity, risks breaking core component visibility states, and is entirely unsustainable across app updates. A native implementation or a configuration toggle is the only safe approach.

### Additional context

Nextcloud selfhost version 32.x

This usability behavior is standard across modern scheduling apps (like Google Calendar or Microsoft Outlook), where clicking the location/resource field reveals a scrollable list of pre-configured rooms immediately.

odzhychko:
Cleaned up version of #8264
Resolves #8724

### Before (main)

image

### After

https://github.com/user-attachments/assets/1c5bd483-d668-4090-b5b4-d3be2457fa14

## 🤖 AI (if applicable)


odzhychko:
@SebastianKrupinski Let me know if you find something with regards to code organization.
@nimishavijay Let me know if you find something with regards to UI/UX.

src/utils/roomFilter.ts
```diff
@@ -0,0 +1,384 @@
...
+
+/**
+ * Check whether an address segment is a postal code
+ *
+ * @param segment Address segment to check
+ * @return True if the segment looks like a postal code
+ */
+function isPostalCode(segment: string): boolean {
+	return /^\d{4,6}(\s*[A-Z]{1,2})?$/i.test(segment)
+}
```
SebastianKrupinski: This only matches:

4–6 digit numeric codes (covers most European codes like German 5-digit, or simple numeric formats)
Optionally followed by 1–2 letters, with optional whitespace before them (case-insensitive) — this is specifically the Dutch format (1098 XG)


It does not handle many common formats:

UK postal codes (SW1A 1AA) — letters before digits, doesn't match
Canadian codes (K1A 0B1) — letter-digit-letter pattern, doesn't match
US ZIP+4 (12345-6789) — hyphen not allowed
3-digit codes (some countries) — below the 4-digit minimum

src/utils/roomFilter.ts
```diff
@@ -0,0 +1,384 @@
...
+			parts.push(`${segment} ${next}`)
+			index++
+			continue
+		}
+
+		parts.push(segment)
+	}
+
+	return parts.join(', ')
+}
```
SebastianKrupinski: We are assuming that the address segments are always in the same order

Region | Typical order | vs. assumed order
-- | -- | --
Netherlands (assumed) | Building, Street, Postal code, City | ✅ matches
Germany/France | Street, Postal code + City (one field) | no building segment; postal code usually already joined with city
UK | Street, Town, County, Postcode (postcode is last) | postcode comes after the city, not before
US | Street, City, State + ZIP | ZIP is bundled with state, not directly adjacent to a bare "city" segment
Japan | Postal code first, then Prefecture → City → Ward → Block → Building (largest → smallest, reversed) | postal code is first, and the "building" is last

"150-0001, Tokyo, Shibuya, 1-1 Building"
the first segment "150-0001" isn't even matched as a postal code by isPostalCode (it has a hyphen, and starts with 3 digits), so it gets wrongly picked as the "building name" instead of the actual building at the end.  the merge only fires when a postal-code-shaped segment is immediately followed by another segment,

"221B Baker Street, London, SW1A 2AA"
"SW1A 2AA" doesn't even match the regex, but even if it did, there's no next segment after it — so no merge happens,

src/utils/roomFilter.ts
```diff
@@ -0,0 +1,384 @@
...
+ * Deliberately not the `roomAddress` of the dav principal: that one joins
+ * room number, story and address in that order, which reads as "2.17, 2,
+ * Kerkstraat 10" and is of little use to a map or navigation application.
+ * The address is kept in the order the backend published it, with the room
+ * number appended, so nothing has to be guessed about what each part means.
+ *
+ * @param room Room to build a location for
+ * @return Location string, or null if the room carries no usable address data
+ */
+export function buildRoomLocation(room: RoomOption): string | null {
```
SebastianKrupinski: Inherits all the the flaws already mentioned again.

Also breaks localization with hard coded concatenation

SebastianKrupinski (commented):
Okay I tested this this works kind of:

Image

As mentioned already all the fancy spiting and join does not really work and only works for "Netherlands" style address format.

src/utils/roomFilter.ts
odzhychko: With regards to usage of `isPostalCode` of `deriveBuildingName`:

Yeah, it's only a guess. That's why https://github.com/nextcloud/calendar/issues/8734 (incl. https://github.com/nextcloud/server/pull/63244) is prossued.

odzhychko:
> Okay I tested this this works kind of:

Your screenshot looks of. Should look like in the description have you checked out `feature/visual-room-browser-v4` (**-v4**)?

SebastianKrupinski:
> > Okay I tested this this works kind of:
> 
> Your screenshot looks of. Should look like in the description have you checked out `feature/visual-room-browser-v4` (**-v4**)?

Yeah, I think so... but here is another take

image

src/utils/roomFilter.ts
odzhychko: With regards to usage of `isPostalCode` in `joinAddressSegments` for `buildRoomLocation` lets continue in https://github.com/nextcloud/calendar/pull/8731#discussion_r3793381364

src/utils/roomFilter.ts
odzhychko: #### Some background

1. Why not just put `roomBuildingAddress` into LOCATION?
   - A "good" LOCATION should also contain the room name and maybe the room floor.

#### About current behaviour

1. When adding a room through the resource search we use the `roomAddress`. (`ResourceList.vue`)

```ts
addResource({ commonName, email, calendarUserType, language, timezoneId, roomAddress }) {
	this.calendarObjectInstanceStore.addAttendee({
		...
	})
	this.updateLocation(roomAddress)
}
```

2. `roomAddress` is derived in nextcloud/cdav-library (with regards to localization same bad ^^)

```
roomAddress: {
get: () => {
	const data = [
		this.roomBuildingRoomNumber,
		this.roomBuildingStory,
		this.roomBuildingAddress,
	]
	return data
		.filter((value) => !!value)
		.join(', ')
},
```

#### Going forward: Good solution

I see these possible good solution in the long term for this problem:

1. Have the resource provider always provide a "good" `roomAddress`.
2. Use a localization string like `t('calendar', '{roomBuildingAddress}, {roomBuildingStory} (Room {roomBuildingRoomNumber}', ...)`
3. Use use `roomAddress` (as in 1.) if provided and fallback to (2.).

#### Going forward this PR

I will just use `roomAddress` to be consistant with what happens in `ResourceList.vue` so that we stay consistently bad :laughing: 

Then we can tackle the problem of "Set a good LOCATION for selected room" in a separate issue.

src/utils/roomFilter.ts
odzhychko: Will be removed with together with `buildRoomLocation` and tackled in a separate issue.
See https://github.com/nextcloud/calendar/pull/8731#discussion_r3793688481

odzhychko:
> > > Okay I tested this this works kind of:
> > 
> > 
> > Your screenshot looks of. Should look like in the description have you checked out `feature/visual-room-browser-v4` (**-v4**)?
> 
> Yeah, I think so... but here is another take
> 

Looks still wrong...

image


Seating capacity should be indicate In the bottom left and not top right.

src/utils/roomFilter.ts
odzhychko: > I will just use roomAddress to be consistant with what happens in ResourceList.vue so that we stay consistently 

Done 4f1ac33

src/utils/roomFilter.ts
odzhychko: > Will be removed with together with buildRoomLocation and tackled in a separate issue.

Done 4f1ac33

src/utils/roomFilter.ts
odzhychko: Are you fine with leaving until we can replace it with the building name from #8734?

src/utils/roomFilter.ts
odzhychko: Separate issue in #8735.

src/utils/roomFilter.ts
odzhychko: Uses `roomBuildingName` now as introduced by https://github.com/nextcloud/cdav-library/pull/1085

SebastianKrupinski:
Hey,

Retested this again last night... Found two minor issues...

a) when there are no rooms, we should hide all the search functions

image

b) I have 2 buildings with floor 1... but they are combined in this drop down in to a single floor... since we now have a building name... maybe display the building name and the floor?

Screenshot 2026-08-21 090102

odzhychko:
> a) when there are no rooms, we should hide all the search functions

Fixed. The button is now hidden, if no rooms are available.

image

odzhychko:
> b) I have 2 buildings with floor 1... but they are combined in this drop down in to a single floor... since we now have a building name... maybe display the building name and the floor?

1. Seems you haven't tested with the changes https://github.com/nextcloud/calendar_resource_management/pull/302 . Because then you would see a filter for "Building".

image

2. As we have a filter for building (if building names are provided by the resource provider), a filter for floors that includes the building seems odd. So I'd like to leave it as is.

SebastianKrupinski:
> > b) I have 2 buildings with floor 1... but they are combined in this drop down in to a single floor... since we now have a building name... maybe display the building name and the floor?
> 
>     1. Seems you haven't tested with the changes [Feat/add room building name property calendar_resource_management#302](https://github.com/nextcloud/calendar_resource_management/pull/302) . Because then you would see a filter for "Building".

My bad, I did pull the latest version but forgot to run npm ci

image

Much better now

SebastianKrupinski (approved):

SebastianKrupinski:
Lol. I guess you had auto merge turned on. Cause I just squashed and re-based it for you and it now merged on its own.

PR #8737, opened 2026-08-17
refactor: stop passing redundant `calendarObjectInstance`

odzhychko:
Stop passing redundant `calendarObjectInstance`.

Every editor related component should use `useCalendarObjectInstanceStore#calendarObjectInstance`.

This prepares the stage for "Advanced events editor API" #8522 which requires proper change detection between the model used by our editor and the model passed to the user.

## 🤖 AI (if applicable)

- Refactoring was done manually.
- Esp. the reasoning

## For Reviewers

I tried creating individual commits with refactoring steps.
This might help with the review.

### DIFF
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -35,17 +35,13 @@ import {
 import logger from '@/utils/logger.js'
 import { getBySetPositionAndBySetFromDate, getWeekDayFromDate } from '@/utils/recurrence.js'
 
-// XXX "_" is used for marking actions and state.
-// A better way would be to use helper functions not exposed in `actions`.
-// "_" is still used because it maked refactoring easier to do and to understand.
-
 export default defineStore('calendarObjectInstance', {
 	state: () => {
 		return {
-			_isNew: null,
+			isNew: null,
 			calendarObject: null,
 			calendarObjectInstance: null,
-			_existingEvent: {
+			existingEvent: {
 				objectId: null,
 				recurrenceId: null,
 			},
@@ -61,17 +57,17 @@ export default defineStore('calendarObjectInstance', {
 		 * @param {string} data.objectId The objectId of the calendar-object
 		 * @param {number} data.recurrenceId The recurrence-id of the calendar-object-instance
 		 */
-		_setCalendarObjectInstanceForExistingEvent({
+		setCalendarObjectInstanceForExistingEvent({
 			calendarObject,
 			calendarObjectInstance,
 			objectId,
 			recurrenceId,
 		}) {
-			this._isNew = false
+			this.isNew = false
 			this.calendarObject = calendarObject
 			this.calendarObjectInstance = calendarObjectInstance
-			this._existingEvent.objectId = objectId
-			this._existingEvent.recurrenceId = recurrenceId
+			this.existingEvent.objectId = objectId
+			this.existingEvent.recurrenceId = recurrenceId
 
 			if (this.calendarObjectInstance.eventComponent) {
 				this.calendarObjectInstance.eventComponent = markRaw(this.calendarObjectInstance.eventComponent)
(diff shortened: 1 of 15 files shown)


src/store/calendarObjects.js
```diff
@@ -246,59 +245,6 @@ export default defineStore('calendarObjects', {
...
-		 * Updates the time of the new calendar object
-		 *
-		 * @param {object} data destructuring object
-		 * @param {CalendarObject} data.calendarObjectInstance Calendar-object to
-		 * @param {number} data.start Timestamp for start of new event
-		 * @param {number} data.end Timestamp for end of new event
-		 * @param {string} data.timezoneId asd
-		 * @param {boolean} data.isAllDay foo
-		 */
-		updateTimeOfNewEvent({ calendarObjectInstance, start, end, timezoneId, isAllDay }) {
```
odzhychko, line 259: note: moved to store/calendarObjectInstance.js‎

src/store/calendarObjectInstance.js
```diff
@@ -2115,20 +2003,65 @@ export default defineStore('calendarObjectInstance', {
...
+		/**
+		 * Updates the time of the new calendar object
+		 *
+		 * @param {object} data destructuring object
+		 * @param {number} data.start Timestamp for start of new event
+		 * @param {number} data.end Timestamp for end of new event
+		 * @param {string} data.timezoneId asd
+		 * @param {boolean} data.isAllDay foo
+		 */
+		_updateTimeOfNewEvent({ start, end, timezoneId, isAllDay }) {
```
odzhychko: note: moved from store/calendarObjects.js‎

odzhychko:
@claude review

src/store/calendarObjectInstance.js
```diff
@@ -1192,65 +1153,22 @@ export default defineStore('calendarObjectInstance', {
...
 			}
 		},
 
-		/**
-		 * @deprecated
-		 * @param {object} data The destructuring object
-		 * @param {object} data.calendarObjectInstance The calendarObjectInstance object
-		 * @param {object} data.sharedData The shared file data to attach
-		 */
-		addAttachmentBySharedData({
```
SebastianKrupinski, line 1211: Why did we remove this?

src/store/calendarObjectInstance.js
```diff
@@ -58,17 +61,17 @@ export default defineStore('calendarObjectInstance', {
 		 * @param {string} data.objectId The objectId of the calendar-object
 		 * @param {number} data.recurrenceId The recurrence-id of the calendar-object-instance
 		 */
-		setCalendarObjectInstanceForExistingEvent({
+		_setCalendarObjectInstanceForExistingEvent({
```
SebastianKrupinski: So I get most of the changes, are to use the internal object instance but I don't get why we need to rename some of the functions with a underscore... that usually means the function is private?

src/store/calendarObjectInstance.js
odzhychko: Just wasn't used.

src/store/calendarObjectInstance.js
odzhychko: `calendarObjectInstance` is huge and hard to reason about.
To help me (and maybe us) I declared all function as private (by using `_`) that are only used inside `calendarObjectInstance`.
Now I can look at the `calendarObjectInstance` and see at a glance what is **currently** really public.

But yeah, it's just an as-is assignment and not a should-be assignment of some code architecture perspective.

SebastianKrupinski: I agree... this would help... but can we leave the private function name changes until we refactor the store to TS?

odzhychko: > I agree... this would help... but can we leave the private function name changes until we refactor the store to TS?

Works for me as-well, I will change it back

odzhychko: Done

SebastianKrupinski (approved):

odzhychko:
Bypassing failing PHP tests as they are not related to these changes.

PR #8740, opened 2026-08-18
fix: remove obsolete resource manager

SebastianKrupinski:
### Summary
- Use resource manager directly


SebastianKrupinski:
No longer needed the interface was restored in server

Issue #8029, opened 2026-03-02
[BUG] Appointment booking page broken on mobile with long descriptions


riparise:
### Steps to reproduce

1. Create and appointment with a multi-line description
2. Open the link in a mobile

### Expected behavior

elements should not overlap.

### Actual behavior

This is how the page is formatted for mobile (here, emulating Galaxy S10 Android 11 on Firefox):

<img width="254" height="561" alt="Image" src="https://github.com/user-attachments/assets/f93897dc-d710-4241-8068-3b2588c9fb34" />

On a desktop screen it works fine:

<img width="1245" height="635" alt="Image" src="https://github.com/user-attachments/assets/1a33f4e4-d887-4e61-992f-a5b69859f4bc" />

For those impatient, the following custom css (app -> enable "custom css", then go to settings -> theming) seems to fix. Keep in mind this is 100% slop AI generated quick-and-dirty fix, and not a proper solution.

```css
.booking__date-selection .booking__date-header {
    margin-top: 30px;              /* Adds a little space between description and calendar */
}

.booking__date-selection .mx-datepicker-popup {
    position: relative !important; /* Overrides absolute */
    top: auto !important;          /* Overrides the -274px calculation */
    transform: none !important;    /* Prevents JS positioning transforms */
    margin-top: 30px;              /* Adds a little space between description and calendar */
}

/* Ensure the wrapper takes up the necessary space */
.booking__date-selection .mx-datepicker {
    width: 100%;
    display: block;
}

.booking__date-selection .booking__time-zone {
    display: none !important;
}
```

### Calendar app version

6.2.0

riparise:
loosely related to #7362 and #4142

SebastianKrupinski:
Hi @riparise 

Thank you for the bug report. 

Since you already seem to have a css fix for the issue, you are welcome to put in a PR to fix the issue. Also, FYI, there is a major restructuring coming in 6.3 (Vue 3 migration), so your changes might be short lived.

riparise:
Thanks for the reply. I will wait and if the issue persists in the 6.3 I will open a merge request

PR #8759, opened 2026-08-24
fix(appointments): prevent unreachable overflow on the booking page

Lennart-Pingpong:
## Summary

The public appointment booking page centers its content vertically inside a
container fixed at `100vh`. When the content is taller than the viewport, the
overflow is split evenly above and below the container; the upper part ends up
at a negative offset and cannot be reached by scrolling.

This replaces the fixed heights with `min-height` / `auto` and uses the `safe`
alignment keyword on both vertical-axis centering declarations.

## How it works

`safe center` is identical to `center` whenever the content fits, and degrades
to `start` when it would overflow. The desktop layout is therefore unchanged,
and the behaviour adapts to dynamic content height — long descriptions, or the
slot column expanding after a date is picked — without introducing a
breakpoint.

## Testing

Verified on a production instance running Calendar <VERSION> by applying the
equivalent declarations as custom CSS:

- Desktop: layout unchanged, content still vertically centered
- iOS Safari, multi-line description: whole page reachable, nothing clipped at
  the top
- Slot column expanded after picking a date: no reflow into unreachable space

## Related

Refs #8029

## Browser support

`safe` alignment: Chrome 93+, Firefox 63+, Safari 16.4+.

### DIFF
--- a/src/views/Appointments/Booking.vue
+++ b/src/views/Appointments/Booking.vue
@@ -261,10 +261,11 @@ export default {
 #appointment-booking {
 	display: flex;
 	justify-content: center;
-	align-items: center;
+	align-items: safe center;
 	width: 100%;
 	align-self: flex-start;
 	min-height: 100%;
+	height: auto;
 }
 </style>
 
@@ -273,8 +274,9 @@ export default {
 	display: flex;
 	width: 100%;
 	align-self: stretch;
+	min-height: 100vh;
 	flex-direction: column;
-	justify-content: center;
+	justify-content: safe center;
 	align-items: center;
 }
 


SebastianKrupinski:
Hi @Lennart-Pingpong 

Thank you for the PR. For PR's like this we do ask that you post a before and after image showing the changes.

Lennart-Pingpong:
Hit the same problem from a different angle — on iOS Safari the booking page is
cut off at the top and the missing part cannot be reached by scrolling. I think
I traced the root cause, in case it is useful here.

Two nested flex containers in `src/views/Appointments/Booking.vue` both center
on the vertical axis, and the inner one is pinned to exactly the viewport
height:

```scss
#appointment-booking {
	display: flex;
	justify-content: center;
	align-items: center;   // vertical axis (flex-direction: row)
	height: 100%;
}

.booking__container {
	display: flex;
	flex-direction: column;
	height: 100vh;              // fixed to viewport height
	justify-content: center;    // vertical axis
	align-items: center;
}
```

Once the content is taller than `100vh`, centering splits the overflow evenly
above and below the container. The part above the container origin sits at a
negative offset, and browsers will not scroll into it — so it is structurally
unreachable rather than merely hidden.

On desktop the content fits and nothing is visible. On mobile it exceeds the
viewport as soon as the description runs a few lines or the slot column
expands after picking a date. Depending on the exact content height this
surfaces as overlapping or as clipping, which I suspect is why this issue and
mine looked like different bugs.

The `safe` alignment keyword fixes it: it behaves exactly like `center` while
the content fits and falls back to `start` on overflow.

```diff
 #appointment-booking {
 	display: flex;
 	justify-content: center;
-	align-items: center;
+	align-items: safe center;
 	width: 100%;
-	height: 100%;
+	height: auto;
 }

 .booking__container {
 	display: flex;
 	width: 100%;
-	height: 100vh;
+	min-height: 100vh;
 	flex-direction: column;
-	justify-content: center;
+	justify-content: safe center;
 	align-items: center;
 }
```

Desktop rendering is unchanged and no media query is needed, since this reacts
to actual content height rather than to a breakpoint.

Verified on a production instance (Calendar 6.5.4, iOS Safari) by applying
the equivalent declarations as custom CSS: the full page becomes reachable,
including after the slot column expands, and the desktop layout is unaffected.

One caveat: this addresses the centering overflow only. The
`.mx-datepicker-popup` positioning in the workaround above may be a separate
problem on top of it — I have not reproduced that part.

Happy to open a PR if this direction looks right.

Lennart-Pingpong:
> Hi @Lennart-Pingpong
> 
> Thank you for the PR. For PR's like this we do ask that you post a before and after image showing the changes.

Of course. 

 ### After the fix
i can scroll top to bottom

IMG_9684
IMG_9685


### before the fix

When using my mobile phone the site would just show a fixes fraction of the site. When pressing a date on the calendar even more of my text up top got cropped. 

IMG_9686

SebastianKrupinski (approved):

Lennart-Pingpong:
Hi,

I filled out the survey, but it wouldn't send. but bottomline: I am happy to be able to contribute to this great project. I use nextcloud productively for work and am a fan of your work.

Beste Grüße
Lennart
Lennart Peters
Filmemacher & Medienpädagoge
Luitpoldstraße 53 · 96052 Bamberg
0176 2894 1818
pingpong-workshops.de
Nachricht von github-actions[bot] ***@***.***> am 8. Sept. 2026, 08:07 +0200:

PR #8780, opened 2026-08-26
fix(freebusy): don't render "undefined" as the title of busy blocks

Rikdekker:
The room availability dialog prints the literal string `undefined` inside every busy block.

## Cause

`getBusySlots()` has two paths. The non-bulk one passes `attendees[0].commonName` as `attendeeName`; the bulk one omits the argument entirely:

```js
// src/services/freeBusySlotService.js
if (bulk) {
    events.push(...freeBusyResourceEventSourceFunction(uri, data.calendarData, data.success, startDateTime, endDateTime, timezoneObject))
} else {
    events.push(...freeBusyResourceEventSourceFunction(uri, data.calendarData, data.success, startDateTime, endDateTime, timezoneObject, attendees[0].commonName, isOrganizer))
}
```

`freeBusyResourceEventSourceFunction` assigns that parameter straight to `title`, so for every bulk caller the title becomes `undefined` — which FullCalendar renders as text.

## Why default to an empty string rather than thread a name through

The bulk callers already display the name elsewhere, so repeating it inside each block would be redundant:

- `RoomAvailabilityModal` builds its own `resources[]` list (`attendee.commonName || attendee.uri.slice(7)`) and renders it as the **resource lane header**.
- `ProposalEditor` only reads `resourceId` from the returned events; it never touches `title`.

So the block itself does not need a label — it just must not print `undefined`.

## Tests

Added a regression test covering both the omitted and the supplied name. Verified it fails without the change (`expected undefined to be `) and passes with it. The wider `fullcalendar/` + `services/` suites stay green (130 tests).

## 🤖 AI (if applicable)

### DIFF
--- a/src/fullcalendar/eventSources/freeBusyResourceEventSourceFunction.js
+++ b/src/fullcalendar/eventSources/freeBusyResourceEventSourceFunction.js
@@ -14,10 +14,7 @@ import { getParserManager } from '@nextcloud/calendar-js'
  * @param {Date} start The start of the fetched time-range
  * @param {Date} end The end of the fetched time-range
  * @param {string} timezone Timezone of user viewing data
- * @param {string} [attendeeName] Name of the attendee, used as title for the event.
- *   Omitted by bulk callers, which render the name in the resource lane header
- *   instead — passing undefined through to `title` made FullCalendar print the
- *   string "undefined" inside every busy block.
+ * @param {string} attendeeName name of the attendee, used as title for the event
  * @param {boolean} isOrganizer Whether or not the user is the organizer of the event
  * @return {object[]}
  */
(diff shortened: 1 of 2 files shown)


src/fullcalendar/eventSources/freeBusyResourceEventSourceFunction.js
```diff
@@ -14,11 +14,14 @@ import { getParserManager } from '@nextcloud/calendar-js'
  * @param {Date} start The start of the fetched time-range
  * @param {Date} end The end of the fetched time-range
  * @param {string} timezone Timezone of user viewing data
- * @param {string} attendeeName name of the attendee, used as title for the event
+ * @param {string} [attendeeName] Name of the attendee, used as title for the event.
+ *   Omitted by bulk callers, which render the name in the resource lane header
+ *   instead — passing undefined through to `title` made FullCalendar print the
+ *   string "undefined" inside every busy block.
```
SebastianKrupinski: There is no need to change the doc

SebastianKrupinski (commented):
Looks good, just undo the doc block modification

src/fullcalendar/eventSources/freeBusyResourceEventSourceFunction.js
Rikdekker: Done, reverted in fcd685b66 — the diff is now the default value alone.

SebastianKrupinski (approved):

PR #8781, opened 2026-08-26
chore: rector

SebastianKrupinski:
### Summary
- created rector config
- applied rector config

### DIFF
--- a/lib/Service/Appointments/BookingService.php
+++ b/lib/Service/Appointments/BookingService.php
@@ -20,7 +20,6 @@ use OCA\Calendar\Exception\NoSlotFoundException;
 use OCA\Calendar\Exception\ServiceException;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Http;
-use OCP\DB\Exception;
 use OCP\DB\Exception as DbException;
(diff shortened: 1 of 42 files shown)


lib/Service/Appointments/BookingService.php
```diff
@@ -140,7 +101,7 @@ public function book(AppointmentConfig $config, int $start, int $end, string $ti
 
 		try {
 			$tz = new DateTimeZone($timeZone);
-		} catch (Exception $e) {
+		} catch (DbException $e) {
```
odzhychko, line 104: issue(non-blocking, out-of-scope): `new \DateTimeZone($timeZone)` never throw `OCP\DB\Exception`. It throws `\Exception`.

But rewrite only made this obvious.
Confusion before was caused by this imports:

```
use OCP\DB\Exception;
use OCP\DB\Exception as DbException;
```

---

Similar mixup exists in `BookingController.php`

odzhychko (approved):

lib/Service/Appointments/BookingService.php
odzhychko: Fixed it in https://github.com/nextcloud/calendar/pull/8782

PR #8797, opened 2026-08-28
fix: event creation with alarm

SebastianKrupinski:
### Summary
- Fixes modal crash when creating a new event with calendars set

image

- Partially reverses #8737

### DIFF
--- a/src/store/calendarObjectInstance.js
+++ b/src/store/calendarObjectInstance.js
@@ -1101,17 +1101,19 @@ export default defineStore('calendarObjectInstance', {
 		/**
 		 *
 		 * @param {object} data The destructuring object
+		 * @param {object=} data.calendarObjectInstance The calendar-object-instance to add the alarm to, defaults to the current one
 		 * @param {string} data.type Type of alarm
 		 * @param {number} data.totalSeconds Total amount of seconds for new alarm
 		 * @param {boolean=} data.isDefault Whether this is the default alarm
 		 */
 		addAlarmToCalendarObjectInstance({
+			calendarObjectInstance = this.calendarObjectInstance,
 			type,
 			totalSeconds,
 			isDefault = false,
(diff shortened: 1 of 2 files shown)


src/store/calendarObjectInstance.js
```diff
@@ -1101,17 +1101,19 @@ export default defineStore('calendarObjectInstance', {
 		/**
 		 *
 		 * @param {object} data The destructuring object
+		 * @param {object=} data.calendarObjectInstance The calendar-object-instance to add the alarm to, defaults to the current one
 		 * @param {string} data.type Type of alarm
 		 * @param {number} data.totalSeconds Total amount of seconds for new alarm
 		 * @param {boolean=} data.isDefault Whether this is the default alarm
 		 */
 		addAlarmToCalendarObjectInstance({
+			calendarObjectInstance = this.calendarObjectInstance,
```
odzhychko, line 1110: thought(non-blocking):  `addAlarmToCalendarObjectInstance` and `removeAlarmFromCalendarObjectInstance` should eventually move to `util/alarm.js`.
 
`util/alarm.js#updateDefaultAlarm` should not access the `calendarObjectInstanceStore` because in general stores can access utils, but utils should never access stores.

Also use of `calendarsStore()` can be removed by passing `calendar` as parameter.

odzhychko (approved):
Code looks good. Works as expected.

Thx for fixing what I broke...

src/store/calendarObjectInstance.js
SebastianKrupinski: >util/alarm.js#updateDefaultAlarm should not access the calendarObjectInstanceStore because in general stores can access utils, but utils should never access stores.

I agree. This will need to be refactored at some point.

And mp, we all miss things... 10 years of spaghetti code to fix.

PR #8809, opened 2026-08-30
feat: open tasks in new tab when clicked in calendar view

### LINKED ISSUE #8722: Clicking task in calendar view should open the task in a new tab

Calendar View has a section near the top for "all day" events.

Tasks with deadlines show up in that section. This is fine and helpful.

However, **clicking** one of those tasks in that view forces the current browser tab to perform a **sequence of events**:
1. leave Calendar View,
2. load Task View,
3. display task details.
 
This is poor UX/UI for several reasons:

- It is unexpected: the user probably does not expect to leave the context of the Calendar View.
- It is very different from what happens if the user clicks on a calendar event. The latter just produces a small overlay, which is quick to load and keeps the user in the same context, i.e. Calendar View.
- Gives user no way to stay in same context, i.e. Calendar View.
- There is no way for user to tell in advance whether they are clicking on a task or a calendar event.
- The **sequence of events** (steps 1-3 above) can take a **long time**, especially on old devices or slow network connections.
- If the user clicked on a task by accident (remember, there is no way for the user to tell in advance), then using the browser's Back button causes the browser tab to leave Task View and go to Calendar View, which again can take a **long time**.

Clicking on a task should open a new tab. That tab should show the Task View, with the clicked task's details displayed.

This is probably a very quick and easy fix: just add `target="_blank"` to the task's `<a>` element.

An alternative would be for the task to be shown in an overlay, as calendar events are.

However, this would be more difficult to develop and test than my proposed solution.

### Additional context

Nextcloud 32.0.12.1

max65482:
Fixes #8722 by opening a task in a new tab when clicked in the calendar view

### DIFF
--- a/src/fullcalendar/interaction/eventClick.js
+++ b/src/fullcalendar/interaction/eventClick.js
@@ -115,5 +115,5 @@ function handleToDoClick(event, route, window, isWidget = false) {
 		return
 	}
 	const url = `apps/tasks/calendars/${encodeURIComponent(calendarId)}/tasks/${encodeURIComponent(taskId)}`
-	window.location.href = generateUrl(url)
+	window.open(generateUrl(url), '_blank')
 }
--- a/tests/javascript/unit/fullcalendar/interaction/eventClick.test.js
+++ b/tests/javascript/unit/fullcalendar/interaction/eventClick.test.js
@@ -310,6 +310,7 @@ describe('fullcalendar/eventClick test suite', () => {
 				protocol: 'http:',
 				host: 'nextcloud.testing',
 			},
+			open: vi.fn(),
 		}
 
 		generateUrl
@@ -328,7 +329,8 @@ describe('fullcalendar/eventClick test suite', () => {
 		expect(generateUrl).toHaveBeenCalledTimes(1)
 		expect(generateUrl).toHaveBeenNthCalledWith(1, 'apps/tasks/calendars/reminders/tasks/EAFB112A-4556-404A-B807-B1E040D0F7A0.ics')
 
-		expect(window.location.href).toEqual('/generated-url')
+		expect(window.open).toHaveBeenCalledTimes(1)
+		expect(window.open).toHaveBeenNthCalledWith(1, '/generated-url', '_blank')
 	})
 
 	it('should do nothing when tasks is disabled and route is public', () => {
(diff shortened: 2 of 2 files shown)


src/fullcalendar/interaction/eventClick.js
```diff
@@ -115,5 +115,5 @@ function handleToDoClick(event, route, window, isWidget = false) {
 		return
 	}
 	const url = `apps/tasks/calendars/${encodeURIComponent(calendarId)}/tasks/${encodeURIComponent(taskId)}`
-	window.location.href = generateUrl(url)
+	window.open(generateUrl(url))
```
odzhychko: note(non-blocking, out-of-scope): Eventually it should be an anchor element `<a/>` (
see #8811)

It would be welcome to fix it directly here for opening task, but not required from my side as even this small change is an improvement.

odzhychko (changes requested):
Broken PHP tests are not related and should be fixed after updating the branch.

Broken JS tests are related to this change.

src/fullcalendar/interaction/eventClick.js
SebastianKrupinski: @odzhychko I don't think its possible to use an anchor here. This is a event triggered by full calendar integration

SebastianKrupinski: Also the open command should have a _blank i believe

window.open("https://www.example.com", "_blank");

odzhychko: `_blank` is the default. But but can't be harmful to add it anyway.

> String. A URL that will be visited when this event is clicked by the user. For more information on controlling this behavior, see the [eventClick](https://legacy.fullcalendar.io/v6/eventClick) callback.

https://legacy.fullcalendar.io/v6/event-object has `url` for that.

SebastianKrupinski: Yeah the url would work for tasks, but not so much for regular calendar events

odzhychko: Eventually it would be good to use URLs for events for better usability. But that is even more out of scope for this.

If user does right-click, the client-side routing takes over and behaves the same as now. If user does middle-click event can be opened in new tab.

max65482: Fixed unit tests and added "_blank" parameter.

odzhychko (changes requested):
Code looks good and works as expected. I'm waiting for tests to finish.

Can you change the commit message to use`feat` instead of `fix`? Because this issue is tagged as enhancement.

max65482:
Sure, changed commit message.

odzhychko (approved):

PR #8818, opened 2026-08-31
fix: migrate defaultReminder properly

SebastianKrupinski:
### Summary
- added a migration to migrate the legacy "defaultReminder" user preference
- removed  "defaultReminder" from back end code
- removed "defaultReminder" from front end code


lib/Migration/Version6070Date20260831000000.php
```diff
@@ -0,0 +1,88 @@
...
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select('userid', 'configvalue')
+			->from('preferences')
+			->where($qb->expr()->eq('appid', $qb->createNamedParameter(Application::APP_ID)))
+			->andWhere($qb->expr()->eq('configkey', $qb->createNamedParameter('defaultReminder')));
+
+		$result = $qb->executeQuery();
+
+		$migrated = 0;
+		while ($row = $result->fetch()) {
```
tcitworld, line 67: No chunking?

SebastianKrupinski: Hey!

So in the current design we updated the preferences for each user one by one. Therefore I was just pulling one record at a time.

I thought about a chunked bulk update, but could decide on a deceive way to handle per-existing "defaultReminderPartDay" and "defaultReminderPartDay" 

Maybe, I should revisit the idea.

lib/Service/CalendarInitialStateService.php
```diff
@@ -62,9 +61,8 @@ public function run(): void {
 		$timezone = $this->config->getUserValue($this->userId, $this->appName, 'timezone', $defaultTimezone);
 		$attachmentsFolder = $this->config->getUserValue($this->userId, 'dav', 'attachmentsFolder', '/Calendar');
 		$slotDuration = $this->config->getUserValue($this->userId, $this->appName, 'slotDuration', $defaultSlotDuration);
-		$defaultReminder = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminder', $defaultDefaultReminder);
-		$defaultReminderPartDay = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminderPartDay', $defaultReminder);
-		$defaultReminderFullDay = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminderFullDay', $defaultReminder);
+		$defaultReminderPartDay = $this->config->getUserValue($this->userId, $this->appName, 'defaultReminderPartDay', 'none');
```
odzhychko: Is it intentional, that `defaultReminderPartDay` and `defaultReminderFullDay` do not fall back to the app wide configurations?

odzhychko, line 64: question: Is it intentional, that `defaultReminderPartDay` and `defaultReminderFullDay` do not fall back to the app wide configurations?

It was the case before and is still done in `PublicViewController.php`.
Whatever is correct, it should be consistent I guess.

lib/Migration/Version6070Date20260831000000.php
```diff
@@ -0,0 +1,88 @@
...
+
+	private function migrateAppDefault(): void {
+		if (!$this->appConfig->hasKey(Application::APP_ID, 'defaultReminder')) {
+			return;
+		}
+
+		$value = $this->appConfig->getValueString(Application::APP_ID, 'defaultReminder', 'none');
+
+		if (!$this->appConfig->hasKey(Application::APP_ID, 'defaultReminderPartDay')) {
+			$this->appConfig->setValueString(Application::APP_ID, 'defaultReminderPartDay', $value);
```
odzhychko: Writing `defaultReminderPartDay`/`defaultReminderFullDay` to `$this->appConfig` has now no effect because it is not used anywhere.

SebastianKrupinski: It did originally because it was used in the public controller. But this app wide setting makes no sense. I've removed it. I could not find a reason why it was implemented like this.

lib/Service/CalendarInitialStateService.php
SebastianKrupinski: I've removed the app wide setting. The app wide setting makes no sense this should only be a per user setting. I've removed it. I could not find a reason why it was implemented like this.

lib/Migration/Version6070Date20260831000000.php
SebastianKrupinski: Hey. Have a look at the updated code, this is the best I could do to chunk this, I was reluctant to chuck the writes also.

tcitworld (approved):

PR #8822, opened 2026-09-01
ci: run PHP test only if something related changed

odzhychko:
## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI


.github/workflows/php-test.yml
SebastianKrupinski: Can you change this to a low runner also

SebastianKrupinski (approved):

SebastianKrupinski:
Don't merge this yet... we should make these changes on the rest of the workflows also

https://github.com/nextcloud/calendar/blob/main/.github/workflows/lint-php-cs.yml

https://github.com/nextcloud/calendar/blob/main/.github/workflows/lint-php.yml

I can do it if you are too busy

.github/changes-filters.yml
SebastianKrupinski, line 1: Call this PHP file filters... or backend filters... or something like that... as we should add the same for front end in a separate PR

odzhychko:
Hm, I think I still need to correctly configure how the summaries run so that CI don't run PHP jobs are not stuck...

.github/changes-filters.yml
odzhychko: All filters go in one file. `backend` is the key.

```
backend:
  - '.github/workflows/**'
  - 'appinfo/**'
frontened:
  - ...
...
```

SebastianKrupinski: Okay, that works too!

But I would still add an additional word workflow-filters or changes-filters or something like that

odzhychko:
@SebastianKrupinski Can be reviewed again in.

Here is how it looks when the tests are skipped https://github.com/nextcloud/calendar/pull/8823

.github/workflows/lint-php-cs.yml
```diff
@@ -50,3 +71,17 @@ jobs:
...
+
+  summary:
+    permissions:
+      contents: none
+    runs-on: ubuntu-latest-low
+    needs: lint
+
+    if: always()
+
+    name: php-cs-lint-summary
```
odzhychko, line 83: Adding this job, we will need to change the required jobs that gate merges.

From lint-php-cs/lint to lint-php-cs/summary

.github/changes-filters.yml
odzhychko: I'll change it after the current jobs ran. I want to make sure what I have till now works as expected ^^

SebastianKrupinski (approved):
Looks good to me

From your user CI test branch
Image

.github/workflows/lint-php-cs.yml
SebastianKrupinski: Why do we need a different job gate? Your test Ci ran correctly

odzhychko: https://github.com/nextcloud/calendar/pull/8823 did not open a PR to `main` so. I think we have the gates only for PRs to `main` (and maybe the stable branches?)

PR #8860, opened 2026-09-05
fix(appointments): validate positive durations

### LINKED ISSUE #8857: Appointments: zero increment causes DivisionByZeroError on public slots endpoint
### Steps to reproduce

1. Create or update an appointment configuration with `increment=0` (the controller currently accepts this value).
2. Open the public appointment booking page or request its available slots endpoint.
3. Request slots for any valid future date.

### Expected behavior

The configuration request should be rejected with a validation error, or the public endpoint should return a controlled 4xx response without an application exception.

### Actual behavior

The request is accepted, but slot generation evaluates the increment in modulo/division operations and throws `DivisionByZeroError`:

- `lib/Controller/AppointmentConfigController.php` accepts the value.
- `lib/Service/Appointments/AppointmentConfigService.php` persists it.
- `lib/Service/Appointments/AvailabilityGenerator.php` uses it at the slot-rounding step.

This results in a server error for anyone opening the public booking link.

### Calendar app

### Additional info

Please validate that `increment` is positive (and likely compatible with the slot rounding rules) before persisting it, and add regression coverage for zero and negative values.

Kumar-s75:
## Summary

- Reject non-positive appointment `length` and `increment` values before persistence.
- Apply the validation to both create and update endpoints.
- Add regression tests for invalid increments.

Fixes #8857.

## Testing

- `php -l lib/Controller/AppointmentConfigController.php`
- `php -l tests/php/unit/Controller/AppointmentConfigControllerTest.php`
- `git diff --check`

### DIFF
--- a/lib/Controller/AppointmentConfigController.php
+++ b/lib/Controller/AppointmentConfigController.php
@@ -120,6 +120,40 @@ class AppointmentConfigController extends Controller {
 	}
 
 	/**
+	 * @throws InvalidArgumentException
+	 */
+	private function validateValues(
+		int $length,
+		int $increment,
+		int $preparationDuration,
+		int $followupDuration,
+		int $timeBeforeNextSlot,
+		?int $dailyMax,
+		?int $start,
+		?int $end,
+		?int $futureLimit,
+	): void {
+		if ($length <= 0) {
+			throw new InvalidArgumentException('Length must be greater than zero');
+		}
+		if ($increment < 5 * 60) {
+			throw new InvalidArgumentException('Increment must be at least 5 minutes');
+		}
+		if ($preparationDuration < 0 || $followupDuration < 0 || $timeBeforeNextSlot < 0) {
+			throw new InvalidArgumentException('Durations must not be negative');
+		}
+		if ($dailyMax !== null && $dailyMax <= 0) {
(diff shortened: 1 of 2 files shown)


lib/Controller/AppointmentConfigController.php
SebastianKrupinski, line 195: Why not validate the rest of the values also?

SebastianKrupinski (commented):
Hi @Kumar-s75 

Thank you for the PR. I agree we should validate these values

Kumar-s75:
Thanks for pointing that out. I expanded the validation beyond `length` and `increment` in commit `cd07df504`:

- `increment` must be at least 5 minutes because smaller values round to zero in slot generation.
- Preparation, follow-up, and buffer durations cannot be negative.
- `dailyMax` and `futureLimit`, when provided, must be positive.
- `start` must be before `end` when both are provided.
- Added data-driven regression coverage for the invalid values.

PHP syntax and diff checks pass locally. The focused PHPUnit test is currently blocked because the `phpunit` executable is not available in this checkout.

SebastianKrupinski:
> PHP syntax and diff checks pass locally. The focused PHPUnit test is currently blocked because the `phpunit` executable is not available in this checkout.

Do "composer i" to install phpunit

SebastianKrupinski (approved):

PR #8861, opened 2026-09-05
fix(appointments): handle invalid booking dates

### LINKED ISSUE #8858: Public appointment slots return 500 for an invalid selected date
### Steps to reproduce

1. Open a public appointment booking link.
2. Call the public bookable-slots endpoint with a valid appointment token and timezone, but use an invalid `dateSelected` value such as `not-a-date`.
3. Inspect the HTTP response and server log.

### Expected behavior

The endpoint should reject the input with a controlled `422 Unprocessable Entity` response explaining that the selected date is invalid.

### Actual behavior

The timezone is validated, but parsing `dateSelected` happens outside the exception handler in `lib/Controller/BookingController.php`. PHP throws a date parsing exception, so the request becomes an unhandled server error instead of a validation response.

### Calendar app

### Additional info

The existing controller test covers an invalid timezone but not an invalid date. Wrapping both date constructions in the validation path and adding a regression test would keep malformed public input from producing a 500.

Kumar-s75:
## Summary

- Return a controlled `422 Unprocessable Entity` response for malformed public booking dates.
- Prevent invalid date input from reaching appointment lookup and slot generation.
- Add a regression test for invalid `dateSelected` values.

Fixes #8858.

## Testing

- `php -l lib/Controller/BookingController.php`
- `php -l tests/php/unit/Controller/BookingControllerTest.php`
- `git diff --check`

### DIFF
--- a/lib/Controller/BookingController.php
+++ b/lib/Controller/BookingController.php
@@ -69,12 +69,17 @@ class BookingController extends Controller {
 			$this->logger->error('Timezone invalid', ['exception' => $e]);
 			return JsonResponse::fail('Invalid timezone', Http::STATUS_UNPROCESSABLE_ENTITY);
 		}
-		// Convert selected date to requesters selected timezone adjusted start and end of day in epoch
-		$startTimeInTz = (new DateTime($dateSelected, $tz))
-			->getTimestamp();
-		$endTimeInTz = (new DateTime($dateSelected, $tz))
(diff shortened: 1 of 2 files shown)


SebastianKrupinski (approved):

SebastianKrupinski:
HI @Kumar-s75 

Thank you for the PR.

PR #8862, opened 2026-09-05
fix(appointments): validate availability input

### LINKED ISSUE #8859: Appointment availability accepts malformed slot and timezone data
### Steps to reproduce

1. Create or update an appointment configuration through the appointment configuration API.
2. Submit an availability object with the required top-level/day keys, but use malformed values such as `timezoneId: "Invalid/Zone"` or a slot with non-numeric `start`/`end` values.
3. Open the public booking page or request available slots for that configuration.

### Expected behavior

Invalid availability should be rejected with a controlled `422 Unprocessable Entity` response when the configuration is saved. A malformed persisted record should not make the public endpoint return an uncaught exception.

### Actual behavior

`validateAvailability()` checks the presence of array keys, but does not validate timezone identifiers, slot value types, time ordering, or ranges. The malformed data is persisted. Later, `AvailabilityGenerator` constructs a `DateTimeZone` and performs timestamp arithmetic on the unchecked values, which can throw an exception or produce invalid slot calculations for a public request.

### Calendar app

### Additional info

Relevant code is in `lib/Controller/AppointmentConfigController.php` and `lib/Service/Appointments/AvailabilityGenerator.php`. Please add validation for timezone and slot types/ranges, plus regression tests for malformed availability payloads.

Kumar-s75:
## Summary

- Validate appointment availability timezone identifiers.
- Reject malformed weekday slot containers and slots with non-integer or reversed timestamps.
- Add regression tests for invalid timezone, day, and slot values.

Fixes #8859.

## Testing

- `php -l lib/Controller/AppointmentConfigController.php`
- `php -l tests/php/unit/Controller/AppointmentConfigControllerTest.php`
- `git diff --check`

### DIFF
--- a/lib/Controller/AppointmentConfigController.php
+++ b/lib/Controller/AppointmentConfigController.php
@@ -82,16 +82,35 @@ class AppointmentConfigController extends Controller {
 		if ($expectedKeys !== $actualKeys) {
 			throw new InvalidArgumentException('Invalid value for availability');
 		}
+		if (!is_string($availability['timezoneId']) || $availability['timezoneId'] === '') {
+			throw new InvalidArgumentException('Invalid value for availability timezone');
+		}
+		try {
+			new \DateTimeZone($availability['timezoneId']);
+		} catch (\Exception $e) {
+			throw new InvalidArgumentException('Invalid value for availability timezone', 0, $e);
+		}
 
 		$expectedDayKeys = ['FR', 'MO', 'SA', 'SU', 'TH', 'TU', 'WE'];
+		if (!is_array($availability['slots'])) {
+			throw new InvalidArgumentException('Invalid value for availability slots');
+		}
 		$actualDayKeys = array_keys($availability['slots']);
 		sort($actualDayKeys);
 		if ($expectedDayKeys !== $actualDayKeys) {
 			throw new InvalidArgumentException('Invalid value for availability slots');
 		}
+		foreach ($availability['slots'] as $daySlots) {
+			if (!is_array($daySlots)) {
+				throw new InvalidArgumentException('Invalid value for availability slots');
+			}
+		}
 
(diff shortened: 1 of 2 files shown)


lib/Controller/AppointmentConfigController.php
```diff
@@ -82,16 +82,35 @@ private function validateAvailability(array $availability): void {
...
 		}
+		foreach ($availability['slots'] as $daySlots) {
+			if (!is_array($daySlots)) {
+				throw new InvalidArgumentException('Invalid value for availability slots');
+			}
+		}
 
 		$slots = array_merge(...array_values($availability['slots']));
 		foreach ($slots as $slot) {
+			if (!is_array($slot) || !isset($slot['start'], $slot['end']) || !is_int($slot['start']) || !is_int($slot['end']) || $slot['start'] >= $slot['end']) {
```
SebastianKrupinski, line 111: The extra "!is_array($slot) ||" is not needed isset already checks for this.

SebastianKrupinski:
Hi @Kumar-s75 

Thank you for the PR

Kumar-s75:
Addressed in commit `2a6874454`. Removed the redundant `!is_array($slot)` condition as suggested; the existing `isset()` and type checks remain unchanged. PHP syntax and diff checks pass locally. The focused PHPUnit test is still blocked because `phpunit` is not available in this checkout.

SebastianKrupinski:
> Addressed in commit `2a6874454`. Removed the redundant `!is_array($slot)` condition as suggested; the existing `isset()` and type checks remain unchanged. PHP syntax and diff checks pass locally. The focused PHPUnit test is still blocked because `phpunit` is not available in this checkout.

Looks like removing the array check made psalm unhappy. I dropped the second commit

SebastianKrupinski (approved):

PR #8871, opened 2026-09-06
fix(editor): prevent close button overlap

### LINKED ISSUE #8799: Overlap of close icon and our coop title
Image

Moving the close button below the main calendar-picker-header div fixes the issue.

Image

abhinavohri:
## Summary

- add a trailing actions slot to the calendar picker header
- move the simple editor close action into the header flex layout
- keep long calendar names from overlapping the close control

Fixes #8799

Before:
image

After:
image

### DIFF
--- a/src/components/Editor/CalendarPickerHeader.vue
+++ b/src/components/Editor/CalendarPickerHeader.vue
@@ -82,6 +82,9 @@
 			:user="ownerUserId"
 			:displayName="ownerDisplayName"
 			:size="24" />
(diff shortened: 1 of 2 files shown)


SebastianKrupinski (changes requested):
Hi @abhinavohri 

Thanks for the PR, I see what you are trying to do here, but I am not crazy about the solution.

I would prefer to keep all the actions in the editor view, instead of having them scattered in the components

You can just add a dummy blank space to the calendar picker header to reserve the space

SebastianKrupinski (approved):

PR #8880, opened 2026-09-06
fix: make period header translation reactive

### LINKED ISSUE #8756: Month name in the period header is untranslated on first render (store defaults to 'en' until mounted())
### Steps to reproduce

1. Set the user's language and locale to a non-English one (Russian in my case). The same happens when the instance defaults are used (`occ config:system:set default_language --value ru` / `default_locale --value ru_RU`) and the user never picked anything.
2. Open the Calendar app in month view with a **fresh page load**.
3. Look at the period label between the ◀ / ▶ buttons.

### Expected behavior

The month name follows the user's locale, e.g. `август 2026`.

### Actual behavior

On the first render the month name is in **English** — `August 2026` — while everything around it is correctly localized (weekday headers, buttons, app title, the rest of the UI).

Clicking ▶ (or ◀) re-renders the label correctly: `сентябрь 2026`. Reloading the page brings the English name back. So the label is only wrong until the first period change.

### Root cause (from reading the source)

The label is rendered by `formatDateRange(date, view.value, settingsStore.momentLocale, false)` in `src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderDatePicker.vue`, which ends up in `moment(value).locale(locale).format('MMMM YYYY')` (`src/filters/dateRangeFormat.js`).

`settingsStore.momentLocale` is initialized to the literal `'en'` in `src/store/settings.js`, and only replaced with the real locale later, in `Calendar.vue`:

```js
async mounted() {
    ...
    await this.loadMomentLocale()   // -> settingsStore.setMomentLocale({ locale })
}
```

The first render of the header happens while the store still holds `'en'`. The picker receives the formatter as a prop (`:format="dateFormatWrapper"`), so the already-rendered label is not recomputed when the store value changes a moment later — it stays English until something forces a re-render, which is exactly what pressing ▶ does.

### Ruled out (all checked in the browser console on the affected page)

* `OC.getLocale()`, `OC.getLanguage()` and `document.documentElement.lang` are all `ru`
* the Russian moment locale chunk is fetched fine: `calendar-node_modules_moment_locale_ru_js.js` → `200`
* `moment().locale('ru').format('MMMM YYYY')` returns `август 2026` on that very page

So moment has the locale data, the app just formats the first label before the store learns about it.

### Calendar app version

6.5.4

### CalDAV-clients used

_No response_

### Browser

Chromium (headless and normal, same result)

### Client operating system

macOS

### Server operating system

Debian (official `nextcloud:34.0.2-apache` Docker image)

### Nextcloud version

34.0.2

odzhychko:
When the moment.js locale was loaded after first rendering the header did not rerender displaying the wrong translation.

Fixes #8756

#### Before

Did not automatically updates translation once moment.js locale is available.
It only changed translation after once changing the date.

https://github.com/user-attachments/assets/5c99977c-76e2-49cd-9821-c24b719463fc

#### After 

Automatically updates translation once moment.js locale is available.

https://github.com/user-attachments/assets/d3cba35a-e16a-4d67-928f-b318150e140f

### DIFF
--- a/src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderDatePicker.vue
+++ b/src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderDatePicker.vue
@@ -45,9 +45,14 @@ const view = computed<string>(() => {
 	return route.params.view
 })
 
-function dateFormatWrapper(date: Date): string {
-	return formatDateRange(date, view.value, settingsStore.momentLocale, false)
-}
+type NcDateTimePickerFormatProp = InstanceType<typeof NcDateTimePicker>['$props']['format']
+const dateFormatFn = computed<NcDateTimePickerFormatProp>(() => {
+	const momentLocaleValue = settingsStore.momentLocale
+	const viewValue = view.value
+	return function(date: Date) {
+		return formatDateRange(date, viewValue, momentLocaleValue)
+	}
+})
 
 const previousLabel = computed(() => {
 	switch (view.value) {
@@ -172,7 +177,7 @@ useHotKey(['p', 'k'], () => navigateTimeRangeBackward())
 		</NcButton>
 		<NcDateTimePicker
 			class="datepicker-button-section__datepicker"
-			:format="dateFormatWrapper"
+			:format="dateFormatFn"
 			:modelValue="selectedDate"
 			:type="view === 'multiMonthYear' ? 'year' : 'date'"
 			@update:modelValue="navigateToDate" />


SebastianKrupinski (approved):

SebastianKrupinski:
Was this just caused by how the function was loaded after the typescript refactor? or would this happen in 6.5 also?

odzhychko:
> Was this just caused by how the function was loaded after the typescript refactor? or would this happen in 6.5 also?

This bug was not introduced during the refactor.
It was introduced during the migration to Vue 3.
So stable6.5 is also affected.

Issue #7534, opened 2025-10-07
Hide cancelled and declined events/meetings from calendar view


SebastianKrupinski:
### Is your feature request related to a problem? Please describe.

As a user I might have lots of cancelled or declined calendar events, cluttering up the calendar view. 

### Describe the solution you'd like

Add an option to calendar settings to hide cancelled or declined events from the calendar view.

### Describe alternatives you've considered

None available

PR #8884, opened 2026-09-07
feat: hide declined events

madbob:
Adding an option to hide declined appointments.

With reference to #7534, I've preferred not to hide also cancelled events, to avoid hiding to the user informations without his active intervention. If an event is cancelled by someone else, I may want to explicitely know it has been cancelled instead of just no longer see it on my calendar and wondering where the event I remember has gone.


GVodyanov:
Hey! Thanks for the PR, it looks good, however could you please make it so that by default declined events are shown for consistency?

GVodyanov (commented):
Also I think the toggle on first load is showing inverted state, when turned on it disables the events

lib/Controller/PublicViewController.php
```diff
@@ -87,6 +87,7 @@ private function publicIndex(string $token):PublicTemplateResponse {
 		$defaultSlotDuration = $this->config->getAppValue($this->appName, 'slotDuration', '00:30:00');
 		$defaultDefaultReminder = $this->config->getAppValue($this->appName, 'defaultReminder', 'none');
 		$defaultShowTasks = $this->config->getAppValue($this->appName, 'showTasks', 'yes');
+		$defaultShowDeclinedAppointments = $this->config->getAppValue($this->appName, 'showDeclinedAppointments', 'yes');
```
SebastianKrupinski, line 90: This should just be a user preference, there is no need to make this globally configurable

lib/Service/CalendarInitialStateService.php
```diff
@@ -50,6 +50,7 @@ public function run(): void {
 		$defaultSlotDuration = $this->config->getAppValue($this->appName, 'slotDuration', '00:30:00');
 		$defaultDefaultReminder = $this->config->getAppValue($this->appName, 'defaultReminder', 'none');
 		$defaultShowTasks = $this->config->getAppValue($this->appName, 'showTasks', 'yes');
+		$defaultShowDeclinedAppointments = $this->config->getAppValue($this->appName, 'showDeclinedAppointments', 'yes');
```
SebastianKrupinski, line 53: This should just be a user preference, there is no need to make this globally configurable

SebastianKrupinski:
Setting off:
image

Setting on:
Screenshot 2026-09-14 095406

SebastianKrupinski (approved):

PR #8886, opened 2026-09-07
fix: hide invitation reply buttons on cancelled event

### LINKED ISSUE #4420: Don't display PARTSTAT buttons when an event has STATUS=CANCELLED

Currently, cancelled events still have the "accept", "decline" and "tenative" buttons shown when an event has been cancelled by the ORGANIZER.

This event was cancelled, but the modal still shows actions are possible:
image

image

For a STATUS=CANCELLED for the event, the modal display is changed to clearly indicate such a status. The "accept", "decline" and "tenative" buttons are not shown any more. Instead we could display a "delete event" button.

none

### Additional context

_No response_

madbob:
Do not include `InvitationResponseButtons` if the event has been cancelled, and add a "This event was cancelled" notice also in `EditFull` (copied from `EditSimple`).

Note: changes to the close button for mobile were required to properly accomodate the textual notice.

Fixes #4420

### DIFF
--- a/src/views/EditFull.vue
+++ b/src/views/EditFull.vue
@@ -161,7 +161,7 @@
 					</div>
 
 					<InvitationResponseButtons
-						v-if="isViewedByAttendee"
+						v-if="isViewedByAttendee && !isCancelled"
 						:attendee="userAsAttendee"
 						:calendarId="calendarId"
 						:narrow="true"
--- a/src/views/EditSimple.vue
+++ b/src/views/EditSimple.vue
@@ -216,7 +216,7 @@
 							:calendar="selectedCalendar" />
 
 						<InvitationResponseButtons
-							v-if="isViewedByAttendee && isViewing"
+							v-if="isViewedByAttendee && isViewing && !isCancelled"
 							class="event-popover__response-buttons"
 							:attendee="userAsAttendee"
 							:calendarId="calendarId"
@@ -371,7 +371,6 @@ export default {
 			boundaryElement: null,
 			isVisible: true,
 			isViewing: true,
-			isCancelled: false,
 			closeMask: false,
 			showCancelDialog: false,
 			cancelButtons: [
@@ -460,7 +459,6 @@ export default {
 			this.hasDescription = false
 			this.hasAttendees = false
 			this.hasAlarms = false
-			this.isCancelled = false
 
 			if (this.calendarObjectInstance) {
 				if (typeof this.calendarObjectInstance.location === 'string' && this.calendarObjectInstance.location.trim() !== '') {
@@ -474,9 +472,6 @@ export default {
 				}
 				if (Array.isArray(this.calendarObjectInstance.alarms) && this.calendarObjectInstance.alarms.length > 0) {
 					this.hasAlarms = true
-				}
-				if (this.calendarObjectInstance.status === 'CANCELLED') {
-					this.isCancelled = true
 				}
 
 				// Reposition after content changes
--- a/src/mixins/EditorMixin.js
+++ b/src/mixins/EditorMixin.js
@@ -140,6 +140,14 @@ export default {
 			return this.calendarObjectInstance?.isAllDay ?? false
 		},
 		/**
+		 * Returns whether or not the event has been cancelled
+		 *
+		 * @return {boolean}
+		 */
+		isCancelled() {
+			return this.calendarObjectInstance.status === 'CANCELLED'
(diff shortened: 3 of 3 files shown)


src/views/EditFull.vue
```diff
@@ -657,6 +661,10 @@ export default {
 			}).length === 0
 		},
 
+		isCancelled() {
+			return this.calendarObjectInstance.status === 'CANCELLED'
+		},
+
```
SebastianKrupinski: This should be in editor mixin's, otherwise this only works in one editor and you are using it in both editors

src/views/EditFull.vue
```diff
@@ -794,7 +802,8 @@ export default {
 			gap: calc(var(--default-grid-baseline) * 2) !important;
 
 			&-close-icon {
-				visibility: visible;
+				display: flex;
+				justify-content: flex-start;
```
SebastianKrupinski: This seems unrealted

src/views/EditFull.vue
```diff
@@ -882,9 +892,7 @@ export default {
 
 			&-close-icon {
 				width: calc(var(--default-grid-baseline) * 4 + 21px);
-				display: flex;
-				justify-content: flex-start;
-				visibility: hidden;
+				display: none;
```
SebastianKrupinski: These changes also seem unrelated

src/views/EditSimple.vue
```diff
@@ -216,7 +216,7 @@
 							:calendar="selectedCalendar" />
 
 						<InvitationResponseButtons
-							v-if="isViewedByAttendee && isViewing"
+							v-if="isViewedByAttendee && isViewing && !isCancelled"
```
SebastianKrupinski, line 219: This won't work because isCancelled is inside another vue component and not editor mixins

SebastianKrupinski (commented):
Hi @madbob 

Thanks for the PR, but there are a couple of issues with it

src/views/EditSimple.vue
madbob: It worked as `isCancelled` was [defined](https://github.com/nextcloud/calendar/blob/e6c338c2e13e8d46bd9ebcd3f3511a0951bc51db/src/views/EditSimple.vue#L463) in `EditSimple` before my intervention.
But I've moved it to `EditorMixin` to avoid useless duplication.

SebastianKrupinski: Ah. Okay, I wasn't aware we where doing that. When I do a review from the website its sometimes hard to get the full picture

SebastianKrupinski:
@madbob I see you addressed the editor mixins' but why do we need the css changes?

madbob:
The reasons of CSS changes lies in

```
.app-full__header__top__first {
  flex-direction: column;
}
```

added to apply a vertical layout to the header (and have the "event cancelled" notice below the title).

Keeping `visibility: hidden` to the "Close" button, it still occupies space in the page (even if invisible) and this alters the alignment of the whole block with the block aside (the one with Save/Update buttons).

Screenshot From 2026-09-08 17-12-16

Changing it to `display: none` the "Close" button doesn't allocate space when hidden.

SebastianKrupinski:
> The reasons of CSS changes lies in
> 
> ```
> .app-full__header__top__first {
>   flex-direction: column;
> }
> ```

Hard to see what you are trying to address without context

Before CSS:

image

After CSS:

image

SebastianKrupinski:
@madbob

I am not crazy about the design choice, the wording is in the wrong place and not very visible.  

Lets keep the PR to "fix: hide invitation reply buttons on cancelled event" and you can make another PR to "feat: indicate cancelled event" because we'll need to get design involved for those changes.

madbob:
Ok.

I just want to point out that `EditSimple` has not been changed (I've moved the definition of `isCancelled`, but nothing in contents nor in CSS. Including the "This event was cancelled" notice, which was already there), all visual changes are in `EditFull`.

Here a more complete comparison screenshot (I had to provide it before... :-\ ).

sample

SebastianKrupinski:
> Ok.
> 
> I just want to point out that `EditSimple` has not been changed (I've moved the definition of `isCancelled`, but nothing in contents nor in CSS. Including the "This event was cancelled" notice, which was already there), all visual changes are in `EditFull`.

Well that is interesting because the screen captures where with and without your code.

SebastianKrupinski (approved):

PR #8890, opened 2026-09-08
fix: use standard hover style for navigation date picker

### LINKED ISSUE #8888: Weird custom hover styling for date picker
When the date picker is hovered, a white border is drawn around it, which also clips into the neighboring button. It should use the standard hover effect of buttons, which is the hover background color.

Image

abhinavohri:
Fixes: https://github.com/nextcloud/calendar/issues/8888

Summary:
The navigation date picker is rendered as an input, so it inherits the global input hover box-shadow, which overlaps the adjacent navigation buttons. This removes that shadow on hover and applies the same hover background color used by NcButton.

Before:
image

After:
image

### DIFF
--- a/src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderDatePicker.vue
+++ b/src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderDatePicker.vue
@@ -201,12 +201,17 @@ useHotKey(['p', 'k'], () => navigateTimeRangeBackward())
 		border-radius: 0 !important;
 		text-align: center;
 		border: 1px solid var(--color-primary-element-light-hover) !important;
-		font-weight: bold;
+		font-weight: var(--font-weight-element);
 		background-color: var(--color-primary-element-light) !important;
 		margin: 0 !important;
 
 		padding: 0 !important;
 		width: calc(var(--default-grid-baseline) * 54) !important;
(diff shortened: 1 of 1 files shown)


SebastianKrupinski:
Hi @abhinavohri 

Thank you for the PR.

SebastianKrupinski:
@kra-mo 

Please have a look at this PR as it attempts to address your issue ticket.

kra-mo (approved):
I see, neat!

SebastianKrupinski:
@kra-mo 

The middle section is a clickable input. Should it not stand out?

image

kra-mo:
@SebastianKrupinski I don't think so. Even though it's an input, it practically works like a button.

src/components/AppNavigation/AppNavigationHeader/AppNavigationHeaderDatePicker.vue
kra-mo: In fact, it should also use `font-weight-element` if we can get that in in this PR too.

```suggestion
		font-weight: var(--font-weight-element);
```

abhinavohri: Done.

SebastianKrupinski (approved):

PR #8905, opened 2026-09-09
fix(attendees): respect sharing restrictions

### LINKED ISSUE #8726: Calendar invitation autocomplete ignores global "shareapi" and privacy restrictions
### Steps to reproduce

Go to Administration settings -> Sharing.
Under "Limit sharing based on groups", select "Allow some groups" and restrict it to a specific group (e.g., "admin").
Under "Privacy settings for sharing", enable "Allow autocompletion to full match when entering the full name...".
Ensure that only "Also allow autocompletion on full match of the user email" is checked, while other full-match options (like User ID or Display name) are disabled.
Log in as a normal user who is NOT part of the allowed "admin" group.
Open the Calendar app, create a new event, and type single characters or letters into the "Attendees" search field.

Image

### Expected behavior

The autocomplete should NOT show any users or email addresses, because sharing is restricted to admins, and incomplete matches are forbidden. This works perfectly in the Files app.

### Actual behavior

The Calendar search field completely ignores these administrative restrictions. It queries the global user database and exposes full names and email addresses to unauthorized users immediately upon typing single characters.

### Calendar app version

6.5.3

### CalDAV-clients used

_No response_

### Browser

Firefox 153.0.3

### Client operating system

Windows 11

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

None

### Nextcloud version

34.0.2

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

_No response_

abhinavohri:
## Description

Use the server share manager to stop attendee autocomplete when sharing is disabled for the current user. This prevents Calendar from exposing contact results to users outside the configured sharing allow list.

Closes https://github.com/nextcloud/calendar/issues/8726

### DIFF
--- a/lib/Controller/ContactController.php
+++ b/lib/Controller/ContactController.php
@@ -24,6 +24,7 @@ use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IRequest;
 use OCP\IUserManager;
+use OCP\Share\IManager as IShareManager;
 use Psr\Container\ContainerExceptionInterface;
 
 /**
@@ -48,6 +49,7 @@ class ContactController extends Controller {
 		private IAppConfig $appConfig,
 		private IConfig $config,
 		private IGroupManager $groupManager,
+		private IShareManager $shareManager,
 		private ?string $userId,
 	) {
 		parent::__construct($appName, $request);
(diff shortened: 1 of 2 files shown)


tcitworld:
I feel like sharing and adding attendees to an event are two different things. Disabling user enumeration should be enough, no need for this.

SebastianKrupinski (changes requested):
Hi @abhinavohri 

Limiting attendees to a sharing concept makes no sense here. This why this has not been implemented in the past.

PR #8918, opened 2026-09-09
fix(proposals): handle network errors without responses

### LINKED ISSUE #8915: ProposalService error handler crashes when Axios error has no response
### Steps to reproduce

1. Open Calendar and navigate to meeting proposals.
2. Disable the network in browser developer tools.
3. Try to list, create, update, delete, convert, or respond to a proposal.
4. Inspect the resulting error or browser console.

### Expected behavior

`ProposalService` should handle a request-level/network failure, log useful information, and preserve the original Axios error as the cause.

### Actual behavior

The error handler can throw a secondary `TypeError`, masking the actual network failure:

```text
TypeError: Cannot read properties of undefined (reading 'headers')
```

### Calendar app version

`6.7.0-dev.1` (`main` at commit `6dcc3a746`)

### CalDAV-clients used

Not applicable

### Browser

Any browser when the proposal request fails before receiving an HTTP response

### Client operating system

Not applicable

### Server operating system

Not applicable

### Web server

Not applicable

### Database engine version

Not applicable

### PHP engine version

Not applicable

### Nextcloud version

Development version compatible with Calendar `main`

Not applicable

### List of activated apps

Not applicable

### Nextcloud configuration

Not applicable

### Web server error log

Not applicable

### Log file

Not applicable

### Browser log

```text
TypeError: Cannot read properties of undefined (reading 'headers')
```

#### Cause

The catch block in `src/services/proposalService.ts` assumes every caught value has an Axios response:

```ts
if (error.response.headers['content-type'] && ...) {
	// ...
} else {
	message = `${error.response.status} ${error.response.statusText}`
}
```

Axios network, DNS, timeout, cancellation, and some browser-level errors can have no `response`. A caught value is also `unknown`, which is why `npm run ts:check` currently reports multiple `TS18046` errors for this block.

#### Suggested fix

Use `axios.isAxiosError(error)`, check whether `error.response` exists before reading it, and fall back to `error.message` or a generic request-failure message. The thrown service error should retain the original error as `cause`.

Please also add tests for both HTTP-response errors and Axios errors without a response.

Kumar-s75:
Proposal requests that fail without an HTTP response previously triggered a secondary TypeError in the error handler. Guard response access and return a useful service error for Axios failures.

Log only the structured response fields `status`, `statusText`, and `ocsError`. Do not retain the Axios error as a cause, because callers log service errors and could otherwise expose request configuration or request data. Unexpected non-Axios errors are rethrown unchanged.

Validation: all four regression tests pass (network, non-JSON HTTP, OCS, and unexpected non-Axios errors). Changed-file ESLint has no errors; seven existing cause-cast warnings remain. `git diff --check` passes.

Closes #8915


src/services/proposalService.ts
```diff
@@ -24,16 +24,18 @@ class ProposalService {
 			})
 			return response.data as T
 		} catch (error) {
-			let message: string
-			if (error.response.headers['content-type'] && error.response.headers['content-type'].includes('application/json')) {
+			let message = error instanceof Error ? error.message : 'Unknown error'
+			if (isAxiosError(error) && error.response?.headers['content-type']?.includes('application/json')) {
 				const ocsError = error.response.data as OcsEnvelope<OcsErrorData>
 				message = ocsError.ocs?.meta?.message ? ocsError.ocs.meta.message : `${error.response.status} ${error.response.statusText}`
 				logger.error('Proposal service transmission error', { ocsError })
```
odzhychko: This is ok, to because we only put `ocsError` into log context.

src/services/proposalService.ts
```diff
@@ -24,16 +24,18 @@ class ProposalService {
...
-			} else {
+			} else if (isAxiosError(error) && error.response) {
 				message = `${error.response.status} ${error.response.statusText}`
 				logger.error('Proposal service transmission error', { error })
```
odzhychko: Something like this is bad if `error` is an Axios error.

~Though this code path was dead.~
~If `error.reponse` was `undifined`/`null` `if (error.response.headers['content-type']` would have already caused a runtime error.~

src/services/proposalService.ts
```diff
@@ -24,16 +24,18 @@ class ProposalService {
...
+			} else {
+				logger.error('Proposal service transmission error', { error })
```
odzhychko: This would cause the full error to end up in logging context later.

src/services/proposalService.ts
```diff
@@ -24,16 +24,18 @@ class ProposalService {
...
 			}
-			throw new Error(`Unexpected error from proposal service: ${message}`)
+			throw new Error(`Unexpected error from proposal service: ${message}`, { cause: error })
```
odzhychko: This would cause the full error to end up in logging context later.

odzhychko (dismissed):
Axios errors should not be rethrown/added as cause at all, because everything thrown from `transceivePost` will end up in the `logger`.

The logger might be configure to send logs to error monitoring and we do not want `error.config` or `error.request` to end up there.

Not sure if `error.config` includes something problematic.
But `error.request` might include personal data.

See https://github.com/axios/axios/blob/1226e6c53eeb0e9d7159bea83556c2ff117b5f2e/index.d.cts#L134 for values on an Axios error.

---

Practical solution:

- If axios error
  - log error with status, statusText and ocsError in context
  - extract message with the current logic
  - throw new error containing the message
    - do not add axios error as a cause
      - add comment to disable lint
- else
  - just rethrow as unexpected error

src/services/proposalService.ts
odzhychko: Never mind. Code path is not dead.

Kumar-s75:
Addressed in the fixup commit: network and non-JSON HTTP failures now log only `{ message }`, and the thrown service error no longer retains the raw Axios error as its cause. The approved `{ ocsError }` context is unchanged. Three regression tests cover network, HTTP and OCS failures and verify that the Axios request configuration cannot reach the outer logger through the cause chain. All three pass. Please take another look when convenient.

Kumar-s75:
Followed the full review's proposed structure in the latest fixup: Axios errors log `{ status, statusText, ocsError }` and become a new message-only error without the raw Axios cause. Non-Axios errors are rethrown unchanged. All four regression tests pass, including the new non-Axios test; lint has no errors. Ready for another review.

odzhychko (commented):
Looks good to me 👍 

Leaving it for the second reviewer as it follows my proposed solution.
Might worth checking that my reasoning did not mix something up.

Kumar-s75:
@kurtmckee Sir, can you review this ?

SebastianKrupinski (approved):

PR #8925, opened 2026-09-11
fix: correct "widht" typos in PropertySelect comments

TianHengZhuang:
Two typos in the SCSS block of `PropertySelect.vue`: `widht` -> `width`.

Comment-only change, no rendering or behaviour impact.


GVodyanov (approved):
Nice, thanks

PR #8931, opened 2026-09-12
fix(proposals): handle invalid timezone fallback

### LINKED ISSUE #8928: Invalid meeting proposal timezone causes ReferenceError in offset fallback
### Steps to reproduce

1. Check out the current `main` branch.
2. Call `getTimezoneOffset()` with an invalid timezone identifier:

```ts
getTimezoneOffset(
    new Date('2026-07-11T00:00:00Z'),
    'Invalid/Timezone',
)
```

Alternatively, run `npm run ts:check`.

### Expected behavior

The timezone formatting error is logged and the function returns its intended fallback offset of `0`.

### Actual behavior

The catch block throws a second error:

```text
ReferenceError: logger is not defined
```

TypeScript also detects the undeclared identifier:

```text
src/services/timezoneOffsetService.ts(48,3):
error TS2304: Cannot find name 'logger'.
```

The problem is in `src/services/timezoneOffsetService.ts`:

```ts
} catch (error) {
    logger.error('Failed to compute timezone offset', { error })
    timezoneOffset = 0
}
```

Unlike other services using the Calendar logger, this module does not import it. The error-handling path therefore fails before the fallback can be returned.

### Calendar app version

Current `main` (6.7.0-dev.1)

### Additional info

A focused regression test using an invalid timezone would ensure the fallback path remains functional.

Kumar-s75:
## Summary

- import the Calendar logger used by the timezone-offset fallback
- add a regression test for invalid timezone identifiers
- preserve the intended UTC offset fallback instead of throwing a `ReferenceError`

## Tests

- `npm run test:unit -- tests/javascript/unit/services/timezoneOffsetService.test.ts --reporter=dot`
- `TZ=UTC npm run test:unit -- --reporter=dot` (519 tests passed)
- `npx eslint src/services/timezoneOffsetService.ts tests/javascript/unit/services/timezoneOffsetService.test.ts`

Closes #8928

### DIFF
--- a/src/services/timezoneOffsetService.ts
+++ b/src/services/timezoneOffsetService.ts
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+import logger from '@/utils/logger.js'
+
 /**
  * Get the offset in minutes between a date's timezone and the given timezone
  *
--- a/tests/javascript/unit/services/timezoneOffsetService.test.ts
+++ b/tests/javascript/unit/services/timezoneOffsetService.test.ts
@@ -1,0 +1,27 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { describe, expect, it, vi } from 'vitest'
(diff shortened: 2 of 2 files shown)


odzhychko:
Duplicate of https://github.com/nextcloud/calendar/pull/8917

PR #8935, opened 2026-09-13
fix(calendar): use sidebar open lifecycle for unscheduled tasks

ThorbenAroundTheWorld:
The Calendar view keeps `NcAppSidebar` mounted while hiding it with `v-show` when unscheduled tasks are closed. At mobile width, the sidebar focus-trap lifecycle can remain active while the UI is hidden, so repeated Tab does not reach the calendar content.

Bind the supported `open` prop to `tasksSidebar && tasksSidebarEnabled`. This keeps `UnscheduledTasksList` mounted for task discovery while allowing `NcAppSidebar` to activate and deactivate its focus lifecycle with the visible state.

Base: `stable6.5` at `c00fc871effbc7b1d33f3bc362fe395e30f86b31` (Calendar 6.5.4).

Validation:
- 320 px component regression uses real `NcContent` and `NcAppSidebar`; 2/2 tests pass.
- The test proves focus-trap activation only while open, deactivation on close, and return focus to the calendar action.
- A `v-show` preimage negative control fails as expected.
- `git diff --check` and the production Rspack build pass.

Native browser verification remains required after an official signed release.


SebastianKrupinski:
Hi @ThorbenAroundTheWorld 

Thank you for the PR and contribution.

SebastianKrupinski:
This should be applied to main

src/views/Calendar.vue
```diff
@@ -75,7 +75,8 @@
 
 		<NcAppSidebar
 			v-if="isAuthenticatedUser"
-			v-show="tasksSidebar && tasksSidebarEnabled"
+			:open="tasksSidebar && tasksSidebarEnabled"
```
SebastianKrupinski, line 78: Using :open by it self breaks, reopening the side panel after the first round of open and close.  You need to add "no-toggle" to disable the internal toggle mode

I made the change for you

SebastianKrupinski (approved):

PR #8951, opened 2026-09-19
feat(talk): record the conversation as the event's conference

Rikdekker:
A Talk link was only stored in LOCATION, or appended to DESCRIPTION when LOCATION already held something. Neither says "this is where the meeting happens online", and for a hybrid meeting LOCATION is needed for the room the event takes place in.

Write the conversation to CONFERENCE (RFC 7986), which is the property meant for a virtual meeting URL. An earlier conversation is replaced rather than stacked, so an event keeps a single conference.

Talk currently only finds events whose call link is in LOCATION, so CONFERENCE alone would make a hybrid meeting invisible there. That is why the existing LOCATION/DESCRIPTION behaviour stays as it is — clients that do not read CONFERENCE keep working exactly as before. nextcloud/spreed#19472 adds the DESCRIPTION side in Talk.

Tested with the resulting calendar objects on Nextcloud 35.0.0 (Calendar 6.6.0-rc.2, Talk 25.0.0), with nextcloud/spreed#19472 applied: a meeting with a room in LOCATION and a conversation is found by Talk, and so is an online-only meeting. The replace-not-stack behaviour was verified against @nextcloud/calendar-js directly.

## 🤖 AI (if applicable)


tcitworld:
Reference: https://github.com/nextcloud/calendar/issues/1758

